PR #1, opened 2022-12-05
super-select: allow growing height if multiple

pascallaliberte:
Fixes https://github.com/bullet-train-co/bullet_train-fields/issues/44


andrewculver:
@ps-ruby Can you test this?

ps-ruby:
> @ps-ruby Can you test this?
I have tested that for event types in New Endpoint Form and it is working fine.



image

ps-ruby (approved):
LGM

PR #7, opened 2022-12-16
Update `bin/hack` for new environment

### LINKED ISSUE #5: Rewrite and simplify `bin/hack`.
Instead of prompting people for which package they want to work on, we should just clone the entire core repo from http://github.com/bullet-train-co/bullet_train-core and link up _all_ of the Ruby gems in the local application's `Gemfile`. You might be able to reuse some of what we do in https://github.com/bullet-train-co/bullet_train-core/blob/main/bin/checkout-and-link-starter-repo for this.

We should prompt asking them which branch they want to check out.

gazayas:
Closes #5.

There are still some things I need to figure out so I've opened this up as a draft, but I did want to mention the `--link` and `--reset` flags since I already had that logic written and didn't want to reinvent the wheel so to speak. If we're okay with just using that to link up all the gems then great, if not I'll reference the script you mentioned in the issue.

First thing's first, I'll try to get the initial logic running properly when cloning the core repository.

### DIFF
--- a/bullet_train/lib/tasks/bullet_train_tasks.rake
+++ b/bullet_train/lib/tasks/bullet_train_tasks.rake
@@ -107,65 +107,37 @@ namespace :bullet_train do
       if flags_with_values.any?
         flags_with_values.each do |process|
           if process[:flag] == :link || process[:flag] == :reset
-            packages = I18n.t("framework_packages").keys.map { |key| key.to_s }
-
-            gemfile_lines = File.readlines("./Gemfile")
-            new_lines = gemfile_lines.map do |line|
-              packages.each do |package|
-                if line.match?(/\"#{package}\"/)
-                  original_path = "gem \"#{package}\""
-                  local_path = "gem \"#{package}\", path: \"local/bullet_train-core/#{package}\""
-
-                  case process[:flag]
-                  when :link
-                    line.gsub!(original_path, local_path)
-                    puts "Setting local '#{package}' package to the Gemfile...".blue
-                    break
-                  when :reset
-                    line.gsub!(local_path, original_path)
-                    puts "Resetting '#{package}' package in the Gemfile...".blue
-                    break
-                  end
-                end
-              end
-
-              line
-            end
-
-            File.write("./Gemfile", new_lines.join)
-            system "bundle install"
+            set_core_gems(process[:flag])
+            system("bundle install")
+            exit
           end
         end
-
-        exit
       end
     end
 
     puts "Welcome! Let's get hacking 💻".blue
-    puts ""
 
+    # Adding these flags enables us to execute git commands in the gem from our starter repo.
+    work_tree_flag = "--work-tree=local/bullet_train-core"
+    git_dir_flag = "--git-dir=local/bullet_train-core/.git"
     framework_packages = I18n.t("framework_packages")
 
     if File.exist?("local/bullet_train-core")
       puts "We found the repository in `local/bullet_train-core`. We will try to use what's already there.".yellow
       puts ""
 
-      # Adding these flags enables us to execute git commands in the gem from our starter repo.
-      work_tree_flag = "--work-tree=local/bullet_train-core"
-      git_dir_flag = "--git-dir=local/bullet_train-core/.git"
-
       git_status = `git #{work_tree_flag} #{git_dir_flag} status`
       unless git_status.match?("nothing to commit, working tree clean")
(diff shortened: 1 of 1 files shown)


gazayas:
Ok, this will clone and link/reset bullet_train-core gems correctly, the only two things I have left to figure out are:
1. `yalc` the npm packages so developers can edit JS too.
2. Have messages when resetting gems similar to what we have when linking them.

gazayas:
Ok, we have the main functionalities down so we can do the following:

1. `bin/hack`: This will clone/check for the `bullet_train-core` and link the gems.
2. `bin/hack --reset`: This will change all the gems in the starter repository's Gemfile back to their original state.
3. `bin/hack --link`: This will set all the gem's paths to `"local/bullet_train-core/#{package_name}"`
4. `bin/hack --link_js bullet_train-fields`: This will watch for changes in the gem that is designated after the flag.

For the fourth point, I found it pretty difficult to just run `yarn watch` in the `bullet_train-core` directory and watch all the JavaScript files. I was able to open a different terminal for each gem that has an npm package and watch changes that way, but I definitely think it would be easier for developers if they didn't have to think about this.

@pascallaliberte is there any way we can watch for changes across all packages? I looked into `package.json` for the different gems, but if we just have one in the `bullet_train-core` directory, does that mean we'll have to create a package for `bullet_train-core` as well? I don't want to delve too much into this and go in the wrong direction, so I'd definitely appreciate your input.

@andrewculver Test it out, let me know how it feels, and I'd be glad to make any changes as necessary! (I feel like it takes a bit initially to start up, but it might just be me).

pascallaliberte:
> @pascallaliberte is there any way we can watch for changes across all packages? I looked into `package.json` for the different gems, but if we just have one in the `bullet_train-core` directory, does that mean we'll have to create a package for `bullet_train-core` as well? I don't want to delve too much into this and go in the wrong direction, so I'd definitely appreciate your input.

We'd have to run multiple yarn watch instances: one for each package which has one of those. Maybe through a Procfile or whatever npm uses for the same multi-process wrapper.

gazayas:
I was able to run multiple processes for each npm package. Here are a couple of new flags:
1. `bin/hack --watch-js`: Watches changes in JS for all npm packages
2. `bin/hack --clean-js`: Resets all npm packages to what they used to be in `package.json`

However, there were a couple of issues I ran into so I'll go through those here.

## Our original code doesn't reflect changes in JS
Unless I'm missing something really big, there seems to be a couple of issues with our original setup unrelated to this PR. I specifically touched upon this briefly in https://github.com/bullet-train-co/bullet_train-base/issues/145, but our JS changes aren't reflected even if we edit the files in the `local` directory. The changes are being watched for so the server responds, but the code doesn't actually show up in the browser because the JS being rendered is the `esm.js` file that's built, not the actual `.js` files that we edit.

For `bullet_train-fields`, you can find this file in the `node_modules` directory:
`node_modules/@bullet-train/fields/app/assets/javascripts/fields.esm.js`

The only way I know to overwrite this is to run `yarn build` in that package's main directory each time we make a change to the JS. After doing so, the changes are finally reflected in the browser. Again I don't want to go down a rabbit hole with this one, but I definitely want to find a simple answer to this so we can edit our JS files with ease.

## bullet_train-sortable throwing an error

When trying to `yalc link` with `bullet_train-sortable` specifically, we're getting this error in the `bin/dev` server (again, this is with the original `bin/develop` code).

```
s               | ✘ [ERROR] Could not resolve "@rails/request.js"
js               | 
js               |     .yalc/@bullet-train/bullet-train-sortable/app/assets/javascripts/bullet-train.esm.js:1:92:
js               |       1 │ ...";import{post as t}from"@rails/request.js";function a(e){var n=e...
js               |         ╵                           ~~~~~~~~~~~~~~~~~~~
js               | 
js               |   You can mark the path "@rails/request.js" as external to exclude it from the
js               |   bundle, which will remove this error.
```

To circumvent this I prompt the developer in `def set_npm_packages` to start their server before setting up the npm packages, but this is something to look out for.

## Running `yarn watch` in parallel
As far as the `yarn watch` commands I've added here, there is a [package that runs yarn commands in parallel](https://github.com/mysticatea/npm-run-all), but just connecting commands with `&` seems to be the main consensus: [Post](https://stackoverflow.com/questions/52151353/yarn-run-multiple-scripts-in-parallel)
(The package mentioned above also talks about running processes in parallel with `&`).

PR #20, opened 2022-12-21
Fix duplicate rendering issue

gazayas:
Closes https://github.com/bullet-train-co/bullet_train/issues/550

## Details

We were wrapping content with the `action_model_select_controller` method which simply calls `yield` when the developer doesn't have Action Models enabled. How we add content to this block can cause it to duplicate content.

## Duplicating with a vanilla Rails app
We get duplicate content if we do the same thing like this:

```ruby
# Write this method in a helper
def foo
  yield
end
```

```erb
# Call a partial named "bar" in any view
<%= foo do %>
  <%= render "bar" do %>
    Duplicated content
  <% end %>
<% end %>
```

```erb
# Yield the content in the `_bar.html.erb` partial
<%= yield %>
```

## CHANGELOG (Option 1)
1. Tell <b>all</b> the developers to update their Super Scaffolded views with this `capture` fix.

## CHANGELOG (Option 2)
1. Tell developers who aren't using Action Models to simply delete the `action_model_select_controller` block
2. Tell developers who already use Action Models to update their views to use the `capture` fix in this PR.

## My opinion
Option 1 is probably best because, although all developers will have to update their Super Scaffolded views with this `capture` fix even if they don't use Action Models, once it's set they won't have to think about it if they decide to upgrade to Action Models in the future. I'm also assuming that there's a reason for having the action model select controller code here even if Action Models isn't defined, so I decided to leave the logic there.

If there is a specific path you'd like to take concerning the CHANGELOG, I'd be glad to write it.

### DIFF
--- a/bullet_train-api/app/views/account/platform/access_tokens/_index.html.erb
+++ b/bullet_train-api/app/views/account/platform/access_tokens/_index.html.erb
@@ -7,7 +7,7 @@
 <% access_tokens = access_tokens.order(:id) unless has_order?(access_tokens) %>
 <% pagy, access_tokens = pagy(access_tokens, page_param: :access_tokens_page) %>
 
-<%= action_model_select_controller do %>
+<% content = capture do %>
   <%= updates_for context, collection do %>
     <%= render 'account/shared/box', pagy: pagy do |p| %>
       <% p.content_for :title, t(".contexts.#{context.class.name.underscore}.header") %>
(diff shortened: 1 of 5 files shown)


gazayas:
One more thing to note is that the content is duplicated or not depending on the presence of `<%= ... %>`.

Again, with the vanilla rails app above:
```erb
<%= foo do %>
  This is duplicated
<% end %>

<%= foo do %>
  <%= "This is not duplicated" %>
<% end %>

PR #25, opened 2022-12-23
Replace custom path logic with `render` prefixes

kaspth:
The idea is that a theme render gets converted from `render "shared/fields/text_field"` to

```ruby
render "fields/text_field", prefixes: ["themes/light", "themes/tailwind_css", "themes/base"]
```

So we can trim a lot of the custom rendering logic.

### DIFF
--- a/bullet_train-themes/app/helpers/theme_helper.rb
+++ b/bullet_train-themes/app/helpers/theme_helper.rb
@@ -41,7 +41,8 @@ module ThemeHelper
 
   def render(options = {}, locals = {}, &block)
     if (theme_path = BulletTrain::Themes.theme_invocation_path_for(options))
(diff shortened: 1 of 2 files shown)


kaspth:
Blegh. Looks like partial renders don't support `prefixes`, so this won't work.

I think we could leverage Rails' built-in `Rails::Engine` view paths setup — where each Engine prepends their `app/views` directory to `ActionController::Base.view_paths` (essentially) — by having a consistent directory structure between themes.

I.e. `app/views/bullet_train/themes/fields/_text_field.html.erb`" and then `view_paths` would determine which theme partial got picked. I'm not sure how to resolve the inner theme render, i.e. light has to render from tailwind_css where we need directories to match.

And then maybe these changes may be more hassle than they're worth ultimately. I'll let it be for a bit.

PR #36, opened 2022-12-28
Add `super-scaffold` command

gazayas:
Originally here: https://github.com/bullet-train-co/bullet_train-super_scaffolding/pull/6
and continued here: https://github.com/bullet-train-co/bullet_train-super_scaffolding/pull/41

I decided to open a new PR since we migrated to `bullet_train-core`.

## Command example
```
> rails g model Foo team:references title:string
> bundle exec super-scaffold crud Foo Team title:text_field
```

## Details
All of the code is the same, except I found I had to require a couple of libraries to make things work properly:
```ruby
# In bullet_train-super_scaffolding/exe/super-scaffold
require "rails/engine"
require "bullet_train/api"

# In bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/engine.rb
require "jbuilder"
```

## Credit
I wasn't able to pull over @indirect's commits, but if we want to rebase it off of those then I can try to open with another PR.


andrewculver:
Going to skip this. Don't want it to be confusing for anyone and we're already so deep with docs and videos that have the `bin/super-scaffold` syntax.

PR #46, opened 2023-01-04
Start documenting partials with Showcase

kaspth:
Uses a version of Showcase where we can fetch from each known Rails engine. So this is what it can look like in a starter app:

Screen Shot 2023-01-04 at 21 44 20


kaspth:
The sidebar needs some UI cleanup though.

Issue #65, opened 2023-01-13
Put screenshots of different color schemes in docs


gazayas:
Developers can change the the color of their UI in `config/initializers/theme.rb`. However, they don't know what colors are available to them unless they search the source code, and even if they did, they'd have to change the color in the initializer, restart the server and check the browser each time they wanted to view a new color scheme

I'd like to:
1. Put the name of each theme along with a screenshot of what it looks like somewhere in the docs.
2. Put a link in `config/initializers/theme.rb` to point them to the docs in case developers aren't sure of what to do.

gazayas:
Closing this one since #106 was merged in.

PR #66, opened 2023-01-13
Add documentation for color schemes

gazayas:
Completes the first task of #65.

For the screenshots, I used the links in #65 because I'm not sure if there's another place where we should host the pictures.


andrewculver:
@gazayas Replaced color schemes with the ability for users to just specify their own Tailwind CSS colors.

PR #70, opened 2023-01-15
Speed up CI

kaspth:
Brings over stuff I learned in https://github.com/bullet-train-co/nice_partials/pull/75

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 orbs:
   ruby: circleci/ruby@0.1.2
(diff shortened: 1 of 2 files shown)


kaspth:
blegh, this is difficult

kaspth:
Ok, this gets bundler caching working for both the Starter Repo and Starter Repo Super Scaffolding CI jobs, which shaves a full minute off `bundle install` time. I have no idea why it's still at ~20 seconds and not 0 though.

Additionally, we set the Firefox version to what's already installed and save another 20-30 seconds in each parallel unit.

PR #73, opened 2023-01-17
Enable 2FA automatically when credentials are set

gazayas:
Found by @eadz and also mentioned in https://github.com/bullet-train-co/bullet_train/pull/519#issuecomment-1340037608

## Backwards compatibility
I've decided to not consider backwards compatibility here because of [this comment](https://github.com/bullet-train-co/bullet_train/pull/519#issuecomment-1339992627), but we can use the following like @eadz suggested if we want to check for either the old environment variable or the Rails 7 credentials value:

```ruby
ENV['TWO_FACTOR_ENCRYPION_KEY'].present? || Rails.application.credentials.active_record_enc
ryption.present?
```

## Turbo issue
We still need to fix the turbo-related issue, but for now you can refer to #50 if you want to test out the logic manually in the browser.

### DIFF
--- a/bullet_train/app/models/concerns/users/base.rb
+++ b/bullet_train/app/models/concerns/users/base.rb
@@ -3,7 +3,7 @@ module Users::Base
 
   included do
     if two_factor_authentication_enabled?
-      devise :two_factor_authenticatable, :two_factor_backupable, otp_secret_encryption_key: ENV["TWO_FACTOR_ENCRYPTION_KEY"]
+      devise :two_factor_authenticatable, :two_factor_backupable
     else
       devise :database_authenticatable
     end
--- a/bullet_train/lib/bullet_train.rb
+++ b/bullet_train/lib/bullet_train.rb
@@ -115,7 +115,7 @@ def cloudinary_enabled?
 end
 
 def two_factor_authentication_enabled?
-  ENV["TWO_FACTOR_ENCRYPTION_KEY"].present?
+  Rails.application.credentials.active_record_encryption.primary_key
 end
 
 # Don't redefine this if an application redefines it locally.
--- a/bullet_train/lib/bullet_train.rb
+++ b/bullet_train/lib/bullet_train.rb
@@ -115,7 +115,7 @@ def cloudinary_enabled?
 end
 
 def two_factor_authentication_enabled?
-  Rails.application.credentials.active_record_encryption.primary_key
+  Rails.application.credentials.active_record_encryption&.primary_key.present?
 end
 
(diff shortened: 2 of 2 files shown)


eadz:
Reading this - https://github.com/tinfoil/devise-two-factor/blob/main/UPGRADING.md 

I can't see any legacy_otp_secret method definition in bullet train. As far as I can tell, that means the upgrade to v5 of devise-two-factor will break two factor for users if they are coming from the old version. 

I don't think this change breaks it, I think it's already broken with the v5 upgrade? Or were there no users using the earlier version?

gazayas:
## `otp_legacy_secret`
Since this PR was closed, I'm assuming we're not going to want to support older versions:
- https://github.com/bullet-train-co/bullet_train-base/pull/142

## Breaking changes for developers on upgrade
You're totally right that things would break for developers upgrading to v5 since we wouldn't be setting `otp_secret_encryption_key: ENV["TWO_FACTOR_ENCRYPTION_KEY"]` anymore in the User base model in the code here in this PR.

## Letting developers do the update manually
I'm looking at the docs though and I wonder if we want to leave the migration from v4 to v5 to the developers instead of handling it ourselves? Developers would have to add `legacy_otp_secret` to their User model, do the following rake task (as mentioned per the docs) and then DELETE the same `legacy_otp_secret` method after running the rake task:

```ruby
# lib/tasks/devise_two_factor_migration.rake

# Use this as a starting point for your task to migrate your user's OTP secrets.
namespace :devise_two_factor do
  desc "Copy devise_two_factor OTP secret from old format to new format"
  task copy_otp_secret_to_rails7_encrypted_attr: [:environment] do
    # TODO: change User to your user model
    User.find_each do |user| # find_each finds in batches of 1,000 by default
      otp_secret = user.otp_secret # read from otp_secret column, fall back to legacy columns if new column is empty
      puts "Processing #{user.email}"
      user.update!(otp_secret: otp_secret)
    end
  end
end
```

I'm all about maintaining backwards compatibility for developers, but it seems counter-intuitive to me to add code that we should someday delete, especially when we don't know if our developers have updated their User columns or not. Also, as the docs say:
> We have attempted to make the upgrade as painless as possible but unfortunately because of the secret storage change, it cannot be as simple as `bundle update devise-two-factor`

I think we could say the same about updating `devise-two-factor` in the context of a Bullet Train app as well. Although it's unfortunate to have to set up manually, I think it would complicate things if we tried to handle the upgrade in-house while keeping code for a 3rd party library update in our code base.

However, that's just my opinion and I'd like to hear everyone else's thoughts in case we want to handle it differently.

PR #74, opened 2023-01-18
Handle invitation for invited user signing up with omniauth

### LINKED ISSUE #71: OAuth Super Scaffolding returns error on sign up
As I was looking into https://github.com/bullet-train-co/bullet_train/issues/578, I tried with a couple of different providers (Google and Spotify specifically) and got the following error when trying to sign up:

OAuth Error

https://github.com/bullet-train-co/bullet_train-core/blob/cff6331fc6efc3002941729d375dce8eed9f854f/bullet_train-integrations/app/controllers/concerns/account/oauth/omniauth_callbacks/controller_base.rb#L4-L6

`request.env.dig("omniauth.params")` in `team_id_from_env` is returning `{}`.

gazayas:
Closes #71.

## Details
Our OAuth sign up/sign in workflow was failing because the method `handle_outstanding_invitation` hadn't been migrated to `bullet_train-core`.

I also decided to add the method directly to this file as a private method instead of via a helper like before because this is the only place where the method appears.

I was able to invite a new user and sign up with Spotify, so this one should be good to go.

### DIFF
--- a/bullet_train-integrations/app/controllers/concerns/account/oauth/omniauth_callbacks/controller_base.rb
+++ b/bullet_train-integrations/app/controllers/concerns/account/oauth/omniauth_callbacks/controller_base.rb
@@ -149,4 +149,22 @@ module Account::Oauth::OmniauthCallbacks::ControllerBase
       redirect_to new_user_session_path, notice: t("omniauth.user.account_not_found", provider: t(auth.provider))
     end
   end
+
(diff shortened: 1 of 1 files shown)


kitebuggy:
I can confirm this fixed the bug for me. Thanks!

frmsaul:
nice!

PR #76, opened 2023-01-19
Fix docs navigation for the 'application-options' link

bojates:
In the docs, the link to the application options doesn't work. This should fix it.


gazayas:
Fixed this along with a number of other things in #89.

MSAdministrator:
This is still broken fyi.

PR #77, opened 2023-01-20
Prevent developers from Super Scaffolding the same content twice

gazayas:
## Details
Developers can technically run the same Super Scaffolding command multiple times, and this was causing duplicate lines to show up in specific files (https://github.com/bullet-train-co/bullet_train/issues/623).

As mentioned in https://github.com/bullet-train-co/bullet_train/issues/623, Rails prevents duplicate model generation, so I think we should have something similar in Bullet Train. However, Rails does this by checking if the constant already exists or not in a method called `class_collisions`:
https://github.com/rails/rails/blob/6e127924b690ead14e058225290c2a7f5b1e1335/railties/lib/rails/generators/base.rb#L278-L281.

We unfortunately can't use this since `bin/super-scaffold` depends on running `rails g model` beforehand and it also does different things depending on the scaffolding type (`crud`, `crud-field`, etc.), so I had to think through what ways make the most sense to check if a developer has Super Scaffolded some content already.

## Thoughts about what to check
Although I don't think this is the most conventional way to prevent developers from running the same command twice, I wrote what makes the most sense to me according to my current knowledge of the scaffolders. This included checking for certain files, the existence of certain modules, etc.

## join-model
Although I want to use `primary_parent` and `secondary_parent` to check for the presence of `has_many :foo, through: :bar` in each respective model, I'm still not entirely sure how to approach this one due to the `class_name` and how it's being defined in models. For example, we use `Project::AppliedTags` in the example from the docs, and because this changes to `:applied_tags` and defines `class_name: 'Project::AppliedTags'` earlier on in the model, I'd have to think through how check for this using the input from the command line.

### DIFF
--- a/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_field_scaffolder.rb
+++ b/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_field_scaffolder.rb
@@ -29,6 +29,7 @@ module BulletTrain
           check_required_options_for_attributes("crud-field", attributes, child)
 
           transformer = Scaffolding::Transformer.new(child, parents, @options)
+          transformer.check_if_has_been_scaffolded(options: {attributes: attributes})
           transformer.add_attributes_to_various_views(attributes, type: :crud_field)
 
           transformer.additional_steps.uniq.each_with_index do |additional_step, index|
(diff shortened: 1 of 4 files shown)


andrewculver:
@gazayas We actually don't want to stop people from running scaffolding twice. This is a common workflow where sometimes people rescaffold something to reset it or get an updated version of it, and then reset everything they don't want. We should just fix the duplicate content insertion.

gazayas:
Got it, I'll see if I can take another shot at the issue again soon.

PR #78, opened 2023-01-21
Fix spacing in Super Scaffolded API test setup

gazayas:
## Setup data
```
rails g model Foo team:references title:string && \
bin/super-scaffold crud Foo Team title:text_field --sidebar="ti.ti-world" && \
rails g model Bar foo:references title:string && \
bin/super-scaffold crud Bar Foo,Team title:text_field && \
rails g model Baz bar:references title:string && \
bin/super-scaffold crud Baz Bar,Foo,Team title:text_field
```

## The problem
When Super Scaffolding a model directly off of `Team`, `@foo` gets scaffolded as a factory to the API controller `setup` method like this:

```ruby
class Api::V1::FoosControllerTest < Api::Test
    def setup
      # See `test/controllers/api/test.rb` for common set up for API tests.
      super

      @foo = build(:foo, team: @team)
      @other_foos = create_list(:foo, 3)
```

However, for any nested models, the variables don't get indented properly:
```ruby
    def setup
      # See `test/controllers/api/test.rb` for common set up for API tests.
      super

      @foo = create(:foo, team: @team)
@bar = create(:bar, foo: @foo)
@baz = build(:baz, bar: @bar)
      @other_bazs = create_list(:baz, 3)
```

This was happening because we use `gsub!` to replace the Super Scaffolding hook in this file in [the Transformer](https://github.com/bullet-train-co/bullet_train-core/blob/fab77efab57126c083e0041f97c0e716560a2ffe/bullet_train-super_scaffolding/lib/scaffolding/file_manipulator.rb#L35):

```ruby
target_file_content.gsub!(in_place_of, content)
```

## The fix
`gsub!` starts replacing after the indentation where it matches the hook, but doesn't add indentation for any of the lines afterwards. We could technically just add space to `join` in the original code:

```ruby
build_factory_setup.join("  \n")
```

Since we already have the file/block manipulators in place though, I thought it would be better to use those instead. I also wrote things this way because the spacing in the file itself might change (I plan on opening a PR about that here in a minute). I know that taking an extra step to open the file just to read the lines and then replace the content afterwards is slower, but I think this makes the code easier to understand, and we can edit indentation more easily.

## The result
Now the setup block looks like this when scaffolding the three models:
```ruby
class Api::V1::BazsControllerTest < Api::Test
    def setup
      # See `test/controllers/api/test.rb` for common set up for API tests.
      super

      @foo = create(:foo, team: @team)
      @bar = create(:bar, foo: @foo)
      @baz = build(:baz, bar: @bar)
      @other_bazs = create_list(:baz, 3)
```

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/block_manipulator.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/block_manipulator.rb
@@ -214,7 +214,7 @@ module Scaffolding
             # If we're shifting a block to the left, we want to safeguard
             # the String so it doesn't delete any excess characters.
             if direction == :left
-              amount.times { line = line.gsub(/^\s/, "") }
+              amount.times { line = line.gsub(/^ /, "") }
             elsif direction == :right
               line = "\s" * amount + line
             end
--- a/bullet_train-super_scaffolding/lib/scaffolding/file_manipulator.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/file_manipulator.rb
@@ -3,7 +3,7 @@ require "scaffolding/block_manipulator"
 # TODO: If we move this and the BlockManipulator into their own gems,
 # we can probably call these methods with something shorter without `Scaffolding::`.
 module Scaffolding::FileManipulator
-  def self.find(lines, needle, within = nil)
+  def self.find(lines, needle, within = 0)
     lines_within(lines, within).each_with_index do |line, line_number|
       return (within + (within ? 1 : 0) + line_number) if line.match?(needle)
     end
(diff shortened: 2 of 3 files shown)


gazayas:
> I also wrote things this way because the spacing in the file itself might change (I plan on opening a PR about that here in a minute)

Here's the PR:
- https://github.com/bullet-train-co/bullet_train/pull/630

I was able to confirm that this works even when we change the spacing of the test file in the PR referenced here.

gazayas:
Ok, I was able to fix the spacing issue I tried to fix in https://github.com/bullet-train-co/bullet_train/pull/630.

Now the `Baz` API controller test for example looks like this when scaffolded:
```ruby
require "controllers/api/v1/test"

class Api::V1::BazsControllerTest < Api::Test
  def setup
    # See `test/controllers/api/test.rb` for common set up for API tests.
    super

    @foo = create(:foo, team: @team)
    @bar = create(:bar, foo: @foo)
    @baz = build(:baz, bar: @bar)
    @other_bazs = create_list(:baz, 3)

    @another_baz = create(:baz, bar: @bar)

    # 🚅 super scaffolding will insert file-related logic above this line.
    @baz.save
    @another_baz.save
  end

  ...
end
```

PR #86, opened 2023-01-26
Add address multi-field partials, add support for self-refreshing form fields

pascallaliberte:
Closes https://github.com/bullet-train-co/bullet_train/issues/631

This PR:
- Introduces `address_field`, our first multi-field `accepts_nested_attributes_for` partial
- Introduces a no-config mechanism for allowing form fields to fetch themselves new options and labels (e.g. Provinces/States/Prefecture depends on the country_id field)
- Introduces some Stimulus controllers for communicating between a dependent field (`dependable_controller.js`) and a `turbo_frame` that updates itself (`refresh_fields_controller.js`) based on changes to the dependent field (e.g. `country_id`).
- Adds `countries.json` and `states.json` to populate country and administrative division (region) fields (ie. Province, State, Prefecture)
- Adds an address field to the Tangible Thing model
- Improves wrapper_options param passthrough support for `super_select` field partial

Joint PR:
- https://github.com/bullet-train-co/bullet_train/pull/635

To Do:
- [X] Add `address_field` field partial
- [x] Add show partials for tables and `show` screens
- [X] Make administrative divisions and postal code fields update themselves on country change
- [x] Make administrative division field disabled if country field empty
- [x] Tighten up the layout by putting City, State/Province, Postal Code on one line when on desktop
- [x] ~Make sure address fields works with browser address autofill~ It works for the most part, except for country and state/province because of select2. See note below.
- [x] Make refreshed fields keep values entered prior to refresh
- [x] Prevent interaction with refreshing fields
- [x] Make tests pass


pascallaliberte:
Browser autofill only works with text fields (address line one, city, postal code), but doesn't work with country and province because they're select2-based fields. We'll have to treat this in another issue.

pascallaliberte:
Some notes on the no-config mechanism for refreshing fields:

The mechanism works by refreshing the current page with the dependent field (`country_id`) passed as a query_string param.

The kicker, though, is that **no changes are needed at the controller level**.

Right inside the `_address_field.html.erb` partial, you'll find:

https://github.com/bullet-train-co/bullet_train-core/blob/418bf24938bdef5549d53685a5d299978f11d988/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_address_field.html.erb#L13-L14

So that when the Stimulus controller updates the turbo_frame containing the province/state and postal_code fields with the updated country_id, the query_string parameter will take precedence.

https://github.com/bullet-train-co/bullet_train-core/blob/a331493dd74c256d72e5870a02c0f90dfa64ca56/bullet_train-fields/app/javascript/controllers/refresh_fields_controller.js#L7-L8

This encapsulation is convenient because at the field partial level, we have everything we need to know the query string param to look for (by knowing the field_name for the field). The Stimulus controller knows the field name as well, from the client side, by simply reading the dependent field's own `name` attribute.

https://github.com/bullet-train-co/bullet_train-core/blob/5ae7d603e4fe7ce97a744fa7facd4ca41a77e2d5/bullet_train-fields/app/helpers/refresh_fields_helper.rb#L1-L4

Question: Is it a problem that we're bypassing the strong_param safeguard in this context?

bullet_train/config/locales/en/addresses.en.yml
```diff
@@ -0,0 +1,128 @@
...
+        heading: *address_two
+      city:
+        _: &city City
+        label: *city
+        heading: *city
+      region_id: &region
+        _: &region_id State / Province / Region
+        label: *region_id
+        heading: *region_id
+        options:
```
andrewculver, line 22: We should populate this field from `Address::Region` vs. defining them like this.

bullet_train/app/models/concerns/addresses/regions/base.rb
```diff
@@ -0,0 +1,13 @@
+module Addresses::Regions::Base
+  extend ActiveSupport::Concern
+  
+  included do
+    self.data = JSON.parse(File.read(File.expand_path('../../../../../config/addresses/states.json', File.dirname(__FILE__))))
```
andrewculver, line 5: `Rails.root` is probably better than `../../../../.....`

bullet_train/app/models/address.rb
```diff
@@ -0,0 +1,24 @@
+class Team < ApplicationRecord
```
andrewculver, line 1: `Team`? Seems off.

pascallaliberte:
I really want to compress all these fields into a block of tight fields, but here's where we're at right now:

CleanShot 2023-01-26 at 19 14 17

bullet_train/config/locales/en/addresses.en.yml
pascallaliberte: I'm adding a new "admin_divisions" entry on each country, which is a key to use in the locale file (there won't be too many) e.g. (regions, provinces_territories, states, prefectures). There's a [table on wikipedia of how each country names its administrative divisions](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country) from which we can populate those keys to be localized for the label.

pascallaliberte:
Now using three columns for City, State/Province, Zip/Postal Code when on Desktop.

CleanShot 2023-01-27 at 11 45 55@2x

bullet_train/config/locales/en/addresses.en.yml
pascallaliberte: Removed the options

pascallaliberte:
@andrewculver I've started with changes to super-scaffolding in separate branch `add-addresses-with-scaffolding`. Do you want me to finish the scaffolding part too, and in this PR?

andrewculver:
@pascallaliberte Let's do Super Scaffolding support in a separate PR.

pascallaliberte:
@andrewculver 

> Can you get CI passing?

The required change to fix tests is in joint PR https://github.com/bullet-train-co/bullet_train/pull/635. Tests passed locally for me.

jagthedrummer:
@pascallaliberte I've rebased this branch against `main` and have fixed the linter errors, but there's a failing test that seems like it's maybe legit. Can you look into it?

Just fyi, CI now looks for a matching branch when pulling down the starter repo, so the tests are running against the `add-addresses` branch in both. The starter repo now also runs tests against the released version of the gems (which it's always done) _and_ against `core` with either a matching branch or using `main` if there's no matching branch. On https://github.com/bullet-train-co/bullet_train/pull/635 I'm seeing failures against the released gems, which I think is to be expected. But when running against this `add-addresses` branch in `core` I'm seeing the same test failure as we're seeing on this PR. (I hope all of this makes sense, it's not exactly the most straightforward setup.)

PR #90, opened 2023-01-30
Remove unneeded methods from Open API controller

gazayas:
It looks like we don't use the `attributes` method anymore, and I don't think we'll need it either now that we have the following in the `automatic_components_for` method:

```ruby
schema_json = jbuilder.json(
  model.new,
  title: I18n.t("#{model.name.underscore.pluralize}.label"),
  # TODO Improve this. We don't have a generic description for models we can use here.
  description: I18n.t("#{model.name.underscore.pluralize}.label"),
)

attributes_output = JSON.parse(schema_json)
```

For example, this will output the following:
```ruby
=> {"type"=>"object",
 "title"=>"Teams",
 "description"=>"Teams",
 "required"=>["id", "name"],
 "properties"=>
  {"id"=>{"type"=>"string", "description"=>"Team ID"},
   "name"=>{"type"=>"string", "description"=>"Team"},
   "time_zone"=>{"type"=>"string", "description"=>"Primary Time Zone"},
   "locale"=>{"type"=>"string", "description"=>"Language"},
   "created_at"=>{"type"=>"string", "description"=>"Signed Up At"},
   "updated_at"=>{"type"=>"string", "description"=>"Updated At"}},
 "example"=>{"id"=>nil, "name"=>nil, "time_zone"=>nil, "locale"=>nil, "created_at"=>nil, "updated_at"=>nil}}
```

## Failing tests
The tests are failing for some reason, but everything's passing locally even when I run `CI=true bundle exec rails test`, so I'm not sure why the tests aren't picking up on on the Platform constant:
```
NameError: uninitialized constant Platform
```


jagthedrummer:
@gazayas, it looks like maybe this isn't needed anymore? The current controller in `main` is now very slim: https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-api/app/controllers/api/open_api_controller.rb

gazayas:
@jagthedrummer @newstler Merged main and applied the same changes in the helper this time :+1:

PR #92, opened 2023-01-31
Remove comment about `platform_agent`

gazayas:
Forgive me if my understanding is incorrect on this one, but after looking through the code I found that we depend on both `platform_agent_of` and `platform_agent` in different places.

However, it's not an overwhelming amount so if we actually just want to remove the column from the database and depend solely on `platform_agent_of`, I'd be glad to undertake the task.

You can see both attributes being called in these scopes in the Membership model:

https://github.com/bullet-train-co/bullet_train-core/blob/fab77efab57126c083e0041f97c0e716560a2ffe/bullet_train/app/models/concerns/memberships/base.rb#L28-L32

Also, another place where `platform_agent` shows up:

https://github.com/bullet-train-co/bullet_train-core/blob/fab77efab57126c083e0041f97c0e716560a2ffe/bullet_train/app/models/concerns/memberships/base.rb#L66-L68

The connection workflow code is still new to me though, so I'll most likely look over it again soon to see if I can get a better understanding of it, and also to address the other TODO that's there.


andrewculver:
@gazayas Yes, let's cut over to `platform_agent_of` entirely.

gazayas:
Now that I'm working through this one, I found one caveat where I think we'll actually need `platform_agent`.

When we create a new Platform::Application, this creates a new Membership with `platform_agent_of_id`. However, when we delete the Platform::Application, it nullifies the Membership's `platform_agent_of_id`, so then we can have no way of telling if the Membership was ever tied to an application if we're going solely off that column.

This means any Platform::Applications that we've deleted will show up as tombstoned Memberships because the column is `nil`.

## A workaround
We could technically workaround this by checking the `user_email` of the record and seeing if it starts with `noreply+` and ends with `@bullettrain.co`. I'll see if I can somehow manage that into the tombstones scope.

gazayas:
Hmm, I'm having some trouble getting the scopes to work properly as well.
The platform_agents scope works fine, but the excluding_platform_agents scope I have written here isn't quite catching all of the records:

```ruby
scope :platform_agents, -> { where("platform_agent_of_id IS NOT NULL OR user_email ~ ?", "^noreply\+.*bullettrain\.co$") }
scope :excluding_platform_agents, -> { where.not("platform_agent_of_id IS NOT NULL OR user_email ~ ?", "^noreply\+.*bullettrain\.co$") }
```

By the way, this is what getting tombstoned and non-tombstoned Platform::Applications looks like:
```
irb(main):001:0> Membership.platform_agents
  Membership Load (1.2ms)  SELECT "memberships".* FROM "memberships" WHERE (platform_agent_of_id IS NOT NULL OR user_email ~ '^noreply+.*bullettrain.co$')
=>                                             
[#<Membership:0x00007ff8b318fea8               
  id: 3,                                       
  user_id: 2,                                  
  team_id: 1,
  created_at: Thu, 16 Feb 2023 11:20:47.444895000 UTC +00:00,
  updated_at: Thu, 16 Feb 2023 11:20:47.457578000 UTC +00:00,
  invitation_id: nil,
  user_first_name: nil,
  user_last_name: nil,
  user_profile_photo_id: nil,
  user_email: nil,
  added_by_id: nil,
  platform_agent_of_id: 1,
  role_ids: ["admin"]>,
 #<Membership:0x00007ff8a9cb1048
  id: 4,
  user_id: nil,
  team_id: 1,
  created_at: Thu, 16 Feb 2023 11:20:54.864382000 UTC +00:00,
  updated_at: Thu, 16 Feb 2023 11:21:01.692856000 UTC +00:00,
  invitation_id: nil,
  user_first_name: "Test Application",
  user_last_name: nil,
  user_profile_photo_id: nil,
  user_email: "noreply+51b4fa358c6ec4b1140b99f06b5078aa@bullettrain.co",
  added_by_id: nil,
  platform_agent_of_id: nil,
  role_ids: ["admin"]>]
```

andrewculver (approved):

andrewculver:
@gazayas OK! You identified the reason why we need this column! :-)

PR #94, opened 2023-02-01
Application-options.md url was missing a /

juliancheal:
Causing a Rails error on the documentation site.

The fix should make the url
https://bullettrain.co/docs/application-options.md
and not
https://bullettrain.co/docs/docs/application-options.md


andrewculver:
@juliancheal 🙏

PR #107, opened 2023-02-15
Add asterisk for required fields

### LINKED ISSUE #104: Light theme doesn't define a 'required' css class
The various form fields use a 'required' css class which the [Field partials](https://bullettrain.co/docs/field-partials) claims should apply a 'stylized asterisk', however in a fresh application I see the applied style, but no definition for it anywhere.

gazayas:
Closes #104.

## Details
The [field partial docs](https://bullettrain.co/docs/field-partials) claim that we "Display a stylized asterisk next to the label of fields that are known to be required," but we don't have an asterisk inside the `<div class="required">...</div>` tag.

This PR adds an asterisk for all `.required` labels.

I added the CSS to the `index` stylesheet since I believe it applies to all types of partials.

image


andrewculver:
My change looks like this:

Screen Shot 2023-02-15 at 5 20 26 PM

gazayas:
Nice 👍
Didn't realize you could apply the Tailwind classes like that.

PR #119, opened 2023-02-18
Add public layout

### LINKED ISSUE #110: Missing 'public' layout
When creating a controller that is not a devise controller the applied layout is 'public', but there is no default implementation of a public template in either the starter repo or in the light theme.

https://github.com/bullet-train-co/bullet_train-core/blob/f87143fbc2dd41a5e97b76be22971f95fb22cbca/bullet_train/app/controllers/concerns/controllers/base.rb#L50-L56

gazayas:
Closes #110.

## Details
Developers can override the `Public::HomeController` by commenting out `include RootRedirect` and adding an `index` action like it states here:

https://github.com/bullet-train-co/bullet_train/blob/82f42b75c7e58537658755d641c3e14261322586/app/controllers/public/home_controller.rb#L2-L4

However, we don't have a public layout or an index html action.

I didn't add `index` here because I figured the developer might want to make their own custom page/action, but I added layout with a message so developers don't get confused when the layout isn't there.


gazayas:
Opening as a draft because we're getting a new error in the starter repository unit tests.

gazayas:
Joint PR:
- https://github.com/bullet-train-co/bullet_train/pull/736

jagthedrummer (approved):

PR #122, opened 2023-02-20
Update doorkeeper token according to user time zone

gazayas:
I originally set out to fix this TODO:

```
# TODO Make this logic configurable so that downstream developers can write different methods for this column getting updated.
```

I mulled over it for a bit and I don't think I'm 100% sure I understand this TODO and how we want to make this part configurable (maybe by accepting `params[:time_zone]` and updating the column with that?), but I saw that we just use `Time.zone.now` to update the column, so I figured it might be good to update it according to the user's time zone if it exists.

If there's a certain way we want to make this configurable, I'd be glad to look over this section again sometime soon.

### DIFF
--- a/bullet_train-api/app/controllers/concerns/api/controllers/base.rb
+++ b/bullet_train-api/app/controllers/concerns/api/controllers/base.rb
@@ -75,7 +75,12 @@ module Api::Controllers::Base
     # TODO Make this logic configurable so that downstream developers can write different methods for this column getting updated.
     if doorkeeper_token
       begin
-        doorkeeper_token.update(last_used_at: Time.zone.now)
+        current_time = unless @current_user.time_zone.nil?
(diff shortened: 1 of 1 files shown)


jagthedrummer:
@gazayas I definitely don't understand what that TODO means, and it seems weird to me to change how that timestamp is set just based on whether on not the user has a timezone selected. It seems like we should always set it in one particular way (maybe UTC?) and then rely on the user's timezone to translate it for display. Would love to learn more about the original intent behind the TODO.

gazayas:
@jagthedrummer Agreed. I think we won't have to worry about this now that #334 has been merged, because the whole idea behind that PR was to unify everything on the back end with ISO8601 and handle all time zone differences on the front end.

I've gone ahead and simply deleted the TODO, but if there's anything you want me to look over I can take a second look.

PR #109, opened 2023-02-15
Adds multiple file upload

andrewculver:
@ammit Not sure why, but https://github.com/bullet-train-co/bullet_train-core/pull/10 was causing `rails test test/system/super_scaffolding_partial_test.rb` to fail after merging, so I reverted it. This PR was created by reverting the revert PR. :-) 

Example failure at https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/1507/workflows/2ab93a9a-1645-4b7d-92c6-21e7ed821dbc/jobs/6513

To run this test locally, you have to do some prep steps:

```
bundle add spring
bundle exec test/bin/setup-super-scaffolding-system-test
```

Once you've done that once you can run the following as many times as you like:

```
rails test test/system/super_scaffolding_partial_test.rb:36
```

PR #124, opened 2023-02-21
Introduce `Scaffolding::Attribute`

gazayas:
## Purpose
The purpose of this PR is to set up `bullet_train-super_scaffolding` to implement a `has_one` scaffolder smoothly.
(The issue to create a `has_one scaffolder` → https://github.com/bullet-train-co/bullet_train/issues/76)

## Classes to implement
For this to happen, I think it would benefit us to extract some logic from the Transformer and pinpoint where and how we want to build our association lines, the reason being that we Super Scaffold different types of associations to our models based on the attributes passed to `bin/super-scaffold`. I think where we build the line to eventually place in a model is an important design decision if we want to handle different types of associations dynamically, and I'm hoping to prevent some clutter by adding a couple of new classes.

I propose the following classes:
1. `Scaffolding::Attribute`
2. `Scaffolding::AssociationLineBuilder`

## Scaffolding::Attribute
As you can see with the first commit I've already added this, and there are a lot of diffs in the Transformer by changing `name` to `attribute.name`, etc. Although this is technically longer, I was able to cut a lot of the attribute setup at the beginning of the `add_attributes_to_various_views` method which I believe makes the code much easier to read. I also think the intent is clearer by using this class so whoever reads our code knows that `attribute` is the key part of this method.

Besides that, we can isolate a lot of the attribute-related logic here. What I'm specifically thinking of is how we declare associations. I was originally thinking of adding a class called `Scaffolding::Association` or a module named `Scaffolding::Association::Helper`, but the more I thought about it, `Scaffolding::Attribute` made the most sense because we don't know if an attribute is technically going to be an association to begin with.

If we can manage association declaration here, I think it will be easier to set up the `has_one` scaffolder when we get there.

## Scaffolding::AssociationLineBuilder
I feel like a lot of this logic will naturally be repeated when we write `has_many` or `belongs_to` lines when Super Scaffolding. I haven't quite fleshed this class out yet, but my plan is to work on this next and cut the PR there. Then we could work with these two classes to implement the `has_one` scaffolder.


gazayas:
I decided to mark the PR here as ready for the following reasons
1. I initially wanted to add both `Scaffolding::Attribute` and `Scaffolding::AssocationLineBuilder` in one PR, but I found out there was a decent amount that I was able to extract with just the Attribute class, so I plan on doing a separate PR just for the AssociationLineBuilder.
2. The conflicts in the transformer aren't actually so bad, but the tests on `main` are currently failing due to some translation issues, so once that's fixed I plan on merging `main`.

andrewculver:
@gazayas This looks awesome! Can you resolve the conflicts and then I'll merge it?

gazayas:
@andrewculver Done!

jagthedrummer:
@gazayas Looks like new conflicts have cropped up. I'd be happy to merge this once they're resolved.

jagthedrummer:
@gazayas ping (Sorry to ping you again, I'm just guessing this one fell through the cracks since you've been all over other PRs that I've asked about.)

gazayas:
Hey @jagthedrummer, I definitely have this one on my radar, it'll just take some careful incision to make sure it works properly so I've been prioritizing the other PRs like CI test failure-related ones. Will ping you when I make some progress!

jagthedrummer:
OK, that sounds great! Sorry to be a pest.

gazayas:
You're good 😃

gazayas:
@jagthedrummer Working on this one, having a hard time though because the super scaffolding tests are passing locally but not on CI for some reason. I know a lot has happened in CI so I hope it's not that, but I'll see what I can do.

gazayas:
@jagthedrummer Done! That one had my head spinning 😵‍💫

jagthedrummer:
@gazayas I'm so sorry I didn't merge this immediately last week, but now we have yet another conflict. 🤦

gazayas:
@jagthedrummer Conflicts resolved! Linking #109 since it was the one that had an affect on the latest merge.

However, the Super Scaffolding partial tests are failing, and it's failing for me locally too on `main`? Not sure what's going on here, will try to investigate.

gazayas:
Note to self: The partial tests are working in isolation, but fail when running the entire test setup script.

gazayas:
I see what's happening, I needed to call this guy in the Attribute class!

スクリーンショット 2023-08-25 20 13 55


THAT was confusing!!

jagthedrummer (approved):
🥳

gazayas:
:partying_face: :sob:

PR #136, opened 2023-03-01
Extract check boxes and radio buttons to `option_tag`

gazayas:
## Details
This PR seeks to place all of the same logic we implement for check boxes and radio buttons into one place. For example, here are a few things that overlap concerning different check box methods and the `radio_button` method:
1. `value`: The value of the check box or radio button we're making
2. `class`: Where we put tailwind styles, etc.
3. `multiple`: I added this as a keyword argument so developers can add this if necessary or skip it if they just want a radio button.

## Another `options` partial
I wanted to make this a partial at first, but we already have the `_options.html.erb` partial, and I thought it would be more intuitive to use `<%= options_tag() %>` since this is similar to writing `<%= form.check_box() %>` or `<%= check_box_tag() %>`.

I'm not entirely sure if we can call this "polymorphic," but I've written the comments as such. I can re-write those too if necessary.

I'm opening this as a draft pull request because there are some other places where I'd like to use `options_tag`.

### DIFF
--- a/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
+++ b/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
@@ -76,18 +76,8 @@ end
 
           <label class="relative flex items-start mb-3">
             <div class="flex items-center h-5">
-
-              <% if multiple %>
-                <% checked_value = form.object.send(method).nil? ? nil : form.object.send(method).map(&:to_s).include?(value.to_s) %>
-                <%= form.check_box method, {
-                  multiple: multiple, checked: checked_value,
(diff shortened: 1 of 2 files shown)


gazayas:
By the way, the tests are failing due to a Ruby version error, but Super Scaffolding check boxes and radio buttons is working for me here locally.

gazayas:
Okay, now that I'm thinking through it more, we could combine this helper with a new partial
named `_option.html.erb` (singular for each individual tag, as opposed to the plural `_options.html.erb` which we already have). Then we can put labels and other things inside that view.

gazayas:
Closing in favor of #148.

I think I over-complicated it with the helper here, but it helped me organize my thoughts.

Issue #140, opened 2023-03-01
Adding `showcase-rails` causes `uninitialized constant ApplicationHelper`.


andrewculver:
Started seeing this after https://github.com/bullet-train-co/bullet_train-core/pull/46. Wasn't able to fix it. Also worth mentioning that I did https://github.com/bullet-train-co/bullet_train-core/pull/139 as well.

You can see this failures in CI at https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train-core/385/workflows/bf67c40c-a284-478d-be33-f055797a04b0/jobs/1169.

To reproduce, clone http://github.com/bullet-train-co/bullet_train and add `gem "showcase-rails"` to the `Gemfile`.

For what it's worth, I actually want this to be a fixed dependency of the `bullet_train` Ruby gem, not the starter repo. :-)

kaspth:
> For what it's worth, I actually want this to be a fixed dependency of the `bullet_train` Ruby gem, not the starter repo. :-)

Cool, I'll make that happen. I made it optional to give us leeway to flip it later, but we can go for this now.

PR #142, opened 2023-03-01
Make Showcase a Bullet Train dependency + updates 🎉

kaspth:
Fixes https://github.com/bullet-train-co/bullet_train-core/issues/140 on the Bullet Train side.


bullet_train/app/views/showcase/engine/_stylesheets.html.erb
```diff
@@ -1 +1 @@
-<%= stylesheet_link_tag "application", "application.light", "showcase" %>
+<%= stylesheet_link_tag "application", "application.light", "showcase", "showcase.highlights" %>
```
kaspth, line 1: I'm gonna fix this up so client apps don't need to have control over including the showcase and showcase.highlights assets.

PR #143, opened 2023-03-01
Run Showcase's tests for bullet_train-themes-light

kaspth:


kaspth:
I added Gemfiles for each gem, since GitHub Actions wanted us to run `bundle lock --add-platform x86_64-linux`.

kaspth:
Started out as trying to get tests to run for just bullet_train-themes-light, and then it was redoing a big chunk of the CI setup, woop woop.

PR #144, opened 2023-03-02
Limit Showcase mount to just development

kaspth:
Since Showcase is now a hard BulletTrain dependency, and since gemspec dependencies can't specify an environment, we'd up mounting the Showcase routes in production. We shouldn't do that.

Due to a version of Showcase's tests using integration tests, we'd also mount the routes in the test environment, which was the reason for using the defined? check.

We can now just use development?.


kaspth:
@andrewculver ah looks like you already found this — I figured out we can simplify the check to just `development?` now.

andrewculver:
@kaspth What about tests? I want people to be able to write integration tests against their components here if needed.

kaspth:
@andrewculver we do have some view level test setup https://github.com/bullet-train-co/showcase#automatic-previews-testing, but true, we can skip this so it's easier to write system ones too.

PR #148, opened 2023-03-02
Introduce `option` partial

gazayas:
This partial houses all of our options logic so we can handle single check boxes, multiple check boxes, and radio buttons all in one place.


gazayas:
If I can get the action models check boxes to work, then I'll mark this one as ready.

gazayas:
There's still the `check_box_tag` at line 43 in the `_options.html.erb` partial to migration to the new partial, but I haven't quite found where this part of the code is rendering yet. Besides that, unit tests, system tests, and the super scaffolding system tests are passing for me.

gazayas:
Okay, this one is ready to go.

I added a comment about bulk selection for multiple check boxes. Basically, we'd need the `Select Multiple` button and logic that exists in Action Models to exists here as well. We could just have the partial live here in `bullet_train-core`, but since it's a private repository I just decided to add a TODO. I can take care of that in a later PR if needed.

gazayas:
https://user-images.githubusercontent.com/10546292/224020860-0df35de7-3f45-49f1-84f3-d8ced545b72a.mov

gazayas:
Ah, meant to add that last commit to #171, doesn't make a difference either way though.

PR #141, opened 2023-03-01
Use `Gem.find_files` in `bt:link` to avoid loading the environment

kaspth:
A potential simplification for our `bt:link` task.

PR #154, opened 2023-03-05
Use Rake directly to lookup the db:seed task

kaspth:
Rake already ships an interface to check tasks, so we can just reuse that.


kaspth:
Seems like the reactivity build failure is unrelated, it happens on #141 too.

PR #157, opened 2023-03-05
Update nice partials to 0.9

kaspth:
Also updates our internal partials to the nice_partials method based syntax.


kaspth:
This is just to start an internal upgrade and see these tests pass upfront as an extra confidence check.

andrewculver (approved):

PR #159, opened 2023-03-05
Allow field partials to use Nice Partials with backwardscompatible fallback

kaspth:


kaspth:
Hm, not sure what's going on here. May be better to leave these as is.

jagthedrummer:
@kaspth What are your thoughts on this now? Is it worth fixing the conflicts and getting tests passing, or should we just close it?

kaspth:
@jagthedrummer oh yeah, I think this would actually be great. I'll try to split these changes out over more PRs targeting fewer partials at a time so it's a little easier to tackle the test failures.

jagthedrummer:
Splitting it into smaller PRs sounds great!

kaspth:
Ok, I'd already spun off a bunch of this work into #157, so I've just rebased on main and then we'll see what makes sense here.

PR #160, opened 2023-03-06
Tweak CI: use newest Ruby orb

kaspth:
Starts us down the path of hopefully having better, easier Bundler caching.


kaspth:
Although our build times fluctuate quite a bit, it does seem like I managed to shave off ~10 seconds. Let's see if using CircleCIs orbs will eventually speed up our builds as new gem versions are used.

PR #161, opened 2023-03-06
Add a recursive method to test if a user can perform a role

adampal:
Users can be assigned roles "through" another role in the yml file. In our default setup, admin users have all viewer permissions via:
```
admin:
  includes:
    - viewer
 ```
However if you want to check for the presence of the viewer role, this doesn't work:
```
membership.roles << Role.admin
membership.roles.include?(Role.find("viewer"))
-> false
```

because it only checks roles directly add to the membership.

Rather than mess with the existing `include?` method, I'm proposing we add a new method called `can_perform_role?` that will check for roles that have been assigned directly and also roles that are included in any assigned roles.

So now with our default setup, this will work:

```
membership.roles << Role.admin
membership.can_perform_role?(Role.find("viewer"))
-> true
```

The code here should work but isn't fully tested yet.
Looking for any thoughts or feedback before I add tests and finish it off.


adampal:
Ok, now fully tested and documented in the Readme.
This should be ready to merge...

adampal:
Sorry... _now_ it includes tests!

jagthedrummer (approved):
Look great!

PR #162, opened 2023-03-06
Skip updating for content that already exists and has been transformed

gazayas:
In https://github.com/bullet-train-pro/bullet_train-action_models/pull/46 we were trying to remove a `has_one` association from a model like this:

```ruby
scaffold_replace_line_in_file(
  "./app/models/scaffolding/completely_concrete/tangible_things/foo.rb",
  "",
  "has_one :team, through: :team",
)
```

`""` becomes a variable called `content`, and since we have the following `if` statement in the [FileManipulator](https://github.com/bullet-train-co/bullet_train-core/blob/0680b1f48fd32b88648a189c43a88109d324df36/bullet_train-super_scaffolding/lib/scaffolding/file_manipulator.rb#L26), it's skipping the file update altogether:
```ruby
if target_file_content.include?(content)
  puts "No need to update '#{file}'. It already has '#{content}'."
else
  puts "Updating '#{file}'." unless silence_logs?
  target_file_content.gsub!(in_place_of, content)
  File.write(file, target_file_content)
  end
end
```

We should only really be doing a check like this for content that has been transformed (i.e. - when we have a model like `Foo` and we use `TangibleThingsController` to change it to `FoosController`), and not all other general strings, so I've ensured we only skip updating when the content is transformed content and already exists in the target file.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/file_manipulator.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/file_manipulator.rb
@@ -23,7 +23,7 @@ module Scaffolding::FileManipulator
       return false
     end
 
-    if target_file_content.include?(content)
+    if options[:content_replacement_transformed] && target_file_content.include?(content)
       puts "No need to update '#{file}'. It already has '#{content}'."
     else
       puts "Updating '#{file}'." unless silence_logs?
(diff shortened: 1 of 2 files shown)


gazayas:
Thinking over this one for a bit, I could also see us adding a `force: true` option to `Scaffolding::FileManipulator.replace_line_in_file` in case we don't want this logic to be specific to transformed content.

PR #163, opened 2023-03-06
Run bullet_train-roles gem tests on CI

kaspth:
cc @adampal


kaspth:
@adampal ok, it was a little more difficult to get this working than I expected, but we're running the roles tests on CI now.

PR #165, opened 2023-03-07
Add documentation for custom `date` and `date_and_time` formats

gazayas:
Context: https://discord.com/channels/836637622432170028/1080497842747092992

We have the logic in place to set custom formats, just no docs explaining how to implement it.


kitebuggy:
Thanks for this, great work. This will help for overruling defaults (e.g. for one-off requirements), but it's not be ideal for mass changes.  I've tried to explain my reasoning in the discord channel and I'd welcome your thoughts.  Again, thanks for documenting this though, which will make it useful for the edge cases.

gazayas:
I'm thinking about using `Time#to_fs` so we can lean on the Rails standard and then we can put custom localizations in the starter repo.

PR #171, opened 2023-03-08
Show a list of the field partials with `bin/super-scaffold`

gazayas:
I tend to forget which data types should be used for our field partials, so I added a flag to list all the information via the terminal:

```
> bin/super-scaffold --field-partials
```

## The result

image


gazayas:
We also had a minor bug with the `--help` flag where we were marking it as an invalid type.

image

gazayas:
I have fixed that in this PR as well.

jagthedrummer (approved):

PR #182, opened 2023-03-11
Standardize `Date` and `DateTime` locales

gazayas:
Currently, we have our own in-house formats for `Date` and `DateTime` objects written as String literals in the dates helper.

@kitebuggy brought up a good point, and we should standardize these objects so they fall back on Rails' standard I18n.

We unfortunately lose the "Today at ..." and "Yesterday at ..." strings we once had, but I think this is a good trade off because we can then write our partials like this:

```erb
<%= render 'shared/attributes/date', attribute: :test_date %> <!-- Uses default format -->
OR
<%= render 'shared/attributes/date', attribute: :test_date, format: :short %>
OR
<%= render 'shared/attributes/date', attribute: :test_date, format: '%m/%d/%Y' %>
```

That way it will make it much easier for developers to lean on the Rails standard way of handling formats, and they can write yaml like the following in their own application without having to eject any files from bt core:

```yml
en:
  time:
    formats:
      long: # custom format here...
```

Here are the default [date formats](https://github.com/svenfuchs/rails-i18n/blob/3c1f34c29fbdc2f8f3a3516cd05d406bf8aaf9cf/rails/locale/en-US.yml#L41-L44) and [time formats](https://github.com/svenfuchs/rails-i18n/blob/3c1f34c29fbdc2f8f3a3516cd05d406bf8aaf9cf/rails/locale/en-US.yml#L211-L214) from the `rails-i18n` gem.


bullet_train/app/helpers/account/dates_helper.rb
```diff
@@ -1,40 +1,24 @@
...
   end
 
-  def local_time(time)
-    return time if current_user.time_zone.nil?
-    time.in_time_zone(current_user.time_zone)
+  def set_to_user_time_zone(timestamp)
+    return timestamp if current_user.time_zone.nil?
+    is_date = timestamp.is_a?(Date)
+    timestamp = timestamp.in_time_zone(current_user.time_zone)
+    is_date ? timestamp.to_date : timestamp
```
kaspth, line 22: I think we could shorten this down to:

```suggestion
    timestamp&.in_time_zone(current_user.time_zone || Time.zone)
```

Rails enhances `Date`  to respond to `in_time_zone` too — I'm always confused with time zone math, but I think it makes more sense to cast dates too.

bullet_train/app/helpers/account/dates_helper.rb
```diff
@@ -1,40 +1,24 @@
...
-  def display_time(timestamp, custom_time_format = nil)
-    local_time(timestamp).strftime(custom_time_format || "%l:%M %p")
+  # If the format is a Symbol, we use the standard Rails formatting.
+  # https://guides.rubyonrails.org/i18n.html#adding-date-time-formats
+  def formatted_timestamp(timestamp, format: nil)
+    format ||= :default
+    format.is_a?(Symbol) ? l(timestamp, format: format) : timestamp.strftime(format)
   end
 
-  def local_time(time)
```
kaspth, line 36: Since this is a public method is there any chance that people are relying on the old method name? I.e. can we just yank it?

bullet_train/app/helpers/account/dates_helper.rb
```diff
@@ -1,40 +1,24 @@
...
 
-  # e.g. 4:22 PM
-  def display_time(timestamp, custom_time_format = nil)
-    local_time(timestamp).strftime(custom_time_format || "%l:%M %p")
+  # If the format is a Symbol, we use the standard Rails formatting.
+  # https://guides.rubyonrails.org/i18n.html#adding-date-time-formats
+  def formatted_timestamp(timestamp, format: nil)
+    format ||= :default
+    format.is_a?(Symbol) ? l(timestamp, format: format) : timestamp.strftime(format)
   end
```
kaspth, line 16: It looks like `I18n.localize`, which Rails delegates to, already handles the Symbol lookup but falls back to a format string based to `strftime`. Have we tried writing this like this?

```suggestion
  def formatted_timestamp(timestamp, format: :default)
    localize(timestamp, format: format)
  end
```

However, if this works, at that point we could also inline the method.

kaspth: Thinking more about this, I wonder if we want these helpers to just be pretty close to this:

```ruby
module Account::DatesHelper 
  def display_date(timestamp, ...)
    localize(local_time(timestamp.to_date), ...) if timestamp
  end

  def display_time(timestamp, ...)
    localize(local_time(timestamp.to_time), ...) if timestamp
  end

  def display_datetime(timestamp, ...)
    localize(local_time(timestamp.to_datetime), ...) if timestamp
  end

  def local_time(timestamp)
    timestamp&.in_time_zone(current_user.time_zone || Time.zone)
  end
end
```

gazayas: Yes, this makes sense to me and is much cleaner to look at. I think I got bogged down in trying to use what was already there, but this is easier to read. I committed what you wrote, but I have an idea to take it a step further by putting all of the logic into one method which I'll mention in a new comment below.

gazayas: Oof, wasn't aware that it would fallback to `strtime`, just passing `format` will definitely work then. As far as the `format: :default` keyword argument, I originally tried writing it the way you mentioned here, but I eventually wrote `format ||= :default` here because `format` kept returning `nil` for me:

image

I found out it was basically because if we explicitly pass `nil` to the method, the default keyword argument just gets overwritten with `nil`. This was happening because I was calling `display_date(timestamp, format: local_assigns[:format])` in the view which was passing `nil` if we don't give it a value when rendering the view.
```
irb(main):001:1* def foo(bar: true)
irb(main):002:1*   puts "Value: #{bar}"
irb(main):003:0> end
=> :foo
irb(main):004:0> foo
Value: true
=> nil                                             
irb(main):005:0> foo(bar: nil)
Value: 
=> nil
```

The keyword arguments should be working now!

bullet_train/app/helpers/account/dates_helper.rb
gazayas: Yeah you're 100% right, I added the method back in case other developers are using it.

gazayas:
I propose the following to make our code DRY. If anything, we can just stick with the latest commits I pushed.

```ruby
module Account::DatesHelper
  def display_date(timestamp, format: :default)
    return nil unless timestamp
    type = timestamp.class.to_s_downcase
    localize(local_time(timestamp).send("to_#{type.to_s.downcase}".to_sym), format: format)
  end

  # We can delete these if we want to make the original method name more broad.
  # I guess it depends on if developers depend on these helpers or not,
  # or if they should just be Bullet Train internals.
  alias_method :display_date_and_time, :display_date
  alias_method :display_time, :display_date

  def local_time(timestamp)
    timestamp&.in_time_zone(current_user.time_zone || Time.zone)
  end
end
```

We'd have to account for `ActiveSupport::TimeWithZone` though since that's what `created_at` attributes are, and these attributes currently use `date_and_time.html.erb` partials ([example here](https://github.com/bullet-train-co/bullet_train-core/blob/7b6aa6e3dada6f7fb2fbb54e60cbb21772defe61/bullet_train-super_scaffolding/app/views/account/scaffolding/completely_concrete/tangible_things/_tangible_thing.html.erb#L17)), which means they use the `display_date_and_time` method as opposed to `display_time`.

We could add a hook like this for the time being:
```ruby
type = timestamp.class == ActiveSupport::TimeWithZone ? "time" : timestamp.class.to_s.downcase
```

However, I wonder if we should just add a `Time` partial and use `display_time` there.

I'll update the tests in the starter repository once we decide for sure on this part.

bullet_train/app/helpers/account/dates_helper.rb
kitebuggy: I think I spot a typo there @gazayas, shouldn't that be:
```
[1] pry(main)> def foo( bar = true)
[1] pry(main)*   puts "Value: #{bar}"
[1] pry(main)* end  
=> :foo
[2] pry(main)> foo
Value: true
=> nil
[3] pry(main)> 
```

kitebuggy: @kaspth Of course, the even shorter alias for localize is just 'l' (similar to translate and 't'), making this even neater still (though arguably harder to read?)

kitebuggy: I would suggest we avoid defaulting to current_user.time_zone in these.  If you look at the code for [current_locale_method](https://github.com/bullet-train-co/bullet_train-core/blob/7b6aa6e3dada6f7fb2fbb54e60cbb21772defe61/bullet_train/app/helpers/account/locale_helper.rb#L2), we check User > Team > default locale.  Unfortunately, that is hardcoded, which is shame (would recommend I18n.default_locale, as per Rails conventions).

However, the concept of the cascading priorities would seem to apply here as well, so perhaps:
```
  def current_time_zone
    current_user.time_zone || current_team.time_zone || Time.zone
  end
```

Would you agree?

kaspth: @gazayas yeah, that's an issue sometimes. I'd write them as `localize(local_time(timestamp.to_x), format: format || :default) if timestamp` then.

@kitebuggy I tend to prefer `l` for app code in ERB and then `localize` for lib code, but it's just personal preference really.

From what I can tell, `current_time_zone` sounds like it fits in this context and is interchangeable with our `current_user.time_zone || Time.zone` handling.

kaspth:
> I propose the following to make our code DRY

I think it's actually the reverse, by deriving the message to send and then needing `alias_method` elsewhere to make it work, it means we're repeating the logic in two places. I also don't think the extra conceptual complexity is worth it for what we get out of it, the condensed method is more complex to read.

> We'd have to account for `ActiveSupport::TimeWithZone` though since that's what `created_at` attributes are, and these attributes currently use `date_and_time.html.erb` partials ([example here](https://github.com/bullet-train-co/bullet_train-core/blob/7b6aa6e3dada6f7fb2fbb54e60cbb21772defe61/bullet_train-super_scaffolding/app/views/account/scaffolding/completely_concrete/tangible_things/_tangible_thing.html.erb#L17)), which means they use the `display_date_and_time` method as opposed to `display_time`.

I don't understand the issue here? What key does `ActiveSupport::TimeWithZone` lookup in the I18n formats?

I don't think burying a hook for `ActiveSupport::TimeWithZone` is a good fit here, it sounds like we'd need to keep `display_date_and_time` agnostic and that it maybe needs to read closer to the original version before this PR (versus my sample thing).

gazayas:
> I also don't think the extra conceptual complexity is worth it for what we get out of it, the condensed method is more complex to read.

I figured that would be the case, I'll keep the file as is then.

> I don't understand the issue here? What key does ActiveSupport::TimeWithZone lookup in the I18n formats?

This is only relevant if we use the `send("to_#{type}")` code I mentioned, since we would get the following error:
```
ActionView::Template::Error: undefined method `to_activesupport::timewithzone'
```

@kitebuggy I added `current_time_zone` 👍

Tests are done, marking this as ready!
- https://github.com/bullet-train-co/bullet_train/pull/696

bullet_train/app/helpers/account/dates_helper.rb
```diff
@@ -1,40 +1,21 @@
...
-  def display_date(timestamp, custom_date_format = nil)
-    return nil unless timestamp
-    if custom_date_format
-      local_time(timestamp).strftime(custom_date_format)
-    elsif local_time(timestamp).year == local_time(Time.now).year
-      local_time(timestamp).strftime("%B %-d")
-    else
-      local_time(timestamp).strftime("%B %-d, %Y")
-    end
+  def display_date(timestamp, format: :defaul)
```
kitebuggy, line 2: Typo: **_:defaul_** should be :default?

kitebuggy:
Test failing due to lack of "Today at ..." and "Yesterday at ..." strings, which where hard coded in display_date_and_time.

May I suggest adding back in the switching logic, but setting the format to a locale defined :today or :yesterday format for these use cases?  This can then be set for the appropriate language strings and formats in a more flexible manner, with the default being something like `today: "Today at %l:%M %p"` in the YAML file.

kaspth:
@gazayas cool, makes sense!

> This can then be set for the appropriate language strings and formats in a more flexible manner, with the default being something like `today: "Today at %l:%M %p"` in the YAML file.

If we're going for this, I don't think hardcoding the date/time format into here is a good default.

I think there could be a version sorta like this:

```
en:
  day_offsets:
    -1: Yesterday
    0: Today
    1: Tomorrow
```

which we could generate with `(timestamp.to_date - Time.zone.today).to_i`. Though I find date math really hard, so we could revisit this another time.

bullet_train-themes/app/views/themes/base/attributes/_date.html.erb
```diff
@@ -1,11 +1,12 @@
 <% object ||= current_attributes_object %>
 <% strategy ||= current_attributes_strategy || :none %>
 <% url ||= nil %>
+<% local_assigns[:format] ||= :default %>
```
kaspth, line 4: iirc, `local_assigns` isn't meant to be mutated, plus we can use the same format as the other attributes:

```suggestion
<% format ||= :default %>
```

bullet_train-themes/app/views/themes/base/attributes/_date.html.erb
```diff
@@ -1,11 +1,12 @@
...
 <% default_message = local_assigns[:default_message] || t('global.formats.timestamp_unavailable') %>
 
 <% if object.send(attribute).present? %>
   <%= render 'shared/attributes/attribute', object: object, attribute: attribute, strategy: strategy, url: url do %>
-    <%= display_date(object.send(attribute), local_assigns[:date_format]) %>
+    <%= display_date(object.send(attribute), format: local_assigns[:format]) %>
```
kaspth, line 9: I wonder if we need the `nil` checks / `||= :default` if we can do this:

```suggestion
    <%= display_date(object.send(attribute), **local_assigns.slice(:format)) %>
```

kaspth (approved):

bullet_train/app/helpers/account/dates_helper.rb
gazayas: Just noticed this, about to push changes here in a bit

bullet_train-themes/app/views/themes/base/attributes/_date.html.erb
gazayas: > I wonder if we need the `nil` checks / `||= :default` if we can do this:

Done!

gazayas:
I pushed some more commits, but I ran into another error with daterangepicker.

When initially creating a record, the `Date` and `DateTime` attributes aren't populated yet, so daterangepicker formats the value for us (see [date_controller.js](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-fields/app/javascript/controllers/fields/date_controller.js)) and the record is created without a problem.

However, when we go to edit it, daterangepicker tries to use the format we declare (or just the default) to update the calendar:

Date Issue 1

Date Issue 2

I fixed this by writing String literals in the field partials to format the values, but I'd really like to use Rails defaults for this one too. If it's not high-priority, I can leave a TODO. Besides the form, everything else works as expected.

gazayas:
If #334 is merged we can probably go ahead and close this one.

jagthedrummer:
@gazayas Just wanted to confirm that you're saying we should close this PR in favor of https://github.com/bullet-train-co/bullet_train-core/pull/334?

gazayas:
@jagthedrummer Okay, this one should be good to go. After merging `main`, you can see that #334 is left untouched, and the default rails formats show up on the show page:

Screenshot from 2023-08-04 12-59-14

## Tests
Also the starter repository minitest should pass after https://github.com/bullet-train-co/bullet_train/pull/696 is merged.

jagthedrummer:
@gazayas There are some conflicts on this one now. 😞

If you can push the branch from https://github.com/bullet-train-co/bullet_train/pull/696 into the main starter repo (instead of your own fork) then you should be able to rebase/merge `main` on this PR and then (hopefully) the tests should pass.

gazayas:
@jagthedrummer Conflicts resolved! With the merge from main, I'm seeing that it had an affect on https://github.com/bullet-train-co/bullet_train-themes/pull/6, and if developers try to use a custom format like this now, it won't break the view but it will just use the `:default` format:

```erb
<%= render 'shared/attributes/date', attribute: :date_test, date_format: "%m/%d" %>
```

However, they CAN write just `format` to customize it:
```erb
<%= render 'shared/attributes/date', attribute: :date_test, format: "%m/%d" %>
```

I've updated the docs to reflect this, and I think we'll be okay with the changes, but I'll perhaps write a deprecation or something in case we need it.

kaspth (approved):

jagthedrummer:
@gazayas I like the idea of a deprecation message, but I think that it should probably also handle both formats for at least a while. Otherwise it's not exactly a deprecation message, but an after-the-fact warning about a breaking change that's already subtly breaking things.

gazayas:
> @gazayas I like the idea of a deprecation message, but I think that it should probably also handle both formats for at least a while, if possible. Otherwise it's not exactly a deprecation message, but an after-the-fact warning about a breaking change that's already subtly breaking things.

@jagthedrummer Good point, I updated the code so it will handle both formats. You can see details in the docs that I updated here as well.

jagthedrummer:
@gazayas what do you think we should do with this now that we reverted https://github.com/bullet-train-co/bullet_train-core/pull/334? Should we just let it hang out until after we can get revert the revert of #334? If so, this is the revert of the revert PR: https://github.com/bullet-train-co/bullet_train-core/pull/390

jagthedrummer:
I'm good with merging this before waiting on the others, but I think that reverting #334 has now made it so that there are legit test failures in this branch. I pulled the branch down and tried to rebase it but there was going to be a lot conflicts to resolve. Surprisingly a `git merge main` went off without a hitch, so I pushed that up, but there are still failures.

If we can get those sorted out, and if @gazayas is still good with it I'm OK with merging this.

gazayas:
@jagthedrummer Ok, we're set.

Since `current_time_zone` was from #334, I replaced this with `current_user.time_zone` in `DatesHelper#local_time`. I also added back a couple of locales that had been removed there.

PR #202, opened 2023-03-18
Raise `class_name`/`vanilla` error correctly for namespaced models

gazayas:
Closes https://github.com/bullet-train-co/bullet_train/issues/701.

## Details
When Super Scaffolding `_id` or `_ids` attributes, we look to see if the model exists with the following [here](https://github.com/bullet-train-co/bullet_train-core/blob/b2c0bd2762286a80bb4bd585df21c9ff56bf504d/bullet_train-super_scaffolding/lib/scaffolding/script.rb#L70-L82):

```ruby
file_name = "app/models/#{attribute_options[:class_name].underscore}.rb"
unless File.exist?(file_name)
  # Raise error that model doesn't exist.
end
```

The problem is, when the model is namespaced, the parent model file exists, it's just not an `ApplicationRecord`:

```ruby
module Asdf
  def self.table_name_prefix
    "asdf_"
  end
end
```

## The Fix
Since we were just checking for file existence, I added an extra check to make sure that the file that exists is not a module to prefix table names.

### Example
```
bin/rails g model Asdf::Foo::DevTeam team:references name:string asdf_id:uuid
bin/super-scaffold crud Asdf::Foo::DevTeam Team name:text_field asdf_id:text_field
```

In this case, we look up `asdf.rb` because of the `asdf_id:uuid` attribute, but the file doesn't exist because the directory hierarchy looks like this:
```
app/
└── models/
    └── asdf/
        ├── foo.rb
        └── foo/
            └── dev_team.rb
```

This is technically okay because we want to raise the error if the model doesn't exist, I just thought I'd bring this up in case we want to look up the models in a different way.

This is also another reason why I changed the `file_name` declaration. Now we're searching the entire model's directory to see if the record exists so we can be sure we found the file if it's there.


gazayas:
I'm getting an error about `bin/checkout-and-link-starter-repo`, I'm going to have to look into this one again soon...

スクリーンショット 2023-03-18 12 25 39

bullet_train-super_scaffolding/lib/scaffolding/script.rb
```diff
@@ -67,8 +67,18 @@ def check_required_options_for_attributes(scaffolding_type, attributes, child, p
...
+        file_name = Dir.glob("app/models/**/*.rb").find { |model| model.match?(/#{attribute_options[:class_name].underscore}\.rb/) } || ""
+
+        # If a model is namespaced, the parent's model file might exist under
+        # `app/models/`, but sometimes these files are modules that resolve
+        # table names by providing a prefix as opposed to an actual ApplicationRecord.
+        # This check ensures that the model really doesn't exist even if the file does.
+        is_table_prefix = if File.exist?(file_name)
+          lines = File.open(file_name).readlines
+          Scaffolding::FileManipulator.find(lines, "self.table_name_prefix")
+        end
```
jagthedrummer, line 79: I suspect that this method could still be problematic. For instance maybe `asdf.rb` exists, but _doesn't_ contain a `table_name_prefix` line and is just an empty module. It seems like we should care more about whether the class name points to a concrete class vs pointing to a module.

Would something like this work:

```ruby
class_name_points_to_a_class = attribute_options[:class_name].constantize.is_a?(Class)
```
or maybe this if we want to ensure that it's an `ActiveRecord`:

```ruby
class_name_points_to_active_record = attribute_options[:class_name].
  constantize.ancestors.include?(ActiveRecord::Base)
```

jagthedrummer (commented):
@gazayas just one question about this. Apologies if I'm missing some context that makes it a dumb question.

bullet_train-super_scaffolding/lib/scaffolding/script.rb
gazayas: @jagthedrummer Not a dumb question at all! I was working under the assumption that the example provided in https://github.com/bullet-train-co/bullet_train/issues/701 could work when `Asdf` is simply a namespace and not a model, but now that I'm looking over the issue again, I'm seeing that it presumes the existence of `Asdf` as a MODEL which is why we want to declare `asfd_id` in that super scaffolding command in first place.

Checking for the class this way makes sense, so I'll see if I can look over the code soon and push some changes that make more sense.

gazayas: I might have to think through this one a bit though, because looking at the error message provided by the `DevTeam` example, we assume that we can accomplish this WITHOUT and ActiveRecord association:

```
If `asdf_id` is just a regular field and isn't backed by an ActiveRecord association, you can skip all this with the `{vanilla}` option, e.g.:

  bin/super-scaffold crud DevTeam Team asdf_id:text_field{vanilla}
```

Will ponder on this a bit

gazayas:
@jagthedrummer Closing this one since it looks like it was solved (details in https://github.com/bullet-train-co/bullet_train/issues/701)

gazayas:
False alarm, we DO need this PR, working on a fix and an explanation now.

bullet_train-super_scaffolding/lib/scaffolding/script.rb
gazayas: @jagthedrummer After working through it mentally, we won't need anything else besides the fix here, and using `*_id` as a regular field without being backed up by an ActiveRecord model will work just fine if we pass `{vanilla}`. Also, just checking for the class should be enough to ensure we display the error message at the proper timing, so I used the logic you suggested to accomplish this.

This PR should be good to go.

jagthedrummer (approved):

PR #206, opened 2023-03-21
Add localizations for file fields

### LINKED ISSUE #204: Text in `file_field` partials should use I18n translations
https://github.com/bullet-train-co/bullet_train-core/blob/b2c0bd2762286a80bb4bd585df21c9ff56bf504d/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_file_field.html.erb#L9-L40

and

https://github.com/bullet-train-co/bullet_train-core/blob/b2c0bd2762286a80bb4bd585df21c9ff56bf504d/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/attributes/_file.erb#L7-L10

Currently, text like `Download Current Document` is hard-coded in the template which makes it necessary to eject this just to change the wording.

gazayas:
Closes #204.

The only file field localization that isn't covered here is `Select Another File`:

https://github.com/bullet-train-co/bullet_train-core/blob/b2c0bd2762286a80bb4bd585df21c9ff56bf504d/bullet_train-fields/app/javascript/controllers/fields/file_field_controller.js#L88

We used [i18n-js](https://github.com/fnando/i18n-js) in the past a while back before the Tailwind repository, so if we  want to implement that again I'd be glad to work on it.

### DIFF
--- a/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
+++ b/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
@@ -50,7 +50,7 @@ end
             }, class: "focus:ring-blue h-4 w-4 text-blue border-slate-300 rounded" %>
           </div>
           <div class="ml-2.5 text-sm pr-4">
-            <%# TODO: should be localized %>
+            <%= t("global.bulk_select.all") %>
             All
           </div>
         </label>
--- a/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
+++ b/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
@@ -51,7 +51,6 @@ end
           </div>
           <div class="ml-2.5 text-sm pr-4">
             <%= t("global.bulk_select.all") %>
-            All
           </div>
         </label>
       <% end %>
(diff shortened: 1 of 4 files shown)


gazayas:
Looking through i18n-js and how to set it up, and it looks like we'd have to add a config file to the starter repository along with some other things, so if we DO want to add it, I think I'll tackle it in another PR.

bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
```diff
@@ -50,7 +50,7 @@ end
             }, class: "focus:ring-blue h-4 w-4 text-blue border-slate-300 rounded" %>
           </div>
           <div class="ml-2.5 text-sm pr-4">
-            <%# TODO: should be localized %>
+            <%= t("global.bulk_select.all") %>
             All
```
jagthedrummer, line 54: I think we need to remove the hard coded `All` on line 54 here.

jagthedrummer (changes requested):
One more string needs to be removed, but looks great otherwise.

bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_options.html.erb
gazayas: :man_facepalming: Just deleted it!

jagthedrummer (approved):

PR #207, opened 2023-03-22
Fix Trix Editor styling

### LINKED ISSUE #150: Trix Editor HTML listings not displaying properly
Bullet point and ordered lists are not displaying properly when editing them through the Trix Editor nor when displaying them in the current theme.  To reproduce, go to the online demo, create a TangibleThing and in the Trix editor, try to create a bullet point list, then save this.

When inspecting the HTML code, the ol & ul list tags are there, as are the list item tags, but they are not rendering correctly in the default/current theme.

gazayas:
Closes #150.

## Details
Tailwind's [preflight styles](https://tailwindcss.com/docs/preflight) were overriding the Trix Editor styles. Adam Wathan posted a workaround [here](https://github.com/tailwindlabs/tailwindcss/issues/989#issuecomment-506555308) which I used. I also added another style to make the toolbar icons clearer to see.

I wanted to change the `color` CSS property for the icons, but since they're all images we'd have to [change the background image](https://github.com/basecamp/trix/issues/464) for each icon, which I figured wasn't worth the effort.

image

## Firefox bug
For some reason Firefox displays a red ring around the trix editor when it's focused. The file in [this issue](https://github.com/tailwindlabs/tailwindcss/discussions/6749) has the same styles that were overriding our `ol` and `ul` tags here so I adjusted the `-moz-focusring` style which originally looked like this:
```css
/*
Use the modern Firefox focus style for all focusable elements.
*/

:-moz-focusring {
  outline: auto;
}
```

This took away the red ring, so I added a comment just to make sure whoever reads the file knows why I added it there.

### DIFF
--- a/bullet_train-themes-light/app/assets/stylesheets/light/fields/trix_editor.css
+++ b/bullet_train-themes-light/app/assets/stylesheets/light/fields/trix_editor.css
@@ -9,6 +9,22 @@ trix-toolbar {
   }
 }
 
+/* Override Tailwind's preflight styles */
+/* https://github.com/tailwindlabs/tailwindcss/issues/989#issuecomment-506555308 */
+trix-editor ul {
+  list-style-type: disc;
+  padding-left: 2.5rem;
+}
+
+trix-editor ol {
+  list-style-type: decimal;
+  padding-left: 2.5rem;
+}
+
+.trix-button {
+  @apply dark:bg-slate-300 !important;
+}
+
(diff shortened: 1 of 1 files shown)


kitebuggy:
Gents, thanks for this but please note this only affects the editor styling and does not solve the rendered view of this content.

For instance:

Screenshot 2023-03-30 at 09 59 57

Results in:
Screenshot 2023-03-30 at 10 00 07

(Tested with Tangible Things on local BulletTrain instance running on latest version)

PR #208, opened 2023-03-22
Fix strong passwords configuration

gazayas:
Joint PR:
- https://github.com/bullet-train-co/bullet_train/pull/705

We added a `BulletTrain::Configuration` script in #23, but it didn't actually work because the default value wasn't being generated:

```
> rails c
Loading development environment (Rails 7.0.4.3)
irb(main):001:0> BulletTrain::Configuration
=> BulletTrain::Configuration
irb(main):002:0> BulletTrain::Configuration.default
=> #<BulletTrain::Configuration:0x00007f475b47eee8>
irb(main):003:0> BulletTrain::Configuration.default.strong_passwords
=> nil
```

This means strong passwords were never being enabled.

## `password_controller.js`
The [Stimulus controller](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-fields/app/javascript/controllers/fields/password_controller.js) which examined the password in real-time was working just fine, but you could still save the password even if the JavaScript displayed a warning.

## Display error correctly
With this fix, you will now see the following message if you submit the form and the password isn't strong enough:

Screenshot from 2023-03-22 20-44-46


jagthedrummer:
@gazayas Can you look into why tests are failing here?

gazayas:
@jagthedrummer Good to go!

PR #209, opened 2023-03-23
Fix scaffolding with sidebar flag

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/706

In our system tests, we pass the sidebar flag like this:
```
--sidebar="ti.ti-world"
```

This doesn't correctly scaffold the icon though, so I fixed it here.
I wanted to originally write this, but the script didn't play well with spaces:
```
--sidebar="ti ti-world"
```

Now we can just pass it like this:
```
--navbar="ti-world"
```

## navbar
I also just changed this to `navbar` since the developer can choose to have a sidebar or a navbar at the top.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb
@@ -1601,8 +1601,18 @@ class Scaffolding::Transformer
 
       if top_level_model?
         icon_name = nil
-        if cli_options["sidebar"].present?
-          icon_name = cli_options["sidebar"]
+        if cli_options["navbar"].present?
+          icon_name = if cli_options["navbar"].match?(/^ti/)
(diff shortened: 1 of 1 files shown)


andrewculver:
@gazayas Any idea why this test is failing? Does this branch need a merge from `main`?

gazayas:
@andrewculver We just need to merge the PR mentioned above at the same time ☝️
Since the tests use `--navbar`, the transformer script in core needed to be changed too.
I'll re-paste here just in case:
- https://github.com/bullet-train-co/bullet_train-core/pull/209

jagthedrummer:
I think we need to coordinate merging of the joint PR, and a new release of `core` with the merging of this PR. And we'll want to bump all of the BT versions in `Gemfile.lock` once `core` has been released.

https://github.com/bullet-train-co/bullet_train/pull/706#issuecomment-1672141721

PR #210, opened 2023-03-23
Documentation: Super Scaffolding Options

### LINKED ISSUE #151: Document Super Scaffolding CLI options
We have different cli options written in the transformer:
1. `cli_options["namespace"]`
2. `cli_options["skip-table"]`
3. `cli_options["skip-form"]`
4. `cli_options["skip-show"]`
5. `cli_options["skip-locales"]`
6. `cli_options["skip-api"]`
7. `cli_options["skip-model"]`
8. `cli_options["only-index"]`
9. `cli_options["skip-views"]`
10. `cli_options["skip-controller"]`
11. `cli_options["skip-routes"]`
12. `cli_options["skip-parent"]`
13. `cli_options["sortable"]`
14. `cli_options["sidebar"]`

I think it would be beneficial for developers if they were aware of the options and if we had a well-documented way of presenting how to use these options.

gazayas:
Closes #151.

As far as the `--skip-*` flags, I'm not sure how many of these we actually want to document.

For example, I'm under the impression that `--skip-table` is only for use internally (check out #199) so I didn't document it here. Maybe some of these can be set within the scope of the Transformer and removed from `cli_options`.

Besides that, I didn't document `--skip-parent` because I don't think we have the logic for it implemented as of now.

## navbar
I also added `--navbar` instead of `--sidebar` if #209 gets merged.

### DIFF
--- a/bullet_train/docs/super-scaffolding/options.md
+++ b/bullet_train/docs/super-scaffolding/options.md
@@ -1,14 +1,24 @@
 # Super Scaffolding Options
 
-There are different flags you can pass to the Super Scaffolding command which gives you more flexibility over creating your model. Add the flag of your choice to **the end** of the command for the option to take effect:
+There are different flags you can pass to the Super Scaffolding command which gives you more flexibility over creating your model. Add the flag of your choice to **the end** of the command for the option to take effect.
 ```
-bin/super-scaffold crud Project team:references description:text_field --sortable
+bin/super-scaffold crud Project Team description:text_field --sortable
 ```
 
 Most of these include skipping particular functionalities, so take a look at what's available here and pass the flag that applies to your use-case.
 
 | Option | Description |
 |--------|-------------|
-| `--sidebar="ti-world"` | Pass the Themify icon or FontAwesome icon of your choice to automatically add it to the navbar* |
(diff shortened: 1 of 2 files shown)


bullet_train/docs/super-scaffolding/options.md
```diff
@@ -0,0 +1,24 @@
...
+bin/super-scaffold crud Project Team description:text_field --sortable
+```
+
+Most of these include skipping particular functionalities, so take a look at what's available here and pass the flag that applies to your use-case.
+
+| Option | Description |
+|--------|-------------|
+| `--sortable` | [Details here](/docs/super-scaffolding/sortable.md) |
+| `--namespace=customers` | [Details here](/docs/namespacing.md) |
+| `--navbar="ti-world"` | Pass the Themify icon or Font Awesome icon of your choice to automatically add it to the navbar* |
```
jagthedrummer, line 14: @gazayas What do you think about changing this to the currently `sidebar` syntax so that we can merge this without needing to wait on the other `sidebar` => `navbar` PRs. (Those are going to need coordination with a new release of `core` and I don't know when that's going to be.)

gazayas: @jagthedrummer Works for me, just went ahead and changed it.

PR #211, opened 2023-03-23
Adjust condition for making navbar link rounded

gazayas:
This is a very minor detail, but the first link in the item menu was getting rounded even when the logo was present:

Screenshot from 2023-03-23 17-21-25


andrewculver:
@gazayas Nice catch! Thanks!

PR #235, opened 2023-04-04
Fix command to publish theme

gazayas:
The first commit here clones from `bullet_train-core` partially just so we populate the new gem directory with the contents of `bullet_train-themes-light`.

I'm a little wary to work with `git push` especially since the SSH path points to `bullet_train-core` as a whole, so for now I've added `binding.pry` to stop the process and check where we're at.

With that being said, the structure of `bullet_train-themes-light` has changed a lot, so it looks like `bullet_train-themes-light/lib/bullet_train/themes/light/custom_theme_file_replacer.rb` is going to take a lot of work. I'll try to dig my heels into it and see how far I can get.

### DIFF
--- a/bullet_train-themes-light/lib/tasks/application.rb
+++ b/bullet_train-themes-light/lib/tasks/application.rb
@@ -79,9 +79,14 @@ module BulletTrain
       end
 
       def self.release_theme(original_theme_name, args)
+        # We only want developers publishing gems off of `bullet_train-themes-light`, so if the task looks
+        # something like `rake bullet_train:themes:foo:release[bar]`, we prevent them from moving any further here.
         if original_theme_name != "light"
           puts "You can only release new themes based off of Bullet Train's Light theme. Please eject a new theme from there, and publish your gem once you've finished making changes.".red
           exit 1
+        elsif original_theme_name.nil?
+          puts "Please run the command with the name of the theme you want to release.".red
+          puts "For example: > rake bullet_train:themes:light:release[foo]"
         end
 
         puts "Preparing to release your custom theme: ".blue + args[:theme_name]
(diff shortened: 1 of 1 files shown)


gazayas:
Working hard on this one. It's still is going to take some time, but trying my best to group all the files that need to be changed dynamically as opposed to writing a bunch of string literals for the files like what's currently in `custom_theme_file_replacer.rb`. Hoping to push something by the end of the week.

gazayas:
Okay, I made some headway on this one. There are still some kinks to smooth out, but with the current state of the PR we should be able to clone `bullet_train-themes-light`, replace all it's contents with the custom theme the developer made, and then remove the custom theme from the starter repository itself.

Before marking this ready for review, I'll have to link up the new gem to make sure the views and everything else are working properly. The git logic at the end of the rake task should be ok, but I plan on double checking it just in case.

gazayas:
## Details
Marking this one as ready!
After creating a custom theme called "densha", I ran the following:

```
> rake bullet_train:themes:light:release[densha]
```

After that I spun up a new app, added the following to the Gemfile, and ran `rake bullet_train:themes:densha:install` and it worked fine:
```ruby
# I didn't push the gem to rubygems.org, but the instructions are still
# there for how to do so when running the release rake command.
gem "bullet_train-themes-densha", git: "git@github.com:gazayas/bullet_train-themes-densha.git"
```

## Afterthoughts
That felt like treading through a field of weeds for a few days (I can say that because I'm the one who wrote the original code 😆). I feel VERY happy with how it turned out though, and I feel like it's much more maintainable/legible than before.

I was also inspired by some of the [transfomer code in Action Models](https://github.com/bullet-train-pro/bullet_train-action_models/blob/987353f55f85766ef616259e982e91e8534405fe/lib/scaffolding/action_model_transformer.rb#L236-L244) to list out the processes via methods, so I ended up writing this:

```ruby
def replace_theme
  rename_cloned_files_and_directories_with_custom_theme_name
  replace_cloned_file_contents_with_custom_theme_contents
  transform_file_contents_with_custom_theme_name
  remove_custom_theme_from_main_app
  set_gem_to_v1
  update_gem_author_and_email
end
```

There are still specific files that have to be handled individually, so although I tried to write things as dynamic as possible, I still had to resort to writing String literals for those special cases.

The `install` command works too, but I want to revisit the `clean` command. Please try this PR out with a theme you have and let me know if it works for you!

gazayas:
I went ahead and deleted the tests for this too since I feel like the nature of this one is really hard to cover with just rails unit tests. Trying out the command and using the locally generated gem in a test Bullet Train application is probably the quickest way to test it for now.

jagthedrummer:
Hey @gazayas, just wanted to see if you're still feeling good about this one. I'm just a _little_ concerned about deleting the test, but I don't know much about publishing BT themes, so my concern may not be warranted.

jagthedrummer:
@gazayas, ping. (Apologies if this is on your radar and you just haven't gotten to it yet.)

gazayas:
Kinda scared of this one 😨
Will try to look at this one next.

jagthedrummer:
We can always punt on it for now if it's not high priority.

gazayas:
@jagthedrummer Sorry for the delay on this one too, the merge conflict actually wasn't as bad as I expected it to be.

As far as the tests, it was mainly little things like checking if string transformations were working properly or not, and I don't think these unit tests are of very high priority. Here's an example of constantizing a string from snake case:

```ruby
# i.e. - foo_bar or foo-bar to FooBar
def constantize_from_snake_case(str)
  str.split(/[_|-]/).map(&:capitalize).join
end
```

This is a difficult PR just to run tests against since we're creating a new repository and we'd have to see if it works in a new starter repository application. Since there are other things that are of higher priority right now I'm not too worried about the tests, but I can re-visit this sooner than later if it's a feature that we want to hone in on.

PR #236, opened 2023-04-09
Fix: Empty examples in schemas in OpenAPI

newstler:
Sends FactoryBot examples to JbuilderSchema for proper attributes and properties examples in schemas.

This fixes empty examples in 

```yaml
components:
    schemas:
        XXXAttributes:
        XXXParameters:
```

### DIFF
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -60,7 +60,7 @@ module Api
       }.merge(locals))
 
       schema_json = jbuilder.json(
-        FactoryBot.example(model.name.underscore.to_sym) || model.new,
+        FactoryBot.example(model.model_name.param_key.to_sym) || model.new,
         title: I18n.t("#{model.name.underscore.pluralize}.label"),
         # TODO Improve this. We don't have a generic description for models we can use here.
         description: I18n.t("#{model.name.underscore.pluralize}.label"),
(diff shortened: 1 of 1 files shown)


bullet_train-api/app/helpers/api/open_api_helper.rb
```diff
@@ -71,7 +71,7 @@ def automatic_components_for(model, locals: {})
       }.merge(locals))
 
       schema_json = jbuilder.json(
-        model.new,
+        FactoryBot.example(model.name.underscore.to_sym) || model.new,
```
jagthedrummer, line 74: @newstler I'm not very familiar with what BT does with OpenAPI, so please excuse me if this is a dumb question. But I'm wondering if the method is ever used in production or production-like environments. The reason I ask is becasue `factory_bot_rails` is current in the `:development, :test` group in the `Gemfile` and would not be available in production, which would then cause this method to fail.

jagthedrummer: After looking at this more closely I see that we use `FactoryBot` all over this file, so I assume that my previously-stated concern is invalid.

jagthedrummer:
@newstler I just rebased this against `main` because I thought that it looked like it was good to go, but now it's getting some test failures. Can you look into that?

newstler:
@jagthedrummer Fixed!

jagthedrummer (approved):

PR #243, opened 2023-04-12
Update source code uri in gemspecs

gazayas:


bullet_train/bullet_train.gemspec
```diff
@@ -43,10 +43,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "devise"
   spec.add_dependency "xxhash"
 
-  # This has been broken since Rails 7.
-  # spec.add_dependency "devise-two-factor"
-  # spec.add_dependency "rqrcode"
-
```
jagthedrummer, line 49: @gazayas I'm curious if you intentionally didn't update `spec.homepage` on line 8 of this file to point to `https://github.com/bullet-train-co/bullet_train-core/tree/main/bullet_train`. It's currently pointing to the starter repo. Is that what we want?

gazayas: @jagthedrummer Good catch, it's not what we want :sweat_smile:

Just changed it!

jagthedrummer (approved):

PR #250, opened 2023-04-17
Cable ready 5.0 updates

gazayas:
After merging `main` a lot of commits showed up here, so if we don't actually want to use this PR that's fine too, but the last two commits show how we can update to the latest version of Bullet Train.

We'll have to submit a PR to update the `cable_ready` version in `package.json`.

The reactivity test is working locally, but I think I'll look over things one more time just to be sure.

cc/ @julianrubisch


gazayas:
Closing this since we already updated to cable_ready 5

PR #252, opened 2023-04-18
Move and rename API method used in controller tests

gazayas:
This method exists in the TangibleThings controller test and the Access Tokens controller test. It looks like this:
```ruby
tangible_thing_data = JSON.parse(build(:scaffolding_completely_concrete_tangible_thing, absolutely_abstract_creative_concept: nil).to_api_json.to_json)
```

I found `to_api_json.to_json` to be confusing since the comment in `to_api_json` has the following comment for the return value:
```ruby
# Returns a hash, not string.
```

I changed the name to avoid confusion, and now that we have all of the packages in the same repository, we can remove the `if defined?(BulletTrain::API)`.


jagthedrummer (approved):

olbrich:
In the future, when you rename methods like this, please use the `ActiveSupport::Deprecation.deprecate_method` to give an appropriate warning before removal.

jagthedrummer:
Good suggestion, @olbrich!

PR #254, opened 2023-04-19
Fix for `api/example_bot` to work on non-postgres apps

abradburne:
reset_pk_sequence is only available for Postgres, not MySQL, so check that the method is available before calling.

### DIFF
--- a/bullet_train-api/lib/bullet_train/api/example_bot.rb
+++ b/bullet_train-api/lib/bullet_train/api/example_bot.rb
@@ -58,6 +58,8 @@ module FactoryBot
     end
 
     def reset_tables!
+      # This is only availble for postgres
+      return unless ActiveRecord::Base.connection.respond_to?(:reset_pk_sequence!)
       @tables_to_reset.each do |name|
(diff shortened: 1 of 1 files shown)


newstler:
That means that in any other non-postgres db we will have missed ids... is there any other way to set id sequence for MySQL?

abradburne:
I'm not well up on MySQL but I've done some investigation. 

It appeared that there is a method in AR, _but_ it isn't implemented for MySQL (or Postgres for that matter).
`reset_sequency(table, col, max_value)`
https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/DatabaseStatements.html#method-i-reset_sequence-21

MySQL docs suggest that you can update it with e.g. 
`ALTER TABLE contacts AUTO_INCREMENT=500` - however, after doing some tests, I can only get it to go forward, never backwards. i.e. you can skip 100 values, but you cannot go back to values that have been deleted.

SQLite apparently does automatically (i.e. always uses the next available int) but I haven't tested.

PR #255, opened 2023-04-19
Add success percentage on outgoing webhooks endpoint table

bhumi1102:


bhumi1102:
@andrewculver This one is ready. but two things: 1. Not sure if creating such a 'capsule' component or the location of where I put it is the right place. Feel free to share better ways or more "BT way" 2. putting the label 'error rate' in the .yml file, I couldn't get the syntax to work anywhere other than in the 'attribute' which is not quite right.

jagthedrummer:
@andrewculver I'm leaving this one for you since it deals with outgoing webhooks.

PR #257, opened 2023-04-20
Invitations API

newstler:
Implements memberships API and adds OpenAPI specs for invitations.
Requires https://github.com/bullet-train-co/bullet_train/pull/737

Replaced `has_many :roles, through: :membership` in `bullet_train/app/models/concerns/invitations/base.rb` with a method as `roles` is not a table.

### DIFF
--- a/bullet_train/app/controllers/concerns/account/invitations/controller_base.rb
+++ b/bullet_train/app/controllers/concerns/account/invitations/controller_base.rb
@@ -76,7 +76,7 @@ module Account::Invitations::ControllerBase
   # POST /invitations/1/resend
   def resend
     @invitation = Invitation.find_by(uuid: params[:id])
-    if @invitation
+    if @invitation.touch
       UserMailer.invited(params[:id]).deliver_later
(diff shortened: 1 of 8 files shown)


jagthedrummer:
@newstler, can you check out the test failures on this?

bullet_train/app/controllers/concerns/account/invitations/controller_base.rb
```diff
@@ -76,7 +76,7 @@ def accept
   # POST /invitations/1/resend
   def resend
     @invitation = Invitation.find_by(uuid: params[:id])
-    if @invitation
+    if @invitation.touch
```
jagthedrummer, line 79: I think this line is causing the test failures. I think the `touch` should probably be inside the conditional instead of _being_ the conditional.

```suggestion
    if @invitation
      @invitation.touch
```

newstler:
@jagthedrummer Sorry, been sick past few days, then totally broke my computer setup 😅 Now reinstalling everything

Implemented your suggestion and that seemed to work! Thank you

jagthedrummer (approved):
Thanks for the work on this @newstler! This will be out in `1.3.16` a little later today.

PR #263, opened 2023-04-24
Refactor LoadsAndAuthorizesResource

aaricpittman:
This PR does the following:

1. Namespaces `LoadsAndAuthorizesResource` to `BulletTrain::LoadsAndAuthorizesResource`. This will allow client apps to reference the module still if their app also defines `LoadsAndAuthorizesResource`.
2. Maintains backward compatibility by still defining `LoadsAndAuthorizesResource` and having it include `BulletTrain::LoadsAndAuthorizesResource`.
3. Gets rid of the `included` block, as it was unnecessary.
4. Wraps class methods in a `class_methods` block and removes `self.`


jagthedrummer:
@aaricpittman, I'm sorry that this has been sitting here for so long. If you don't mind would you update your the branch in this with `main` so that we can be sure all of the tests are passing? (We've done a lot of work on the test suite lately and should now be testing much more thoroughly than before.)

jagthedrummer:
Thanks for the work on this @aaricpittman. I ended up needing to make a few changes to get the linter happy so that tests would pass when we merge to `main`. Since I don't have permission to push to your branch I started a new branch and am opening a new PR off of that one. So closing this in favor of https://github.com/bullet-train-co/bullet_train-core/pull/508

PR #267, opened 2023-04-28
Render table rows as collections

julianrubisch:
Changes the way table rows in super scaffolded index views are rendered from a loop to a native `render partial: ... collection: ...` call.

This has multiple advantages:

1. It is highly optimized for Russian doll caching by adding `cached: true` (which will use multi-fetch if the cache store supports it)
2. It provides a `counter` local variable inside the rendered partial


kaspth:
This also fits with #331, so I decided to merge this. I'll take responsibility if anything goes wrong.

PR #268, opened 2023-04-29
Add environment variable check for `billing_enabled?` helper

### LINKED ISSUE #244: `billing_enabled?` should probably check if billing code is configured as well as installed
If I install the billing module but do not set `ENV['STRIPE_SECRET_KEY']` I still get a "Billing" option added to the team menu, which leads to a subscription experience that doesn't work due to the missing secret.

It would be nice if `billing_enabled?` would check for the presence of that critical variable so that the feature can be toggled on and off via an ENV change instead of by deploying code.

gazayas:
Closes #244.

It seems like we want this to be dependent on the presence of the gem itself as opposed to the environment variable, but I included this check just in case since subscriptions won't work if the API key isn't there.

Also, since the `bullet_train-billing` and `bullet_train-billing-stripe` gems are each their own thing, I can look into this deeper if we want to separate the logic and only check for the Stripe API key when using `bullet_train-billing-stripe`.

cc/ @jagthedrummer

### DIFF
--- a/bullet_train/lib/bullet_train.rb
+++ b/bullet_train/lib/bullet_train.rb
@@ -82,7 +82,7 @@ def inbound_email_enabled?
 end
 
 def billing_enabled?
-  defined?(BulletTrain::Billing)
+  ENV["STRIPE_SECRET_KEY"].present? && defined?(BulletTrain::Billing)
 end
 
 # TODO This should be in an initializer or something.


jagthedrummer:
I just realized one additional bit of info that I want to mention here for context.

`BulletTrain::Billing` is a _namespace_ and not really a _gem_. A number of BT packages contain classes and modules in that namespace and just the presence of one of those packages probably shouldn't trigger `billing_enabled?` to be true.

gazayas:
Yeah, I guess I'm just curious as to why we add `bullet_train-billing` to the Gemfile like the [docs](https://bullettrain.co/docs/billing/stripe) say to do. Maybe `bullet_train-billing` is only hosted on Gemfury and not here in the private repositories like `bullet_train-billing-stripe`, etc.?

Maybe we should split this into `billing_installed?` and `billing_enabled?`, and then for `billing_enabled?` we can check env vars:
```ruby
def billing_installed?
  defined?(BulletTrain::Billing)
end

def billing_enabled?
  ENV["STRIPE_SECRET_KEY"].present? || ENV["OTHER_KEY_FOR_BILLING"].present? # || ...
end
```

This would be consistent with say how we handle Devise 2FA, because it's always technically <i>installed</i>, just not necessarily enabled.

gazayas:
Scratch that, just saw the repo here https://github.com/bullet-train-pro/bullet_train-billing

jagthedrummer:
I do think there's something to the idea of making a difference between `billing_installed?` and `billing_enabled?`. But I also think that `billing_installed?` is kind of ambiguous and hard to define given the number of packages that push code into the `BulletTrain::Billing` namespace. We may want to be more specific about exactly which billing features are installed.

PR #270, opened 2023-04-29
Fix: OpenAPI warnings

newstler:
Closes https://github.com/bullet-train-co/bullet_train/issues/734
Requires https://github.com/bullet-train-co/bullet_train/pull/749

- Fixes warnings shown by `yarn exec redocly lint http://127.0.0.1:3000/api/v1/openapi.yaml`
- Includes https://github.com/bullet-train-co/bullet_train-core/pull/236 and https://github.com/bullet-train-co/bullet_train-core/pull/258
- Test `test/controllers/api/open_api_controller_test.rb` in starter repo was rewritten to fail on warnings
- Upgraded `jbuilder-schema` to `>= 2.2.0`
- Added Jbuilder hack module in `bullet_train-api/lib/jbuilder/values_transformer.rb` to modify attributes produced by external classes, by now it replaces `ActionText::RichText` object with it's `body` value. More can be added if needed.

### DIFF
--- a/bullet_train-api/lib/jbuilder/values_transformer.rb
+++ b/bullet_train-api/lib/jbuilder/values_transformer.rb
@@ -2,7 +2,7 @@
 
 require "jbuilder"
 
-module Hacks
+module ValuesTransformer
   def _set_value(key, value)
     value = value.body if value.is_a?(ActionText::RichText)
 
@@ -10,4 +10,4 @@ module Hacks
   end
 end
 
-::Jbuilder.prepend Hacks
+::Jbuilder.prepend ValuesTransformer
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,4 +46,4 @@ RUBY VERSION
    ruby 3.2.2p53
 
 BUNDLED WITH
-   2.3.8
+   2.4.12
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -60,7 +60,7 @@ module Api
       }.merge(locals))
 
       schema_json = jbuilder.json(
-        FactoryBot.example(model.name.underscore.to_sym) || model.new,
+        FactoryBot.example(model.model_name.param_key.to_sym) || model.new,
         title: I18n.t("#{model.name.underscore.pluralize}.label"),
         # TODO Improve this. We don't have a generic description for models we can use here.
         description: I18n.t("#{model.name.underscore.pluralize}.label"),
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -60,7 +60,7 @@ module Api
       }.merge(locals))
 
       schema_json = jbuilder.json(
-        model.new,
+        FactoryBot.example(model.name.underscore.to_sym) || model.new,
         title: I18n.t("#{model.name.underscore.pluralize}.label"),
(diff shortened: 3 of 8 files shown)


bullet_train-api/lib/jbuilder/values_transformer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "jbuilder"
+
+module ValuesTransformer
+  def _set_value(key, value)
+    value = value.body if value.is_a?(ActionText::RichText)
```
gazayas, line 7: This reminds me of the comment in the Transformer about rich text attributes

https://github.com/bullet-train-co/bullet_train-core/blob/aad074837f28f3dae7351707225bf8d385bf4047/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb#L1140

We skip trix editor attributes when super scaffolding jbuilder files, but should we add the body somehow over there in the transformer as well?

gazayas (commented):
LGTM :+1: The Open API controller test wasn't passing for me before this PR, but now it's good.

bullet_train-api/lib/jbuilder/values_transformer.rb
newstler: Well, anyway those attributes are exposed for `ScaffoldingCompletelyConcreteTangibleThing` when `HIDE_THINGS` is not enabled. So this Jbuilder hack lets us show `ActionText::RichText` attributes properly (as `string` instead of whole object). I think if those were skipped before, now they can be added. Also, if there are any other attributes that should be exposed differently in Jbuilder, they might also be added to this file.

I afraid I didn't fully understand what you meant by `should we add the body somehow over there in the transformer as well?` :)

newstler:
@gazayas Try with Open API controller test from the [corresponding branch](https://github.com/bullet-train-co/bullet_train/pull/749) in starter repo, it was rewritten.

I also updated description.
Not sure if it's right, but this PR includes 2 other PRs that it needs to work correctly :)

bullet_train-api/lib/jbuilder/values_transformer.rb
gazayas: > I afraid I didn't fully understand what you meant by should we add the body somehow over there in the transformer as well? :)

I just wasn't sure how we should handle rich text attributes with the API, but after looking into it a bit, I'm thinking I'll make a different PR to return the `ActionText::RichText` object as a whole instead of just the body (that seems to make the most sense to me). As far as what we return, I think we can leave the decision up to Andrew. I'm fine with how things are now though. Thanks!

Here is an example of `bar` as a trix editor attribute:
Screenshot from 2023-05-08 20-22-11

newstler: Well, that data in API feels completely useless for me, as it is the internal ActionText stuff. Also, I am pretty sure warnings were there as when we create a record, we send `bar` as a string. And I'm not sure we now support different types of same attribute for Attributes and Parameters. Should we?

gazayas: > Well, that data in API feels completely useless for me, as it is the internal ActionText stuff, and as an API user I can't do anything with that, it's only clutters up the information.

Good point, I didn't consider this. I hadn't run the the ActionText::RichText API object against the OpenAPI test so that makes sense about the warnings. Thanks for explaining!

jagthedrummer:
@newstler Can you look at fixing the test failures here? They seem to be related to the OpenAPI stuff that's changed in this PR and/or the one in the starter repo.

newstler:
@jagthedrummer Fixed!

PR #275, opened 2023-05-04
Upload profile photo with ActiveStorage when Cloudinary isn't enabled

gazayas:
Closes https://github.com/bullet-train-co/bullet_train/issues/576

## Details

I opened #274 so we can upload photos properly, but I've also added the fix here just in case. This pull request gets the bare minimum done for uploading photos and saving them with ActiveStorage.

Screenshot from 2023-05-04 12-00-29

## Things to get done after this pull request
1. Add `image_controller` as a new Stimulus controller to display the image before updating.
2. Add an image partial which doesn't depend on `_cloudinary_image.html.erb`
3. Refactor the Users helper.

For 1 and 3, we could lump them into this PR if we want to so I'll at least start working on it, but I'm okay with making them into separate PRs.

For 2, if we make a basic image partial I think it would be better to separate the PR and then add tests over in the starter repository.


gazayas:
I went ahead and refactored the User helper. I kind of want to delete the `profile_header` methods because they aren't used anywhere, just wouldn't want to remove them if other developers are using them.

gazayas:
I thought we would need to create an entirely new Stimulus controller, but ActiveStorage's [representation](https://guides.rubyonrails.org/active_storage_overview.html#displaying-images-videos-and-pdfs) seems to do the trick.

The only thing is, I had to install [libvips](https://www.libvips.org/install.html) manually:
```
> sudo apt-get install libvips
```

I've implemented it here in the `file_field` partial, so if we're okay going down that route, we won't have to add any extra PRs after this. Maybe just a test in the starter repository to ensure file previews appear after uploading, and maybe a few tweaks to the file_field Stimulus controller.

Currently looks like this:
Screenshot from 2023-05-04 14-24-22

jagthedrummer:
@gazayas, this seems like a great addition, but there are some conflicts now. Can you look at resolving them?

gazayas:
@jagthedrummer Conflicts resolved and all tests including the Super Scaffolding system tests are passing. Just running into #380 now, will try to solve that.

jagthedrummer:
@gazayas in https://github.com/bullet-train-co/bullet_train-core/pull/383 I worked out the branch problem that was affecting the tests for this PR. Can you merge/rebase `main` and see if that does it for the tests on this?

gazayas:
@jagthedrummer Done!

PR #282, opened 2023-05-07
Refactor `with_attribute_settings`

kaspth:
Seems like we can just use one hash to store the settings in. And then use keyword argument defaults to preserve the old behavior of keeping the previous setting in case no override is passed.

This also uses `rescue` so we revert to the previous values in case an exception is thrown.

### DIFF
--- a/bullet_train/app/helpers/attributes_helper.rb
+++ b/bullet_train/app/helpers/attributes_helper.rb
@@ -1,32 +1,17 @@
 module AttributesHelper
   def current_attributes_object
-    @_attributes_helper_objects&.last
+    @_current_attribute_settings&.dig(:object)
   end
 
   def current_attributes_strategy
-    @_attributes_helper_strategies&.last
+    @_current_attributes_settings&.dig(:strategy)
   end
 
-  def with_attribute_settings(options)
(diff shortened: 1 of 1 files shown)


bullet_train/app/helpers/attributes_helper.rb
```diff
@@ -1,32 +1,17 @@
...
     yield
-
-    if options[:strategy]
-      @_attributes_helper_strategies.pop
-    end
-
-    if options[:object]
-      @_attributes_helper_objects.pop
-    end
+  rescue
```
olbrich, line 14: Maybe this should be `ensure` instead of `rescue` so it won't swallow exceptions thrown in the block.

kaspth: @olbrich you're totally right, I thought something was up — thank you!

jagthedrummer (approved):
@kaspth This looks good to me, and I just rebased it against `main` and force pushed to get a fresh CI run, and all tests are passing. Do you have any concerns about merging this? (Just asking since it's been open for a while.)

kaspth:
@jagthedrummer thanks for the extra set of eyes!

Issue #288, opened 2023-05-10
Invitation workflow should use `secure_compare`?


andrewculver:
From @olbrich:

> I was just looking around at the invitation_only workflow for BT and noticed this... https://github.com/bullet-train-co/bullet_train-core/blob/20ec101e9bb23f0858c00b2976edd6633bb8421a/bullet_train/app/helpers/invitation_only_helper.rb#L2-L4.  I suspect that this should probably use secure_compare https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html#method-c-secure_compare to avoid timing attacks.

PR #289, opened 2023-05-10
use secure_compare for invitation keys

bhumi1102:
closes https://github.com/bullet-train-co/bullet_train-core/issues/288

**Overview**
Updated the comparison for `invitation_keys` to use `secure_compare`

**Testing Done**
Set values in `ENV['INVITATION_KEYS']` locally and navigate to the `/invitation?key=XYZ`. Ensure that for keys that match one of the ones set in the environment variable, we are redirected correctly to create a new user or a new team. For keys that do not match, we are redirected to 404.

### DIFF
--- a/bullet_train/app/helpers/invitation_only_helper.rb
+++ b/bullet_train/app/helpers/invitation_only_helper.rb
@@ -1,6 +1,14 @@
+require 'active_support/security_utils'
+
 module InvitationOnlyHelper
   def invited?
-    session[:invitation_key].present? && invitation_keys.include?(session[:invitation_key])
+    return false unless session[:invitation_key].present?
+
+    result = invitation_keys.find do |key|
+      ActiveSupport::SecurityUtils.secure_compare(key, session[:invitation_key])
+    end
+
+    result.present?
   end
 
   def show_sign_up_options?


olbrich (commented):
This is definitely an improvement, but still lets an attacker determine how many invitation keys are present using timing attacks.

If we make another array for the keys to compare like `Array.new(10) { invitation_keys[_1] }` then the list will always contain 10 items. That number could be larger.

Not really a big deal, since guessing any individual one would be difficult anyway.

olbrich (approved):

bhumi1102:
> This is definitely an improvement, but still lets an attacker determine how many invitation keys are present using timing attacks.
> 
> If we make another array for the keys to compare like `Array.new(10) { invitation_keys[_1] }` then the list will always contain 10 items. That number could be larger.
> 
> Not really a big deal, since guessing any individual one would be difficult anyway.

Yeah, picking a proper array size would the open question for that. Currently per documentation, the user can  "set INVITATION_KEYS to one or more comma-delimited values".  I suppose there could an upper limit.  Either way, as far as timing attacks based on string comparison time, `secure_compare` would make that difficult enough.

jagthedrummer (approved):

Issue #294, opened 2023-05-15
super-scaffold oauth uses hardcoded values


jjjessen:
Running super-scaffold on any oauth provider e.g.
bin/super-scaffold oauth-provider omniauth-auth0 auth0 Oauth::Auth0Account id_key secret_key

yields the following output on the command line

You'll probably need to `bundle install`.
You'll need to configure API keys for this provider in `config/application.yml`, like so:

  id_key: ...
  secret_key: ...

I believe its problematic that these values are hardcoded when running super-scaffold. The secrets are also hardcoded in bullet_train.rb

def auth0_enabled?
  ENV['id_key'].present? && ENV['secret_key'].present?
end

I suggest that super-scaffold instead uses environment variables and on the command line output removes the suggestion of having hardcoded values in config/application.yml

gazayas:
@jjjessen The oauth-provider scaffolding is correct, but I can understand if the help prompt is confusing. This is what the prompt looks like right now when running `bin/super-scaffold oauth-provider`

```
> bin/super-scaffold oauth-provider

🚅  usage: bin/super-scaffold oauth-provider <omniauth_gem> <gems_provider_name> <our_provider_name> <PROVIDER_API_KEY_IN_ENV> <PROVIDER_API_SECRET_IN_ENV> [options]
```
`<PROVIDER_API_KEY_IN_ENV>` and `<PROVIDER_API_SECRET_IN_ENV>` represent the names of the variables (hence the "IN_ENV"), not the values themselves. Therefore, the command should look like this:

```
> bin/super-scaffold oauth-provider omniauth-auth0 auth0 Oauth::Auth0Account AUTH0_API_KEY AUTH0_PROVIDER_SECRET
```

Again, you literally write the <b>environment variable name</b> here and not the value. The value is to be placed in `config/application.yml` later where all of your other environment variables should live, as per the output when running the scaffolder:

```
You'll probably need to `bundle install`.

You'll need to configure API keys for this provider in `config/application.yml`, like so:

  AUTH0_API_KEY: ...
  AUTH0_PROVIDER_SECRET: ...
```

However, again, I can understand how the help prompt can be confusing so I'll try to write up a PR to fix that.

jjjessen:
sorry, now I understand that it's my inability to read the documentation :-) I'm closing the case, sorry for the disturbance

PR #295, opened 2023-05-16
Write clearer instructions for the oauth-provider scaffolder

gazayas:
Related to #294.

I've personally made the same mistake before, so I wrote clearer instructions so other developers don't run into the same issue.


jagthedrummer (approved):
This seems much clearer and less open to misinterpretation. 👍

PR #298, opened 2023-05-16
Fix: Produce really underscored DOM ids from a title

### LINKED ISSUE #278: `underscore` is insufficient for producing DOM ids from a title
In [this menu subsection](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-themes-light/app/views/themes/light/menu/_subsection.html.erb#L13) we use `.underscore` to try to turn a human-readable title into a DOM id.

```erb
<div ... id="<%= title.underscore %>">
```

But if the `title` is two words like `"Test Team"` then the id is evaluated to be:

```erb
<div ... id="test team">
```

It would be better to use `parameterize` which would result in:

```erb
<div ... id="test-team">
```

Or possibly `parameterize('_')` to get:

```erb
<div ... id="test_team">
```

newstler:
Closes #278.

Replaces `.underscore` with `.parameterize(separator: '_')` in producing DOM ids from a title.

Using `parameterize` instead of `gsub` or `tr` not only downcases the string and replaces spaces with underscores, but also handles other special characters by converting them into their ASCII equivalents, which is useful for creating URL-friendly strings.

### DIFF
--- a/bullet_train-themes-light/app/views/themes/light/menu/_subsection.html.erb
+++ b/bullet_train-themes-light/app/views/themes/light/menu/_subsection.html.erb
@@ -10,7 +10,7 @@
   </div>
 <% else %>
   <div>
-    <div class="inline-block relative group" data-controller="desktop-menu" data-action="keydown->desktop-menu#hideSubmenu" id="<%= title.underscore %>">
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
Yes!

PR #300, opened 2023-05-17
Avoid Super Scaffolding Stripe as an OAuth provider

gazayas:
On `main`:

```
> bin/super-scaffold oauth-provider omniauth-stripe-connect stripe_connect Oauth::StripeAccount STRIPE_CLIENT_ID STRIPE_SECRET_KEY --icon=ti-money

🚨 Before doing the actual Super Scaffolding, you'll need to generate the models like so:

  rails g model Oauth::StripeAccount uid:string data:jsonb user:references
  rails g model Integrations::StripeInstallation team:references oauth_stripe_account:references name:string
  rails g model Webhooks::Incoming::Oauth::StripeAccountWebhook data:jsonb processed_at:datetime verified_at:datetime oauth_stripe_account:references

However, don't do the `rake db:migrate` until after you re-run Super Scaffolding, as it will need to update some settings in those migrations.

```

However, these models already exist.

This is happening because we extracted the integration logic for Stripe into `bullet_train-integrations-stripe`. Before, these models were directly in `app/models/` so the following `unless` statement in the scaffolder made sense:
```ruby
unless File.exist?(oauth_transform_string("./app/models/oauth/stripe_account.rb", options)) &&
File.exist?(oauth_transform_string("./app/models/integrations/stripe_installation.rb", options)) &&
File.exist?(oauth_transform_string("./app/models/webhooks/incoming/oauth/stripe_account_webhook.rb", options))
```

Since we extracted the files into `bullet_train-integrations-stripe`, I edited the statement so it will run when trying to Super Scaffold a Stripe integration.

## Error when trying to log in
I'm not sure if this has to do with the scaffolder or something else I'm overlooking, but I'm still missing something because I can't log in, and I get this JSON response:
```
error:
  message: "No application matches the supplied client identifier"
```

If anyone wants to run the scaffolder with this branch and let me know if they get the same results or not, that would be helpful.


gazayas:
After working through this one for a bit it seems that the API error I was getting above was because I needed to use a client id provided in the "Connect" tab in Stripe, and since I only have test data set up for now I didn't actually go through with logging in. Basically, I don't believe it's a Bullet Train issue, just a configuration issue.

Since this is the main oauth provider, my assumption is that it's working and we just didn't update this file during the Great Unbundling.

jagthedrummer:
@gazayas I can't really tell from your last comment whether this is a PR that we still need. If it is can you see about resolving the conflicts?

jagthedrummer:
@gazayas, another ping on another PR, just in case it's not already on your radar.

gazayas:
> @gazayas I can't really tell from your last comment whether this is a PR that we still need. If it is can you see about resolving the conflicts?

@jagthedrummer Conflicts resolved! In my last comment, I was just saying that I hadn't gone through all the steps to set up Stripe on their website, but all the logic on our side should be good to go.

After merging main, I got some linter errors that apparently weren't implemented in previous PRs, so I addressed those here as well.

jagthedrummer (approved):

PR #303, opened 2023-05-20
Change path search order of resolver

gazayas:
Context: https://discord.com/channels/836637622432170028/836637623048601633/1109145560117686363

When a partial isn't found, we get the partial resolver itself (`partial_resolver.html.erb`), which raises the error shown in the Discord chat.

The problem is that this happens with the `partial_path` method, and since that value was being set to `result[:absolute_path]`, the resolver didn't bother to run `|| locale_path`.

## An issue with `locale_path` itself
After making the fix things were working fine for the following locales:
1. `en.account.onboarding.user_details.edit.header`
2. `en.devise.headers.sign_in`

The resolver was also working for model names like `Teams::Base`. However, `locale_path` was returning `nil` for `en.users.self.password.help` (a locale that can be found on the sign in page). This seems to be an issue with `locale_path` itself, so I think I'll have to look into that a little bit more.

I also wrote a more generalized message for when the value isn't found.

### DIFF
--- a/bullet_train/lib/bullet_train/resolver.rb
+++ b/bullet_train/lib/bullet_train/resolver.rb
@@ -89,7 +89,7 @@ module BulletTrain
         package_name: nil,
       }
 
-      result[:absolute_path] = file_path || class_path || partial_path || locale_path
+      result[:absolute_path] = file_path || class_path || locale_path || partial_path
 
       # If we get the partial resolver template itself, that means we couldn't find the file.
       if result[:absolute_path].match?("app/views/bullet_train/partial_resolver.html.erb")
--- a/bullet_train/lib/bullet_train/resolver.rb
+++ b/bullet_train/lib/bullet_train/resolver.rb
@@ -92,10 +92,10 @@ module BulletTrain
       result[:absolute_path] = file_path || class_path || locale_path || partial_path
 
       # If we get the partial resolver template itself, that means we couldn't find the file.
-      if result[:absolute_path].match?("app/views/bullet_train/partial_resolver.html.erb")
-        puts "We could not find the partial you're looking for: #{@needle}".red
+      if result[:absolute_path].match?("app/views/bullet_train/partial_resolver.html.erb") || result[:absolute_path].nil?
+        puts "We could not resolve the value you're looking for: #{@needle}".red
         puts ""
-        puts "Please try passing the partial string using either of the following two ways:"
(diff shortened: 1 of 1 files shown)


gazayas:
Concerning the locales issue, you can find the details at #306.

With the fix here, we will now only return keys to the browser when those keys actually have a corresponding translation. The sign in screen now looks like this (you can compare it to the previous screenshot in #306):

Screenshot from 2023-05-22 20-29-33

jagthedrummer:
Actually, on second thought, I'm not 100% convinced this is a good approach. Since those translations are missing we end up not showing a translation to the end user, which is fine. But I think it might be useful to show developers that there's an opportunity to provide those translations if they want to.

For instance, if someone wanted to add placeholders for email and password the current situation makes it clear that there are translation keys that could be provided to fill in those "blanks". This PR would make it so that the possibility of using those translation keys isn't obvious, which may lead people down the path of trying to eject and modify templates in order to add a placeholder, when all they really need to do is to supply a missing translation.

jagthedrummer:
Just to make sure I wasn't misunderstanding I just tried adding those missing translations to a locale file and they do show up.

```
en:
  users:
    self:
      email:
        placeholder: somebody@somewhere.com
```

CleanShot 2023-08-01 at 12 20 36

So I think the current behavior is preferable to hiding the fact that there's an opportunity to provide translations strings that BT doesn't provide by default.

If we're concerned about "missing" translation keys being confusing I think it would be better to add the missing keys to translation files with an empty string (if we want the BT default to be that they don't appear).

jagthedrummer:
I guess this PR might actually be doing two things, one of which I think we should keep (tweaking the resolve order), and the other I think we should not (hiding translation keys if a translation isn't provided), if it's possible to separate them.

gazayas:
> Actually, on second thought, I'm not 100% convinced this is a good approach. Since those translations are missing we end up not showing a translation to the end user, which is fine. But I think it might be useful to show developers that there's an opportunity to provide those translations if they want to.

I see, that didn't even cross my mind. I'll look over the PR and see what I can do. I'm thinking I might provide a help text of the key isn't found and let developers know that they can add a string themselves.

gazayas:
@jagthedrummer I reverted the changes and simply wrote a help message so developers know that the locale's value just might not be present. Here's what it looks like before adding the value and after.

Screenshot from 2023-08-04 10-40-43

jagthedrummer (approved):
🤘 This looks like an excellent improvement.

PR #304, opened 2023-05-20
Stack registration password and confirmation on mobile

goulvench:
The password and confirmation fields & labels overlap on narrow screens.


jagthedrummer (approved):
Good call!

PR #307, opened 2023-05-22
Prefer `.public_send` over `.send` in partials

### LINKED ISSUE #115: Prefer `.public_send` over `.send` in partials
https://github.com/bullet-train-co/bullet_train-core/blob/f87143fbc2dd41a5e97b76be22971f95fb22cbca/bullet_train-themes/app/views/themes/base/attributes/_text.html.erb#L6

It would probably be better to use `public_send` in all the partials so that we at least respect private methods.

newstler:
Closes #115 

Instead of `object.send(attribute).present?` it would probably be better to use `object.respond_to?(attribute)`. For other cases `public_send` should work fine.

### DIFF
--- a/bullet_train-themes/app/views/themes/base/attributes/_belongs_to.html.erb
+++ b/bullet_train-themes/app/views/themes/base/attributes/_belongs_to.html.erb
@@ -3,12 +3,12 @@
 <% disable_link ||= false %>
 <% url ||= nil %>
 
-<% if object.send(attribute).present? %>
+<% if object.public_send(attribute).present? %>
   <%= render 'shared/attributes/attribute', object: object, attribute: attribute, strategy: strategy, url: url do %>
     <% if disable_link %>
-      <%= object.send(attribute).label_string %>
+      <%= object.public_send(attribute).label_string %>
     <% else %>
-      <%= link_to object.send(attribute).label_string, url || [:account, object.send(attribute)] %>
+      <%= link_to object.public_send(attribute).label_string, url || [:account, object.public_send(attribute)] %>
     <% end %>
   <% end %>
 <% end %>
--- a/bullet_train-themes/app/views/themes/base/attributes/_option.html.erb
+++ b/bullet_train-themes/app/views/themes/base/attributes/_option.html.erb
@@ -2,8 +2,8 @@
 <% strategy ||= current_attributes_strategy || :none %>
 <% url ||= nil %>
 
-<% if object.send(attribute).present? %>
+<% if object.public_send(attribute).present? %>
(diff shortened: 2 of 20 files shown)


bullet_train-themes/app/views/themes/base/attributes/_option.html.erb
```diff
@@ -2,8 +2,8 @@
 <% strategy ||= current_attributes_strategy || :none %>
 <% url ||= nil %>
 
-<% if object.send(attribute).present? %>
+<% if object.respond_to?(attribute) %>
```
gazayas, line 5: Since we're looking to see if the attribute value is populated as opposed to whether the record responds to the value, we should be using `object.public_send(attribute).present?` here instead. As you can see here `option_value` belongs to the record, it's just `nil`. (We can use `any?` in the `_options.html.erb` partial though because `object.public_send(attribute)` there is an array).

```
[1] pry(#<#<Class:0x00007fc80cea6b00>>)> object
=> #<Scaffolding::CompletelyConcrete::TangibleThing:0x00007fc7ae0384a0
 id: 39,
 absolutely_abstract_creative_concept_id: 30,
 text_field_value: "My value for this text field",
 button_value: "three",
 cloudinary_image_value: nil,
 date_field_value: nil,
 email_field_value: "me@acme.com",
 password_field_value: "secure-password",
 phone_field_value: "+12015518321",
 super_select_value: "one",
 text_area_value: "Long text for this text area field",
 created_at: Wed, 21 Jun 2023 00:26:54.487924000 UTC +00:00,
 updated_at: Wed, 21 Jun 2023 00:26:54.496684000 UTC +00:00,
 sort_order: nil,
 date_and_time_field_value: nil,
 multiple_button_values: ["four", "five"],
 multiple_super_select_values: ["five", "six"],
 color_picker_value: "",
 boolean_button_value: true,
 option_value: nil,
 multiple_option_values: []>
[2] pry(#<#<Class:0x00007fc80cea6b00>>)> object.respond_to?(attribute)
=> true
[3] pry(#<#<Class:0x00007fc80cea6b00>>)> object.send(attribute).present?
=> false
[4] pry(#<#<Class:0x00007fc80cea6b00>>)> attribute
=> :option_value

```

bullet_train-themes/app/views/themes/base/attributes/_belongs_to.html.erb
```diff
@@ -3,12 +3,12 @@
 <% disable_link ||= false %>
 <% url ||= nil %>
 
-<% if object.send(attribute).present? %>
+<% if object.respond_to?(attribute) %>
```
gazayas, line 6: Same here with `object.public_send(attribute).present?`. Fixing these two spots should make the tests pass.

gazayas (approved):

PR #310, opened 2023-05-29
Add super select config target to get options from main app

gazayas:
Closes https://github.com/bullet-train-co/bullet_train/issues/783

Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/799

Since we have no way of editing Super Select options from the starter repository, I added this target to the Super Select stimulus controller so we can customize these elements to our liking without having to eject any files.

Linking up to the JS code in bullet_train-core can be super flaky so I'm trying my best to test everything well. I'll have to double check my work on this one (especially with the spread operator to merge the two objects).

Also just curious if this presents any security issues since we're technically passing in raw JavaScript from the DOM.

Anyways, I worked through a number of ideas in the [original issue](https://github.com/bullet-train-co/bullet_train/issues/783) in case anyone wants to look over those or provide a better solution.

### DIFF
--- a/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
+++ b/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
@@ -6,7 +6,7 @@ const select2SelectedPreviewSelector = ".select2-selection--single"
 const select2SearchInputFieldSelector = ".select2-search__field"
 
 export default class extends Controller {
-  static targets = [ "select" ]
+  static targets = [ "select", "superSelectConfig" ]
   static values = {
     acceptsNew: Boolean,
     enableSearch: Boolean,
@@ -46,6 +46,7 @@ export default class extends Controller {
     let options = {
       dropdownParent: $(this.element)
     };
+    const options = {...options, ...this.superSelectConfigTarget.dataset["superSelectConfig"]}
 
     if (!this.enableSearchValue) {
       options.minimumResultsForSearch = -1;
--- a/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
+++ b/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
@@ -79,7 +79,8 @@ export default class extends Controller {
     options.width = 'style';
 
     // Merge in custom options.
-    options = {...options, ...JSON.parse(this.optionsOverride)}
+    const custom_options = Object.keys(this.optionsOverride).length > 0 ? JSON.parse(this.optionsOverride) : {}
+    options = {...options, ...custom_options}
 
     this.cleanupBeforeInit() // in case improperly torn down
     this.pluginMainEl = this.selectTarget // required because this.selectTarget is unavailable on disconnect()
--- a/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
+++ b/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
@@ -6,11 +6,12 @@ const select2SelectedPreviewSelector = ".select2-selection--single"
 const select2SearchInputFieldSelector = ".select2-search__field"
 
 export default class extends Controller {
-  static targets = [ "select", "superSelectConfig" ]
+  static targets = [ "select" ]
   static values = {
     acceptsNew: Boolean,
     enableSearch: Boolean,
     searchUrl: String,
+    select2Options: String,
   }
   
   // will be reissued as native dom events name prepended with '$' e.g. '$change', '$select2:closing', etc
@@ -26,6 +27,11 @@ export default class extends Controller {
 
   get isSelect2LoadedOnWindowJquery() {
     return window?.$?.fn?.select2 !== undefined
+  }
+
+  get optionsOverride() {
+    if (!this.hasSelect2OptionsValue) { return {} }
+    return this.select2OptionsValue
   }
 
   connect() {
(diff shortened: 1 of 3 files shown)


pascallaliberte:
@gazayas 

I think overall the approach of a per-instance way to declare select2 options overrides is a good one. I also liked the [`extend` approach you had mentioned in the original issue](https://github.com/bullet-train-co/bullet_train/issues/783#issuecomment-1559324186), but I think that'd be a second step, and we can combine the two (see below).

That said, I'd like to suggest a few tweaks. Let me know if you considered these already and it's why you chose your implementation (I'm talking in particular about the custom target you added).

1. A way to pass the select2 options directly from the `render` call to avoid ejecting the partial. In this case, in the erb partial,  we could add a new param for the partial called `select2_options` that expects json. I'm calling it like this because if we eventually switch to a different plugin than `select2`, we'll inform people to use a different param name, and the old one will be easy to search for and remove. I'm also calling it "options" instead of configs because that's how the select2 docs call these (options).
2. That `select2_options` would then, if not empty, be passed to the `fields--super-select` stimulus controller by way of a `value` param in a data attribute in `wrapper_options[:data]`. Let's call that data attribute `fields--super-select-select2-options-value` and in the Stimulus controller that would be a value called `select2Options` of type `Object`.
3. In the Stimulus controller, you'd check for `this.hasSelect2OptionsValue` and if found, merge them into the `options` using a splat operator like you have [on line 49](https://github.com/bullet-train-co/bullet_train-core/blob/bfe9a4fe2af594a9901210001e10b4d3d8a4f27a/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js#L49).
4. I'd however consider pushing line 49 further down so this `select2OptionsValue` would act as an override. It's sometimes dangerous to allow overrides of `objects` (it's dangerous when options are declared as nested objects as in chart.js for example), but in this case it isn't really a problem. So i'd push it down below `options.width = 'style';`

## Pure JSON, no JS callbacks or functions, so an option to `extend`

The caveat to this is that this new `select2_options` partial param will only work with pure JSON that doesn't include any javascript callbacks, so people won't be able to include a custom `options.ajax` for example.

For those situation, allowing people to extend the `super_select_controller` and declare their own options would be better. That would mean that they'd also need to eject the `_super_select.html.erb` partial to rename it, but I think that's going to happen infrequently enough to be expected.

To help with that possibility, this is what I'd do.

1. Keep the `select2OptionsValue` described above, but have it read via a `get optionsOverride()` getter function (below).
2. If people want to extend the `super_select_controller`, all they'd need would be to re-implement this `get optionsOverride()` in their own controller.

```
get optionsOverride() {
  if (!this.hasSelect2OptionsValue) { return {} }
  return this.select2OptionsValue
}
```

```
// replacing line 49, which would be pushed further down
options = {...options, ...this.optionsOverride} // no () needed when calling the getter
```

```
// extended_super_select_controller.js
import SuperSelectController from "@bullet-train/fields"

export default class extends SuperSelectController {
  get optionsOverride() {
    return {
      ajax: {
        url: this.searchUrlValue,
        dataType: 'json',
        data: function(params) {
          // custom implementation
        }
      }
    }
  }
}
```

gazayas:
Hey @pascallaliberte, I love all of this feedback, I'll try to respond to the best of my ability.

Concerning the first tweak you suggested:
> 1. A way to pass the select2 options directly from the render call to avoid ejecting the partial.

I originally wanted to avoid passing this to the partial because I thought the html would get clunky if we passed in a raw JSON object to the render call. However, in this PR I eventually pass the JSON within the `_super_select.html.erb` partial itself to a content tag anyways, so yeah we can pass it to the render call directly and then just keep the helper that I wrote in https://github.com/bullet-train-co/bullet_train/pull/799. It seems like the controller override is a better option, so I think I'll try to do that.

> I'm also calling it "options" instead of configs because that's how the select2 docs call these (options).

I also wanted to simply call it `options`, but I know we already use `options` in the render calls and I didn't want to make it too loose as to avoid confusion. `select2_options` would definitely be better in this situation.

As far as 2-4, I'm totally cool with those suggestions too.

## Overriding options
I'll try to work on overriding the controller, I'll push something if I make some progress! Just to double check, if we go with the controller override, we wouldn't need to pass JSON via the render call and we would just pull it from the markup in the stimulus controller div tag, right?

pascallaliberte:
> I originally wanted to avoid passing this to the partial because I thought the html would get clunky if we passed in a raw JSON object to the render call.

Yeah, I see what you mean about being clunky. But I don't think something like this is so bad... you feel it would be of bad taste to, say, put something like this in the docs?

```erb
<%= render 'shared/fields/super_select', method: :forwarder_id,
  options: { include_blank: true},
  html_options: {},
  choices: @shipment.valid_forwarders.map { |membership| [membership.label_string, membership.id] },
  select2_options: {
    allowClear: true,
    debug: true
  }
   %>  
```

And then inside the `super_select.html.erb` partial:

```ruby
select2_options ||= nil
# ...
wrapper_options[:data]["#{stimulus_controller}-select2-options-value"] = select2_options.to_json if select2_options.present?
```

gazayas:
I see, I can go ahead and push some commits with these changes and then we'll see what Andrew thinks too!

jagthedrummer (approved):
@gazayas and @pascallaliberte, this seems good to me but I don't know a lot about the BT implementation of super select. Are y'all still feeling good about it?

gazayas:
I'm still down!

PR #311, opened 2023-05-29
Use an empty transformer to replace legacy methods

gazayas:
There were some methods in the OAuth provider scaffolder named `legacy_*` which are duplicates of what we have in the transformer. I found that by just instantiating an empty transformer, we can have access to all of those methods without tainting the files we're actually we're generating with this scaffolder.

I deleted the TODO in `bullet_train-super_scaffolding/lib/scaffolding/script.rb` because if you look at the old [tailwind repository](https://github.com/bullet-train-co/bullet-train-tailwind-css/blob/3b35c5c72342086153e058ee310567e89cc80424/bin/super-scaffold#L12-L13), you can see that the same TODO is referring to those legacy methods.


jagthedrummer:
@gazayas, if this is something we still want can you look at the conflict?

gazayas:
@jagthedrummer Done, had to implement the change from #249 to the empty scaffolder [here](https://github.com/bullet-train-co/bullet_train-core/pull/311/files#diff-1ce907ed12d58ba58732adf82f7298842f98f79d464be28e32b59e24fc2358cbR150)

jagthedrummer (approved):

PR #312, opened 2023-06-05
Implement Masamune

gazayas:
This is a small taste of what [Masamune](https://github.com/gazayas/masamune-ast) can do to replace content in files.

I'll try to explain what's going on here:
```ruby
File.open("./app/helpers/application_helper.rb", "r") do |f|
  msmn = Masamune::AbstractSyntaxTree.new(f.readlines.join)
  current_theme = msmn.symbols.first[:token]
  new_content = msmn.replace(type: :symbol, old_token: current_theme, new_token: theme_name)
end
```

After making the `msmn` object, we grab all the symbols from the file. There's only one, so we call `msmn.symbols.first`.

```ruby
msmn.symbols.first
#=> {:line_number=>5, :index_on_line=>5, :token=>"foo"}
```
Here we get the token of the symbol and the line number it's on, along with how far down the line it is (`index_on_line`).

(The result here is a simple hash, but these results are always derived from a node class. In this case, it comes from `Masamune::AbstractSyntaxTree::SymbolLiteral` which you can find in `msmn.node_list`)

In the original code, we were looking for the Symbol with this regular expression which will break if the spacing is off:
```ruby
current_theme_regexp = /(^    :)(.*)/
```

I personally think pulling the Symbol from the abstract syntax tree will be more reliable down the road, so I hope we can implement more code with Masamune to ensure our file replacement logic grabs and changes the correct variables/symbols/etc. in a more precise manner.

## Potential use in the Transformer
I tried looking for ways to implement this in the Transformer since there's a lot of file replacement going on there, but I'm not quite sure where to start. I definitely would like to see what you think about this one though, and if you have any ideas as to where we could use Masamune in the Transformer (or anywhere else for that matter) as well.


gazayas:
## Finding namespaces with Masamune
I decided to add another short snippet with Masamune.

This time in the Routes File Manipulator, we find all `namespace` invocations by grabbing all lex nodes which match the token `"namespace"`.

We then use `msmn.symbols` to find the first symbol right after the namespace on the same line.

## Lex Nodes
In this example I use `LexNode` instances which add another helpful layer to accessing tokens in the source code.

For example, comments are simply considered to be void statements in the original abstract syntax tree, and the actual comment string isn't provided. However, you can find [comment information in the lex node list](https://github.com/gazayas/masamune-ast/pull/27). For that reason, I'm hoping it'll be helpful to use both the original ast along with lex nodes.

gazayas:
You can see in the last comment how I used Lex Nodes to find `namespace` tokens, but now I've updated the code so it's more specific.

```ruby
namespaces = msmn.method_calls.select { |method| method[:token] == "namespace" }
```

This will ensure that we only get methods, and that we only get the `namespace` method.

gazayas:
[Here is another example](https://github.com/bullet-train-co/bullet_train/pull/801#issuecomment-1595580658) of Masamune, this time finding gems and comments and removing them from the Gemfile.

jagthedrummer:
@gazayas, what are your thoughts on this currently? Do you think Masamune is something we should use? And/or is it something you want to support long-term? (It does look pretty handy!)

gazayas:
@jagthedrummer Although there isn't currently a super strong use-case for it, I think it can give us more confidence when editing source code files during super scaffolding.

This PR is more of a proposal to generate discussion, to have core team members try it out, and to see if we want to use it.

I shared my thoughts in [this article](https://dev.to/gazayas/analyzing-and-editing-ruby-source-code-in-bullet-train-3l8) concerning how I feel about it, but I'm okay with waiting and seeing what everyone else thinks.

I'm personally down to use it, as long as it doesn't bog down readability 🙂

bullet_train-themes-light/lib/tasks/application.rb
```diff
@@ -148,26 +150,26 @@ def self.release_theme(original_theme_name, args)
...
+        new_content = nil
+        File.open("./app/helpers/application_helper.rb", "r") do |f|
+          msmn = Masamune::AbstractSyntaxTree.new(f.readlines.join)
+          current_theme = msmn.symbols.first[:token]
+          new_content = msmn.replace(type: :symbol, old_token: current_theme, new_token: theme_name)
+        end
+
+        File.open("./app/helpers/application_helper.rb", "w") do |f|
+          f.puts new_content
+        end
```
kaspth, line 163: What do you think about using a `Pathname` here? (it would need a `require "pathname"` too), but like this:

```ruby
helper = Pathname.new("./app/helpers/application_helper.rb")

msmn = Masamune::AbstractSyntaxTree.new helper.readlines.join
current_theme = msmn.symbols.first[:token]

helper.write msmn.replace(type: :symbol, old_token: current_theme, new_token: theme_name)
```

kaspth (approved):
Nice, seems good to me!

Great job on Masamune! Let me know if you want me to look into using YARP internally, since that's Ruby's default parser now.

gazayas:
@kaspth Thank you for the review! I didn't know `Pathname` could do all of that! I went ahead and implemented those changes as you suggested.

Also, you're right about the symbol search not being entirely safe. I've edited it so that we first grab the line number of the method definition, and then find the first symbol that appears after it:

```ruby
current_theme_def_line_no = msmn.method_definitions(name: "current_theme").pop[:line_number]
current_theme = msmn.symbols.find { |node| node[:line_number] > current_theme_def_line_no }[:token]
```

I ideally would like to grab method definition BLOCKS and then search for the symbols within the scope of the block itself, but I haven't implemented that in Masamune yet.

With this code at least, you can put symbols before and after the method definition and it will grab the symbol we want properly:

```
> rake bullet_train:themes:light:eject[foo]
> rake bullet_train:themes:light:install
```

(As a side note, the eject command is currently broken against `main`)
## YARP
I haven't done my research on YARP at all, which is no bueno :weary: Any work to improve parsing through Masamune is totally welcome though, and I'm flexible as long as we are replacing our code in a readable and efficient manner!

kaspth (approved):

jagthedrummer:
> ```
> > rake bullet_train:themes:light:eject[foo]
> > rake bullet_train:themes:light:install
> ```
> 
> (As a side note, the eject command is currently broken against `main`)

@gazayas I just tried this with my starter repo linked to the `main` branch of `core` and both commands seemed to work. And they also work for me on this branch. If it's broken for you let's start an issue about that.

PR #309, opened 2023-05-26
Make the pricing page use the correct theme

jagthedrummer:
This makes the pricing layout use the same theme that the rest of the app uses.

Fixes: https://github.com/bullet-train-pro/bullet_train-billing/issues/8

PR #313, opened 2023-06-07
Apply application theme colors to pricing layout

gazayas:
Continuation of #309.

I changed a couple of the styles, let me know what you think!
@jagthedrummer

Looks like this now with emerald (There are a couple of solid lines around the bg gradient, not sure if we want to change it though)

image


jagthedrummer:
That looks great! Let's do it!

gazayas:
Thank you!

jagthedrummer (approved):

PR #315, opened 2023-06-08
Fix request safety issue in set_locale

3pns:
See https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests

Setting the locale using I18n.locale = 'es' during the request/response lifecycle can leak to other request/responses in the same thread.

Using I18n.with_locale instead limit the locale change to the underlying block and does not leak the value of I18n.locale to other requests. It is also the recommended way to handle locale change as per the rails doc.

### DIFF
--- a/bullet_train/app/controllers/concerns/controllers/base.rb
+++ b/bullet_train/app/controllers/concerns/controllers/base.rb
@@ -85,15 +85,14 @@ module Controllers::Base
     end
   end
 
-  def set_locale
-    I18n.locale = [
+  def set_locale(&action)
+    locale = [
       current_user&.locale,
       current_user&.current_team&.locale,
       http_accept_language.compatible_language_from(I18n.available_locales),
       I18n.default_locale.to_s
     ].compact.find { |potential_locale| I18n.available_locales.include?(potential_locale.to_sym) }
-    yield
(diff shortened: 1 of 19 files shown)


bullet_train/app/controllers/concerns/controllers/base.rb
```diff
@@ -85,15 +85,14 @@ def enforce_invitation_only
...
+  def set_locale(&action)
+    locale = [
       current_user&.locale,
       current_user&.current_team&.locale,
       http_accept_language.compatible_language_from(I18n.available_locales),
       I18n.default_locale.to_s
     ].compact.find { |potential_locale| I18n.available_locales.include?(potential_locale.to_sym) }
-    yield
-    I18n.locale = I18n.default_locale
+    I18n.with_locale(locale, &action)
```
gazayas, line 95: LGTM :+1:

Should we set `around_action :set_locale` here too as per the docs?

gazayas: 🤦‍♂️ Sorry about that, didn't even notice.

gazayas (approved):

jagthedrummer (approved):

PR #319, opened 2023-06-10
Use standard Ruby syntax for leftover files in standard.yml

gazayas:


gazayas:
I'm somewhat confused with this one because all of these are passing locally. It looks like the inline comments to disable Standard Ruby aren't working.

gazayas:
Not seeing anything about CI integration in the [docs](https://github.com/standardrb/standard#ignoring-errors).

gazayas:
We rescue the following in `uri_filtering.rb`
```ruby
rescue IPAddr::InvalidAddressError, ArgumentError # standard:disable Lint/ShadowedException
```

However, the [Rubocop documentation](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/ShadowedException) states the following:
> An exception is considered shadowed if it is rescued after its ancestor is, or if it and its ancestor are both rescued in the same ‘rescue` statement. In both cases, the more specific rescue is unnecessary because it is covered by rescuing the less specific exception.

Here we can see `IPAddr::InvalidAddressError` is derived from `ArgumentError`:
```ruby
irb(main):001:0> IPAddr::InvalidAddressError.ancestors
=> 
[IPAddr::InvalidAddressError,
 IPAddr::Error,
 ErrorHighlight::CoreExt,
 ArgumentError,
 StandardError,
 Exception,
 ActiveSupport::Dependencies::RequireDependency,
 ActiveSupport::ToJsonWithActiveSupportEncoder,
 Object,
 RequireAll,
 PP::ObjectMixin,
 ActiveSupport::Tryable,
 JSON::Ext::Generator::GeneratorMethods::Object,
 DEBUGGER__::TrapInterceptor,
 Kernel,
 BasicObject]
```

Therefore, I removed the more specific class, `IPAddr::InvalidAddressError`.

gazayas:
I also left the [rule to disable Lint/UselessAssignment](https://github.com/bullet-train-co/bullet_train-core/blob/350023aa4c3116f4df89c85f6dd2d7dbf8261bbf/bullet_train-api/config/routes.rb#L5) in `bullet_train-api/config/routes.rb` because we Super Scaffold `collection_actions` to the routes file:

https://github.com/bullet-train-co/bullet_train-core/blob/350023aa4c3116f4df89c85f6dd2d7dbf8261bbf/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb#L340

PR #322, opened 2023-06-17
Remove `showcase-rails` as dependency in bullet_train

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/801

Showcase gives us some issues when trying to set up things in API mode, so I plan on moving it to the development group in the Gemfile in the starter repository.


jagthedrummer:
Closing this since we're not focusing on an API mode for now.

PR #323, opened 2023-06-20
Move sessions controller logic into the base concern.

andrewculver:


bullet_train/app/controllers/sessions_controller.rb
```diff
@@ -1,35 +1,3 @@
 class SessionsController < Devise::SessionsController
```
gazayas, line 1: I was able to override the controller in the starter repository with the same declaration here (along with including `Sessions::ControllerBase`), so everything's working fine on my end!

PR #325, opened 2023-06-20
Add format to number partial

### LINKED ISSUE #324: Add formatting for `number` partial
Context: https://discord.com/channels/836637622432170028/836637623048601633/1120579068149108846

gazayas:
Closes #324.

We can now do the following (in this case if the attribute is a float):
```erb
<%= render 'shared/attributes/number', attribute: :loan, format: "$%.2f" %>
```

Result:
```
$500.00
```


bullet_train-themes/app/views/themes/base/attributes/_number.html.erb
```diff
@@ -1,10 +1,15 @@
...
+<% format ||= nil %>
 <% options ||= {} %>
 
 <% if object.send(attribute).present? %>
   <%= render 'shared/attributes/attribute', object: object, attribute: attribute, strategy: strategy do %>
-    <%= object.send(attribute) %>
+    <% if format %>
+      <%= sprintf(format, object.send(attribute)) %>
+    <% else %>
+      <%= object.send(attribute) %>
```
olbrich, line 12: I recommend using `public_send` instead of `send` whenever possible. Most of the time it will be functionally equivalent, but it might help avoid some subtle bugs.

gazayas: Sure, since we use it in different places across the repository, I would prefer to change everything at once as opposed to just applying it to this PR alone.

I'm not too familiar with the bugs or subtle differences, would you mind elaborating a bit? Thank you!

newstler: @gazayas I had this PR changing `send` to `public_send` anywhere I could find. But I have no idea why the test is failing there https://github.com/bullet-train-co/bullet_train-core/pull/307

olbrich: I'd be happy to help sort out the test, but I'm honestly not sure how to run that test locally.

newstler: I remember having this same issue before and @andrewculver managed to fix it somehow :)

gazayas: @newstler No problem, I'm glad I could help! I also went ahead and used `public_send` in this PR :smiley:

PR #328, opened 2023-06-20
Remove unneeded lines from docs

gazayas:
I'm assuming this refers to generating a migration which file attachments don't require, but I don't think we actually need this because you can scaffold a crud-field document with the other steps written here, and there's no command/code written after this particular sentence, so I decided to delete it.

### DIFF
--- a/bullet_train/docs/field-partials/file-field.md
+++ b/bullet_train/docs/field-partials/file-field.md
@@ -8,8 +8,6 @@ In addition, Bullet Train has integrated the direct-uploads feature of Active St
 
 ## Example
 
-Add a 'document' file as an attachment to a `Post` model:
-
 Add the following to `app/models/post.rb`:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


bullet_train/docs/field-partials/file-field.md
```diff
@@ -8,8 +8,6 @@ In addition, Bullet Train has integrated the direct-uploads feature of Active St
 
 ## Example
 
-Add a 'document' file as an attachment to a `Post` model:
```
jagthedrummer, line 11: I think the purpose of this line was to describe in a general way what the example is about, but I don't think it was worded all that well. I propose instead of removing it that we change it to something like:

The following steps illustrate how to add a `document` file attachment to a `Post` model.

jagthedrummer (changes requested):
Made an alternate suggestion here. What do you think, @gazayas?

bullet_train/docs/field-partials/file-field.md
gazayas: Yes, I think saying "The following steps illustrate..." which is descriptive as opposed to saying "Add this" will be easier in reading the flow of the docs here. Adding now

jagthedrummer (approved):

PR #332, opened 2023-06-29
Adding missing themes translations

3pns:
This replace hardcoded strings with i18n calls for several themes partials.


gazayas:
Awesome! I'm personally ok with the `themes` prefix, but I'll take a little bit more time to review.

bullet_train-themes-light/app/views/themes/light/conversations/_card.html.erb
```diff
@@ -7,7 +7,7 @@
       <img src="<%= membership_profile_photo_url(conversation.last_message&.membership || current_membership)  %>" title="<%= current_membership.name %>" alt="<%= current_membership.name %>" class="md:w-14 md:h-14 w-12 h-12 rounded-full mx-auto border-1 border-white border-2 flex-none">
     </div>
     <div class="w-4/5 md:w-full lg:w-4/5 pl-4 pr-1 flex-initial truncate">
-      <p class="text-xs text-slate-300 text-right hover:no-underline"><%= "#{time_ago_in_words conversation.last_message.created_at} ago" unless conversation.last_message.nil? %></p>
+      <p class="text-xs text-slate-300 text-right hover:no-underline"><%= "#{t('themes.time_ago', time: time_ago_in_words(conversation.last_message.created_at))}" unless conversation.last_message.nil? %></p>
```
olbrich, line 10:
```suggestion
      <p class="text-xs text-slate-300 text-right hover:no-underline"><%= "#{t('themes.time_ago', time: time_ago_in_words(conversation.last_message.created_at))}" if conversation.last_message.present? %></p>
```
Maybe slightly more readable?

bullet_train-themes-light/app/views/themes/light/conversations/_message.html.erb
```diff
@@ -126,7 +126,7 @@
             <path d="M0 57.7c.1 145.1 4.1 224 14.5 281.1 19.4 106.3 65.2 156.4 166.3 181.7 48.7 12.1 111.4 18.8 211.7 22.5 23.8.9 196.6 2.8 268.2 3H683l.2 67.8.3 67.8 170.3-102.1L1024 477.3l-170.2-102-170.3-102-.3 67.8-.2 67.9h-75.8c-184.7 0-266.9-4.8-327.2-19-34.2-8.1-62.8-21.1-81.2-37-36.1-31.3-52.3-76.8-58.8-165.5-2-28-3-69.2-3-127.8V0H0v57.7z"/>
           </svg>
           <div class="flex flex-col mr-10 min-w-max">
-            <div class="pt-0 mr-2 ml-4 text-xxs"><%= reply.membership.label_string %> reply: </div>
```
gazayas, line 129: Any reason to the colon outside of the locale? One thing that comes to mind is full-width text in Japanese where "：" is different from what we have here, ":". I'd rather be safe and include this in the locale.

gazayas (commented):
Looking over things, I'm totally okay with the `themes` prefix in translation keys. Thanks for working this one out!

bullet_train-themes-light/app/views/themes/light/conversations/_message.html.erb
3pns: The intent was to make it reusable. I changed the key to themes.light.conversations.message.reply so it is view-specific and included the ":"

3pns:
I implemented the requests from @pascallaliberte and @gazayas  : 
- split keys per themes rails engine
- split keys in between navigation/fields/action_models sub prefix
Please let me know your opinion.

bullet_train-themes-light/app/views/themes/light/conversations/_message.html.erb
gazayas: Got it!

bullet_train-themes-light/config/locales/themes/navigation.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  themes:
+    navigation:
```
gazayas: I'm actually fine if we move all of these to `bullet_train-themes` since they're not scoped under `light`, but if we keep it here maybe we can add `light:` to the hierarchy, or just put these locales inside `bullet_train-themes-light/config/locales/themes/light.yml`.

gazayas (commented):
Besides the one comment about `light`, everything here works for me!

bullet_train-themes-light/config/locales/themes/navigation.yml
pascallaliberte: I'm thinking the following:

* I'd keep the locale files throughout each package, and like you did @3pns to have the translation string used in that package in a locale file (`navigation.en.yml`, `fields.en.yml`, etc) within that package. So disagreeing with @gazayas a little bit here. And my argument follows the next point...
* I'd remove the `themes` grouping entirely.

I feel these translation files aren't for a `themes` concept. They're just UI strings in partials that are bundled with the theme packages. A `themes.en.yml` would be great if BT had a UI for picking the theme settings (e.g. to change the accent color to a custom color just for your account). But the theme includes strings used for navigation (and that goes in a `navigation.en.yml` in the package that defines it, etc), for fields, for specific private gems (action_models), and so on...

All these strings are also overridable in the main app. It's easy to create a `navigation.en.yml` in the main app, and override just the `close_app_menu` entry. We don't know that it was in a theme-specific package, only that I'm modifying that UI string in my own app.

pascallaliberte:
@3pns I'd include `.en` in the file name, to follow the pattern of the main repo, even if the file is inside a `en/` folder. What do you think?

bullet_train-themes-light/config/locales/themes/navigation.yml
pascallaliberte: That means that:
* in the case of `light.yml`, its strings would be in a `conversations.en.yml`
* in the case of `time_ago`, that would simply be in a `en.yml` at the base of the `en/` directory in that package.

pascallaliberte:
Thanks for doing this work @3pns !

bullet_train-themes-light/config/locales/themes/navigation.yml
gazayas: I see where you're coming from about using `themes`. Like I said I'm open either way, I think I just prefer it all to be unified so it's all grouped together in a meaningful way, so I agree with the statement you made here:

> in the case of light.yml, its strings would be in a conversations.en.yml

If we do this, we don't have to worry about using `light` in our keys. Why don't we do that so this one can be ready to merge?

@3pns Sorry to have you make all of these changes! Can we remove `themes` and `light` from the keys you wrote up and just use general files like `conversations.en.yml`? If we do that then I'll go ahead and approve the changes.

3pns:
@gazayas @pascallaliberte I updated the PR taking into account the new conventions we discussed during the past few days. Let me know if I missed something!

gazayas (approved):
:muscle:

pascallaliberte (approved):
This is great

jagthedrummer:
@3pns, this looks great and I'd love to merge it. Can you look at resolving the conflicts on this one?

3pns:
@jagthedrummer Done ! Thanks for the follow up.

3pns:
@jagthedrummer I fixed the conflict, please note there is a knapsack pro token error in the starter repo minitest suite.

jagthedrummer:
Thanks, @3pns! I think the test failure has something to do with running the tests from your branch in your own repo vs running from a branch directly in the `core` repo. To confirm that suspicion I push a version of this branch directly to `core` and opened https://github.com/bullet-train-co/bullet_train-core/pull/382, and the tests passed as you can see here: https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train-core/1069/workflows/65102032-1312-49b7-974a-cf15467a85d5

I'm going to merge this PR and close the other one. Thank you for your work on this.

PR #333, opened 2023-06-30
Add invisible_captcha to user registration

julianrubisch:
In my latest BT installation I noticed some bot signups.

I thought that installing invisible_captcha would be a sensible default for all apps.

### DIFF
--- a/bullet_train/lib/bullet_train/engine.rb
+++ b/bullet_train/lib/bullet_train/engine.rb
@@ -19,6 +19,12 @@ begin
 rescue LoadError
 end
 
+begin
+  # And the same holds true for invisible_captcha
+  require "invisible_captcha"
+rescue LoadError
+end
+
 module BulletTrain
   class Engine < ::Rails::Engine
   end
--- a/bullet_train/Gemfile.lock
+++ b/bullet_train/Gemfile.lock
@@ -28,6 +28,7 @@ PATH
       hiredis
       http_accept_language
       image_processing
+      invisible_captcha
       microscope
       nice_partials (~> 0.9)
       pagy
@@ -202,6 +203,8 @@ GEM
     image_processing (1.12.2)
       mini_magick (>= 4.9.5, < 5)
       ruby-vips (>= 2.0.17, < 3)
+    invisible_captcha (2.1.0)
+      rails (>= 5.2)
     json (2.6.3)
     language_server-protocol (3.17.0.2)
     loofah (2.20.0)
--- a/bullet_train/app/controllers/concerns/registrations/controller_base.rb
+++ b/bullet_train/app/controllers/concerns/registrations/controller_base.rb
@@ -1,9 +1,9 @@
 module Registrations::ControllerBase
   extend ActiveSupport::Concern
 
-  invisible_captcha only: :create
-
   included do
+    invisible_captcha only: :create
+
     def new
       # We have to set the session here because Safari wouldn't save it on a redirect to this URL.
       if params[:invitation_uuid]
--- a/bullet_train/app/controllers/concerns/registrations/controller_base.rb
+++ b/bullet_train/app/controllers/concerns/registrations/controller_base.rb
@@ -1,6 +1,8 @@
 module Registrations::ControllerBase
   extend ActiveSupport::Concern
 
+  invisible_captcha only: :create
+
   included do
     def new
       # We have to set the session here because Safari wouldn't save it on a redirect to this URL.
--- a/bullet_train/app/views/devise/registrations/new.html.erb
+++ b/bullet_train/app/views/devise/registrations/new.html.erb
@@ -6,6 +6,8 @@
        <%= render 'account/shared/notices' %>
        <%= render 'account/shared/forms/errors', form: f %>
 
+       <%= invisible_captcha %>
+
        <%= render 'shared/fields/email_field', form: f, method: :email, options: {autofocus: true} do %>
          <% content_for :help do %>
            <%= link_to t('devise.links.have_account'), new_user_session_path %>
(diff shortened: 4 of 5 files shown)


jagthedrummer:
I love the idea of including this, but I'm on the fence about having it turned on by default, and with no obvious way to disable it. Could we make a config option that would allow devs to toggle it on or off via an initializer or something?

@andrewculver @gazayas @kaspth, what do y'all think?

gazayas:
Yeah, I think something like `captcha_enabled?` which checks an environment variable in [bullet_train.rb](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train/lib/bullet_train.rb) would be perfect :ok_hand:

julianrubisch:
Will see to it when I'm back from 🏝️

bullet_train/lib/bullet_train/engine.rb
```diff
@@ -19,6 +19,12 @@
 rescue LoadError
 end
 
+begin
+  # And the same holds true for invisible_captcha
+  require "invisible_captcha"
```
kaspth, line 24: @julianrubisch why do we need to hoist the loading of the gem? It doesn't look like we've got `view_paths` that would conflict like in the case of `devise` and `showcase-rails`.

julianrubisch: If I only remembered! 😅

This PR has been dangling for a while, but I remember there _was_ a reason for it.

Can look into it when I'm back at a computer later this month.

kaspth: I just checked CI and it looks it's that `invisible_captcha` otherwise isn't available in the controller yet, when you do `invisible_captcha only: :create`. 

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train-core/841/workflows/b7e81028-49cf-4fab-9ef9-37148c15ddec/jobs/2589

So it's because we're inserting `invisible_captcha` when `Registrations::ControllerBase` is loaded, which means Bullet Train is loaded before `invisible_captcha`! Ok, that makes sense to me.

kaspth:
@julianrubisch I can also take this over, if you'd rather enjoy your vacation without worrying about it 😄

bullet_train/lib/bullet_train/engine.rb
julianrubisch: Ah yes. Otherwise I probably wouldn't have wasted a commit on it 😅

kaspth:
@julianrubisch I didn't get to this if you're back from vacation and want to follow it up. I can still take a look too!

jagthedrummer:
I pulled this down and have run through signup a few times trying to trigger this and I seem to be getting inconsistent results.

I'm going to the signup page and then inspecting the form and disabling the CSS rules that hide the honeypot field. Then I'm putting something in the honeypot, filling out the rest of the form, and hitting the "Sign Up" button.

CleanShot 2023-09-01 at 11 54 38

Sometimes the registration works like normal and the `invisible_captcha` seems to be missing the fact that I've populated the honeypot.

Other times it seems that `invisible_captcha` does catch that I've populated the honeypot and it blocks sign up, but it doesn't return an error or anything. It appears that the form just doesn't submit.

CleanShot 2023-09-01 at 12 03 50

When this happens I can see an error in the rails console that says:

```
[Invisible Captcha] Potential spam detected for IP 127.0.0.1. Honeypot param 'registration.oqafjyulnzdbevkhr-' was present.
Filter chain halted as #<Proc:0x000000010e1bf030 /Users/jgreen/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/invisible_captcha-2.1.0/lib/invisible_captcha/controller_ext.rb:12> rendered or redirected
```

Weirdly the console also says that the request was successful:

```
Completed 200 OK in 1ms (ActiveRecord: 0.0ms | Allocations: 642)
```

I can see in the browser dev tools that it does return a `200 OK`:

CleanShot 2023-09-01 at 12 04 44

And there's an error in the browser console that says:

```
Error: Form responses must redirect to another location
```

CleanShot 2023-09-01 at 12 07 19

If I then click "Sign Up" again the form gets blanked out (it loses info I filled in) and I get an error saying:

```
Sorry, that was too quick! Please resubmit.
```

CleanShot 2023-09-01 at 12 13 50

So I think something isn't quite right about the integration. I have no idea why it's blocking sign up sometimes, but not all the times when I fill out the honeypot.

I was also reading through the docs for `invisible_captcha` and I think there are at least a couple of config options that we need to figure out how to handle so that people have a reasonably good experience when deploying the app. (Or maybe these are arguments for leaving it disabled by default, and then having some docs about what steps need to be taken to enable it.)

From the docs:
> `secret`: customize the secret key to encode some internal values. By default, it reads the environment variable `ENV['INVISIBLE_CAPTCHA_SECRET']` and fallbacks to random value. Be careful, if you are running multiple Rails instances behind a load balancer, use always the same value via the environment variable.

I think that means that if someone were to deploy to Heroku (for instance) and have their app configured to run on multiple dynos that the captcha would be flaky and might even trigger false positives (that is, block legitimate users from signing up). Some discussion of that possibility [in this issue](https://github.com/markets/invisible_captcha/issues/97).

There's also a [whole section in the docs about running multiple Rails instances](https://github.com/markets/invisible_captcha/issues/97). I think we'd want to have that bit of configuration in the repo and setup to be automatically enabled if the `captcha_enabled?` flag (proposed above) is turned on.

I wouldn't be surprised if the flakiness I'm seeing is related to one of these issues.

It also looks like adding this breaks a few tests, so we'll need to figure out how to either disable this in testing, or we'll need to modify tests to accommodate.

Still love the idea of having this included and easy to turn on for people who want it.

yshmarov:
love this gem, using it for a long time now. 

reasonable defautls that work for me:

```
# config/initializers/invisible_captcha.rb
InvisibleCaptcha.setup do |config|
  config.visual_honeypots = false
  config.timestamp_threshold = 1
  config.timestamp_enabled = !Rails.env.test?
  config.spinner_enabled = !Rails.env.test?
  config.honeypots = %w[subtitle_ic nickname_ic company_ic address_ic website_ic]
end
```

PR #334, opened 2023-07-05
Themes localization

3pns:
The goal here is :
- allow using different Date and DateTime display formats in bullettrain fields.
- simplify Date and DateTime handling in the backend by using ISO8601 instead of a localized date/datetime format as rails and all modern languages and frameworks support iso8601 natively.

This is what I did :
- changed all themes date and datetime related form fields data format from an American localized Date and Datetime Strings to ISO8601, added a hidden display field that is connected to the datetimepicker
- removed existing assignments and scaffolding of assign_date_and_time and assign_date_and_time assignments as Ruby On Rails handle iso8601 parsing natively.
- add datetimepicker translations and fields formats
- kept assign_date_and_time and assign_date for backward compatibility.


gazayas:
Hey @3pns, I know this is a draft PR so let me know if I should wait a bit longer, there were just a couple of things I noticed:
1. `am_pm?` isn't being picked up, but if you move it to `bullet_train/app/helpers/account/dates_helper.rb` it will work.
2. There are still some spots where `options` hasn't been changed to `data_options` or `display_options` in the date-related partials (i.e. - the keyword argument for `render 'shared/fields/field'`, the line which has `unless options[:disabled]`, etc.), causing an error when trying to create a new Super Scaffolded record.

Feel free to ignore if you're already working on these.

Definitely excited to move to Rails' standard use of ISO8601!

gazayas:
Also there was talk a while ago to put all of our formats in a locale file, so I'm glad to see them there.

pascallaliberte:
Please disregard these notes if they were still on your list, I know this is still a draft PR:

1. If `rubyTzNameToUnixTz` can be a ruby-side helper, that will save on the js payload size
2. If `moment` is used elsewhere, and I presume the date range plugin requires it anyway, I'd stick with it for now
3. I saw the `t-value` as a way to pass many translation keys. I'm not sure about... either the name or the fact that it's passing a lot of keys? I guess it beats having a value for each, but a value for each gives you the ability to set defaults. What do you think?
4. I saw the use of `.trigger`. Please make sure that the native `change` event still gets published before you're done. See `super_select_controller.js` if you need to `dispatchNativeEvent` separately.

bullet_train-fields/app/javascript/controllers/fields/date_controller.js
```diff
@@ -3,14 +3,21 @@ require("daterangepicker/daterangepicker.css");
 
 // requires jQuery, moment, might want to consider a vanilla JS alternative
 import 'daterangepicker';
+import moment from 'moment-timezone'
```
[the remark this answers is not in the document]
olbrich: As @pascallaliberte suggests, it's probably better to stick with the existing dependency for now. I didn't realize it was already a dependency.

pascallaliberte:
Some more notes. If I'm missing something please disregard:

1. for `t-daterangepicker`, I see that you're now passing all these strings for customizing the picker UI. How about the name `pickerLocale` which would be named `picker-locale-value` in the HTML and `pickerLocaleValue` in the Stimulus controller. I also see you're making a deep copy so you can customize the values sent down to the picker.
2. Do we still need the `applyButton` and `cancelButton` value params if we have a bigger `pickerLocale` value param ? I'm not sure developers are passing per-picker customizations for the Apply and Cancel buttons.
3. for `tzNameToTzId`, what do you think about the following. If the time zone field isn't shown, that whole translation list isn't needed, and if the time zone field is available (because the user's browser's current time zone differs from the team time zone), then you could read from the time zone select's option fields? It's in keeping with the idea that Stimulus prefers reading state from the HTML when possible.

3pns:
At this stage I believe I have addressed all the concerns raised in the previous messages.
@pascallaliberte I have also reverted the timezone picker UX to its original UX as discussed on Discord.
I am thinking to add a current_time_zone helper. Is acceptable in regard to the scope of this PR or should it be done in a separate issue?

pascallaliberte:
@3pns 

> I am thinking to add a current_time_zone helper. Is acceptable in regard to the scope of this PR or should it be done in a separate issue?

Yeah, go for it. It fits well with this PR I think

pascallaliberte:
@3pns Before we merge, I think it'll be important to make sure an invalid date entered by the user doesn't send a blank value in the iso8601 `field`.

What do you think of these statements?

1. an empty value is valid, so entering a date in the wrong format into the `displayField` can't be caught by developers even if they wanted to
2. the stimulus controllers are where this problem is dealt with: populating the iso8601 `field` with something which can be re-populated into the `displayField` on the form re-displaying from errors.

I'm thinking populating the iso8601 `field` with the raw user-entered value if it's not in the right format, on the server it fails the iso8601 check, and the erroneous value is automatically re-populated into the `displayField` (via the Stimulus controller)?

pascallaliberte:
Two improvements you could make, not required for the merge however:

Replacing these lines:
```
if(!["", null].includes(this.fieldTarget.value)){
  $(this.displayFieldTarget).trigger("apply.daterangepicker");
}
```
with a call to an `ensureDisplayFieldIsUpdated()` method maybe, returning early if the fieldTarget is empty?

Also, the `currentTimeZone()` method would be a good candidate for a `get currentTimeZone()` getter method, then you could access it via `this.currentTimeZone`.

pascallaliberte:
> @3pns Before we merge, I think it'll be important to make sure an invalid date entered by the user doesn't send a blank value in the iso8601 `field`.

After our discussion in Discord, let's drop this requirement.

It's a rare occurrence that a user would add an improper format by hand, and we can address this scenario later through client-side validation, showing a BT-styled error message via Stimulus, maybe via a `<template data-fields--date-target="errorTemplate">` with an empty copy of a BT `:error` partial...

gazayas:
Concerning the API, we can verify the JSON output from the corresponding system tests in the starter repository (specifically under the logic generated from [assert_proper_object_serialization](https://github.com/bullet-train-co/bullet_train/blob/56914548b0a2718300eaa3a733eb9b168afb9214/test/controllers/api/v1/scaffolding/completely_concrete/tangible_things_controller_test.rb#L27)). Here's a look at the data in question with the following example model.

### Example Model
```
rails g model Project team:references title:string due_date:date meeting:datetime && \
bin/super-scaffold crud Project Team title:text_field due_date:date_field meeting:date_and_time_field
```

# `Date` and `DateTime` on `main`
```
[1] pry(#<Api::V1::ProjectsControllerTest>)> project_data
=> {"id"=>31,
 "team_id"=>26,
 "title"=>"MyString",
 "due_date"=>"2023-07-22",
 "meeting"=>"2023-07-22T05:08:18.000Z",
 "created_at"=>"2023-07-21T20:10:16.436Z",
 "updated_at"=>"2023-07-21T20:10:16.436Z"}
[2] pry(#<Api::V1::ProjectsControllerTest>)> project
=> #<Project:0x00007f822e39f000
 id: 31,
 team_id: 26,
 title: "MyString",
 due_date: Sat, 22 Jul 2023,
 meeting: Sat, 22 Jul 2023 05:08:18.000000000 UTC +00:00,
 created_at: Fri, 21 Jul 2023 20:10:16.436178000 UTC +00:00,
 updated_at: Fri, 21 Jul 2023 20:10:16.436178000 UTC +00:00>
[3] pry(#<Api::V1::ProjectsControllerTest>)> project.due_date.class
=> Date
[4] pry(#<Api::V1::ProjectsControllerTest>)> project.meeting.class
=> ActiveSupport::TimeWithZone
```

# `Date` and `DatimeTime` on this Pull Request
```
[1] pry(#<Api::V1::ProjectsControllerTest>)> project_data
=> {"id"=>5,
 "team_id"=>1,
 "title"=>"MyString",
 "due_date"=>"2023-07-22",
 "meeting"=>"2023-07-22T05:14:31.000Z",
 "created_at"=>"2023-07-21T20:15:11.168Z",
 "updated_at"=>"2023-07-21T20:15:11.168Z"}
[2] pry(#<Api::V1::ProjectsControllerTest>)> project
=> #<Project:0x00007f10d069bb98
 id: 5,
 team_id: 1,
 title: "MyString",
 due_date: Sat, 22 Jul 2023,
 meeting: Sat, 22 Jul 2023 05:14:31.000000000 UTC +00:00,
 created_at: Fri, 21 Jul 2023 20:15:11.168665000 UTC +00:00,
 updated_at: Fri, 21 Jul 2023 20:15:11.168665000 UTC +00:00>
[3] pry(#<Api::V1::ProjectsControllerTest>)> project.due_date.class
=> Date
[4] pry(#<Api::V1::ProjectsControllerTest>)> project.meeting.class
=> ActiveSupport::TimeWithZone
```

It's also worth noting that, when Super Scaffolding on `main` and then pointing the gems to this PR, the JSON data we're getting back is the same.

## Grape Swagger
However, if you're running on an older version of our API back when it was Grape Swagger, I suggest checking these tests against that platform. I don't currently have an older project set up to verify, but you could do so under the `assert_proper_object_serialization` method I mentioned earlier.

bullet_train-fields/app/controllers/concerns/fields/date_and_time_support.rb
```diff
@@ -2,6 +2,12 @@ module Fields::DateAndTimeSupport
   extend ActiveSupport::Concern
 
   def assign_date_and_time(strong_params, attribute)
```
gazayas, line 4: I was able to reproduce the error you mentioned, we'll probably have to use another format for this line since the attribute is an iso8601 datetime now:
```ruby
strong_params[attribute] = time_zone.strptime(strong_params[attribute], t("global.formats.date_and_time"))
```

Still looking into it though, I'll try to see what option works best for us here.

gazayas: We could potentially parse the field to a ISO8601 `DateTime` object in and use this format instead.

```ruby
time_zone.strptime(DateTime.parse(strong_params[attribute]).iso8601, '%Y-%m-%dT%H:%S:%L%z')
```

gazayas: We might be able to do something like this until they update their strings like the deprecation your wrote tells them to.

```ruby
iso8601_format = "%Y-%m-%dT%H:%S:%L"
begin
  strong_params[attribute] = time_zone.strptime(strong_params[attribute], iso8601_format)
rescue ArgumentError => e
  # If developers are on >= 1.2.26, the format we get is different.
  iso8601_date = DateTime.strptime(strong_params[attribute], t("global.formats.date_and_time")).iso8601
  strong_params[attribute] = time_zone.strptime(iso8601_date, iso8601_format)
end
```

3pns:
> Concerning the API, we can verify the JSON output from the corresponding system tests in the starter repository (specifically under the logic generated from [assert_proper_object_serialization](https://github.com/bullet-train-co/bullet_train/blob/56914548b0a2718300eaa3a733eb9b168afb9214/test/controllers/api/v1/scaffolding/completely_concrete/tangible_things_controller_test.rb#L27)). Here's a look at the data in question with the following example model.
> 
> ### Example Model
> ```
> rails g model Project team:references title:string due_date:date meeting:datetime && \
> bin/super-scaffold crud Project Team title:text_field due_date:date_field meeting:date_and_time_field
> ```
> 
> # `Date` and `DateTime` on `main`
> ```
> [1] pry(#<Api::V1::ProjectsControllerTest>)> project_data
> => {"id"=>31,
>  "team_id"=>26,
>  "title"=>"MyString",
>  "due_date"=>"2023-07-22",
>  "meeting"=>"2023-07-22T05:08:18.000Z",
>  "created_at"=>"2023-07-21T20:10:16.436Z",
>  "updated_at"=>"2023-07-21T20:10:16.436Z"}
> [2] pry(#<Api::V1::ProjectsControllerTest>)> project
> => #<Project:0x00007f822e39f000
>  id: 31,
>  team_id: 26,
>  title: "MyString",
>  due_date: Sat, 22 Jul 2023,
>  meeting: Sat, 22 Jul 2023 05:08:18.000000000 UTC +00:00,
>  created_at: Fri, 21 Jul 2023 20:10:16.436178000 UTC +00:00,
>  updated_at: Fri, 21 Jul 2023 20:10:16.436178000 UTC +00:00>
> [3] pry(#<Api::V1::ProjectsControllerTest>)> project.due_date.class
> => Date
> [4] pry(#<Api::V1::ProjectsControllerTest>)> project.meeting.class
> => ActiveSupport::TimeWithZone
> ```
> 
> # `Date` and `DatimeTime` on this Pull Request
> ```
> [1] pry(#<Api::V1::ProjectsControllerTest>)> project_data
> => {"id"=>5,
>  "team_id"=>1,
>  "title"=>"MyString",
>  "due_date"=>"2023-07-22",
>  "meeting"=>"2023-07-22T05:14:31.000Z",
>  "created_at"=>"2023-07-21T20:15:11.168Z",
>  "updated_at"=>"2023-07-21T20:15:11.168Z"}
> [2] pry(#<Api::V1::ProjectsControllerTest>)> project
> => #<Project:0x00007f10d069bb98
>  id: 5,
>  team_id: 1,
>  title: "MyString",
>  due_date: Sat, 22 Jul 2023,
>  meeting: Sat, 22 Jul 2023 05:14:31.000000000 UTC +00:00,
>  created_at: Fri, 21 Jul 2023 20:15:11.168665000 UTC +00:00,
>  updated_at: Fri, 21 Jul 2023 20:15:11.168665000 UTC +00:00>
> [3] pry(#<Api::V1::ProjectsControllerTest>)> project.due_date.class
> => Date
> [4] pry(#<Api::V1::ProjectsControllerTest>)> project.meeting.class
> => ActiveSupport::TimeWithZone
> ```
> 
> It's also worth noting that, when Super Scaffolding on `main` and then pointing the gems to this PR, the JSON data we're getting back is the same.
> 
> ## Grape Swagger
> However, if you're running on an older version of our API back when it was Grape Swagger, I suggest checking these tests against that platform. I don't currently have an older project set up to verify, but you could do so under the `assert_proper_object_serialization` method I mentioned earlier.

Can you please confirm/refute my assumptions based on your comments.
- bullet_train-api in its current form is not impacted by this PR as there are no changes in related json input and output.
- the old api system status based on grape/swagger is unknown, however in theory it should not be impacted as grape rely on its own serializing system.

gazayas:
@3pns
> bullet_train-api in its current form is not impacted by this PR as there are no changes in related json input and output.

Correct. Also, the concern you mentioned on Discord about removing `assign_date_and_time` for the access token is fine since the code on `main` actually doesn't process the `last_used_at` attribute as expected.

```
     4: def assign_date_and_time(strong_params, attribute)
     5:   attribute = attribute.to_s
     6:   time_zone_attribute = "#{attribute}_time_zone"
 =>  7:   binding.pry
     8:   if strong_params.dig(attribute).present?
     9:     time_zone = ActiveSupport::TimeZone.new(strong_params[time_zone_attribute] || current_team.time_zone)
    10:     strong_params.delete(time_zone_attribute)
    11:     strong_params[attribute] = time_zone.strptime(strong_params[attribute], t("global.formats.date_and_time"))
    12:   end
    13: end

[1] pry(#<Account::Platform::AccessTokensController>)> attribute
=> "last_used_at"
[2] pry(#<Account::Platform::AccessTokensController>)> strong_params.dig(attribute).present?
=> false
[3] pry(#<Account::Platform::AccessTokensController>)> strong_params.each_value {|val| p val}
"Default Access Token"
=> #<ActionController::Parameters {"description"=>"Default Access Token"} permitted: true>
```

`last_used_at` is not included in the [strong_params](https://github.com/bullet-train-co/bullet_train-core/blob/ceee201286bdbd9219b1c8fd199df0418d7c1139/bullet_train-api/app/controllers/api/v1/platform/access_tokens_controller.rb#L37-L50), and whether this is intentional or not, I would have to confirm with @andrewculver. I would say as far as the scope of this PR though, you should be clear to leave `assign_date_and_time` deleted for the access token.

> the old api system status based on grape/swagger is unknown, however in theory it should not be impacted as grape rely on its own serializing system.

Correct, and as far as I can tell the values that are being serialized are still uniform, so the API should be okay.

pascallaliberte (approved):

gazayas (approved):

bullet_train-outgoing_webhooks/app/controllers/account/webhooks/outgoing/deliveries_controller.rb
```diff
@@ -63,17 +63,13 @@ def destroy
...
+    params.require(:webhooks_outgoing_delivery).permit(
       :event_id,
       :endpoint_url,
       :delivered_at,
       # 🚅 super scaffolding will insert new fields above this line.
       # 🚅 super scaffolding will insert new arrays above this line.
     )
-
-    assign_date_and_time(strong_params, :delivered_at)
     # 🚅 super scaffolding will insert processing for new fields above this line.
```
kaspth, line 73: I think we've broken super scaffolding's ability to insert lines since the code expects a `strong_params` local variable: https://github.com/bullet-train-co/bullet_train-core/pull/334/files#diff-eda613639ab27832251c13ff26e97c46614576602167c5be468c0a5639160d3dR1111

@jagthedrummer @gazayas do either of you know if super scaffolding can/will insert lines in this controller?

jagthedrummer: @kaspth I honestly don't know enough about how super scaffolding works under the hood to be able to make a confident assessment. But based on the line you linked to (and given my limited context) I'd say that you're likely right that removing the `strong_params` local variable will have some unintended consequences. Is there a good/easy way to write a test for that? (Both to confirm or disconfirm your hunch, and to make sure we don't break it in the future.)

kaspth: > Is there a good/easy way to write a test for that?

I'm not sure. @gazayas do you know how to write Super Scaffolding tests for something like the above?

gazayas: @kaspth @jagthedrummer
Thanks for pointing this out! I see what you're saying, so I have a couple of questions about how to handle this, and depending on that we can either add `strong_params = ...` back or keep the current changes.

## Description
We recently merged this PR: #330.

Since we've been slowly moving Bullet Train logic out of the starter repository and into packages, I feel like sometimes there is cruft that gets leftover that we simply moved over as-is because the code was working, which is why I was ok with deleting the TODO in the PR mentioned above. The strong parameters here for example were originally just in the API controllers in the starter repository if my memory serves me correctly. For all other scaffolded helpers like `assign_date` here, you're right, we would need `strong_params` for any other scaffolded methods to work. Since there are no helpers here anymore though, I don't think we need it. The `strong_params` handlers are pretty much the same across the board for different controllers, so I can understand if this was just copied over to get things working for the time being.

## My initial thoughts on the change
The current code is fine as-is because we can't Super Scaffold to `deliveries_controller.rb` since it's in `bullet_train-core`, and it does not serve as a template which is copied over like `tangible_things_controller.rb`, etc.

## Questions
However, technically speaking, a developer could eject this file to their own project if they wanted to.
1. Do we want to be able to Super Scaffold to this file after a developer were to eject it?
2. Also, we have `bin/super-scaffold incoming-webhooks` but nothing for outcoming webhooks. Should we add a scaffolder for outcoming webhooks? If so, I can write tests for it. I can't say I'm too familiar with that end of things to make an executive decision concerning it, but if we need it I'd be glad to look into and work on it.

jagthedrummer: 🤔 I'm gonna need to absorb this and think about it a bit.

gazayas: Same here, I saw that you shared a similar thought in #385 so I'll try to keep that in mind too.

kaspth: I didn't expect this to generate quite this much context, which is ok! But I'm thinking we could just leave this as is for now and revisit later. Like @gazayas mentioned the controller is unlikely to be super scaffoldable.

jagthedrummer: @gazayas @kaspth, just so you know I ended up reverting this PR in https://github.com/bullet-train-co/bullet_train-core/pull/389. It was breaking tests in the starter repo and it seems like it had legitimately broken date and date/time pickers. I was able to generate a revert PR without a ton of trouble, so I figured a rollback was better than trying to sort out exactly what was wrong while everything remained in a broken state.

I generated this PR to revert the revert so that someone can pick back up with this: https://github.com/bullet-train-co/bullet_train-core/pull/390

PR #335, opened 2023-07-10
Insert load hooks for collaborator gems

kaspth:
This means collaborator gems, particularly the bullet_train-billing(-*) ones, can hook in and help define extra functionality. Here's how that would look:

```ruby
module BulletTrain::Billing::Teams::Base
  extend ActiveSupport::Concern

  included do
    has_many :billing_subscriptions, class_name: "Billing::Subscription", dependent: :destroy, foreign_key: :team_id
  end
end

ActiveSupport.on_load(:bullet_train_teams_base) { include BulletTrain::Billing::Teams::Base }
```

Collaborator gems must use concerns otherwise features won't be included into the class that includes `::Teams::Base`. That's because regular Ruby modules don't support this kind of deep nesting but concerns have specifically added that.

Note: this is also how Rails works internally and expects you to hook into its libraries, like Active Record.

### How to ship this

- [x] Merge this PR.
- [ ] Merge https://github.com/bullet-train-pro/bullet_train-billing/pull/19
- [ ] Merge https://github.com/bullet-train-pro/bullet_train-billing-stripe/pull/10
- [x] Merge bullet_train-billing-usage PR (not needed after all!).
- [ ] Open a new PR to `bullet_train-core` that removes all the billing related extensions in `Teams::Base`, then merge.
- [ ] Release new gem versions.

PR #337, opened 2023-07-10
Remove unused `Billing::Usage::TeamSupport` include

kaspth:
`TeamSupport` was renamed to `HasTrackers` in https://github.com/bullet-train-pro/bullet_train-billing-usage/commit/e4d3aedaae06d10d8700df747c0b2d79cfb2a3d8

Additionally, the guide recommends adding `Billing::HasTrackers` (the guide needs updating to `Billing::Usage::HasTrackers` too) directly.

So we aren't expecting this code to run and we can remove it.

### DIFF
--- a/bullet_train/app/models/concerns/teams/base.rb
+++ b/bullet_train/app/models/concerns/teams/base.rb
@@ -27,10 +27,6 @@ module Teams::Base
       if defined?(Billing::Stripe::Subscription)
         has_many :billing_stripe_subscriptions, class_name: "Billing::Stripe::Subscription", dependent: :destroy, foreign_key: :team_id
       end
-
-      if defined?(Billing::Usage::TeamSupport)
-        include Billing::Usage::TeamSupport
-      end
     end
 
     # validations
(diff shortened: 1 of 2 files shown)


bullet_train/app/models/concerns/teams/base.rb
```diff
@@ -27,10 +27,6 @@ module Teams::Base
       if defined?(Billing::Stripe::Subscription)
         has_many :billing_stripe_subscriptions, class_name: "Billing::Stripe::Subscription", dependent: :destroy, foreign_key: :team_id
       end
-
-      if defined?(Billing::Usage::TeamSupport)
-        include Billing::Usage::TeamSupport
-      end
```
kaspth, line 33: Alternatively, if we want to fire these lines we can update to this:

```ruby
     if defined?(Billing::Usage::HasTrackers)
        include Billing::Usage::HasTrackers
      end
```

However, we should do that in the billing usage gem after https://github.com/bullet-train-co/bullet_train-core/pull/335 ships.

PR #341, opened 2023-07-12
Run minitest suite for modified gems

gazayas:
The script is working for me locally, but we're getting a false positive on CircleCI, so I'll have to come back to this one to ensure the modified gems' tests are running and passing.

If we can manage that, then we can go ahead and delete the CircleCI config.yml files within each gem.


gazayas:
It's an eager loading issue, pretty sure the same as [this one](https://github.com/bullet-train-co/bullet_train-themes-tailwind_css/issues/2).

gazayas:
Here's a workaround for what the TangibleThingsController we added a while back, but I don't think we've been keeping track of each gem's unit tests, so I might have to dig into this and write workaround for each gem.

- https://github.com/bullet-train-co/bullet_train-super_scaffolding/pull/79/commits/08174beadb6c663ee2336a12ecca91eb167628c9

gazayas:
This one's a little difficult because we have a lot of cross-dependencies in `bullet_train-core`, so some of the dependencies inside the gems themselves point to the latest release as opposed to the other local gems in `bullet_train-core`. Trying to point the different gems to local paths, not having any luck so far though.

gazayas:
Trying to make all the cross dependencies local is proving to be a little more difficult said than done, but I'm okay with this script being something we can run locally before we merge any branches, so I'll go ahead and mark it as ready for review for now.

bin/run-gem-minitests.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Get names of modified files compared to the main branch.
+DIFF_DATA=$(git diff --name-status main)
```
kaspth, line 4: Should we convert this file to Ruby? I'm not the best at Bash, so this seems complicated to maintain.

I wonder if we could do something like this? (though we'd probably have to fiddle with regex)

```ruby
#!/usr/bin/env ruby

gems = `git diff --name-status main`.each_line.grep(/^bullet_train-/).uniq
gems.each do |gem|
  `cd #{gem} && bundle install && bin/rails test test/` # We're passing test/ to also run system tests.
end
```

gazayas: Done!

Unfortunately the eager loading issue is still a problem, so I think I'm just going to have to sit down with it and get as far as I can. There are a lot of moving parts so I'm not sure what needs to be included where, but if you have any insight on that side of things that would be super helpful! In the meantime, I'll try to look over it again and see what I can do.

gazayas:
Also just as a mental note the tests passing with the ruby script yield a false positive right now, so I'll have to work on that too

gazayas:
> Here's a workaround for what the TangibleThingsController we added a while back, but I don't think we've been keeping track of each gem's unit tests, so I might have to dig into this and write workaround for each gem.
> 
> * [bullet-train-co/bullet_train-super_scaffolding@08174be](https://github.com/bullet-train-co/bullet_train-super_scaffolding/commit/08174beadb6c663ee2336a12ecca91eb167628c9)

This fix worked when `bullet_train-super_scaffolding` was the only gem that had this issue, but now trying to implement this fix for other constants like `Account::ApplicationController` won't help since there's information in these controllers that we need to make tests pass. For example, `devise` isn't defined when we do this with the users controller.

```ruby
if defined?(Account::ApplicationController)
  class Account::UsersController < Account::ApplicationController
    include Account::Users::ControllerBase

    private

    def permitted_fields
      [
        # 🚅 super scaffolding will insert new fields above this line.
      ]
    end

    def permitted_arrays
      {
        # 🚅 super scaffolding will insert new arrays above this line.
      }
    end

    def process_params(strong_params)
      # 🚅 super scaffolding will insert processing for new fields above this line.
      strong_params
    end
  end
else
  class Account::UsersController
  end
end
```

I wonder if there's a workflow we can implement that doesn't require eager loading? I know it's not ideal, but it would be better than not being able to run tests at all.

gazayas:
Slowly coming to the mindset that Starter Repository tests are the most important thing we need, and although it would be ideal to enable these tests, there are just too many inter-dependencies to get it all working with eager loading.

gazayas:
@kaspth Thanks for confirming! I'll leave things for the time being then, if we really do need these tests to run sometime down the road though I'll be sure to check out https://github.com/ioquatix/bake.

PR #343, opened 2023-07-13
Minor refactoring in Transformer

gazayas:
When transforming strings, I felt that it would be easier to keep track of everything if we just singularlized the strings instead of keeping two separate lists.


jagthedrummer:
@gazayas, is this still WIP, or do you think it's ready to go?

gazayas:
@jagthedrummer Just checked tests locally, this one's good to go.

PR #346, opened 2023-07-14
Fix: Correctly interpret bool ENV vars truthiness

vfonic:
This issue is that Figaro treats all ENV vars in `config/application.yml`
as strings. When we set `HIDE_THINGS: false`, it gets interpreted as
`HIDE_THINGS = "false"` which makes it look truthy. But Rails has the
solution for testing truthiness.

I've also removed what looks like a generic `README.md` file.

Let me know if you'd like me to add the `README.md` back and or make any changes to the PR.


bullet_train/test/bullet_train_test.rb
```diff
@@ -4,4 +4,22 @@ class BulletTrainTest < ActiveSupport::TestCase
   test "it has a version number" do
     assert BulletTrain::VERSION
   end
+
+  test "When HIDE_THINGS is not truthy #scaffolding_things_disabled? returns false" do
+    old_hide_things = ENV["HIDE_THINGS"]
+    ENV["HIDE_THINGS"] = "false"
+
+    refute scaffolding_things_disabled?
```
vfonic, line 12: I made sure this fails before I wrote the fix.

bullet_train/lib/bullet_train.rb
```diff
@@ -104,12 +104,20 @@ def webhooks_enabled?
   true
 end
 
+def hide_things?
+  ActiveModel::Type::Boolean.new.cast(ENV["HIDE_THINGS"])
```
gazayas, line 108: If this is something specific to Figaro, I'm not sure we have to cast booleans here just to make it work. I think we should go along with whatever Figaro wants us to do (just write `0`?) and adjust our docs accordingly.

gazayas (commented):
Thanks for this! I left a comment about the `HIDE_THINGS` environment variable.

bullet_train/lib/bullet_train.rb
gazayas: Looks like we can just write `HIDE_THINGS:` (or just comment it out if needed).

vfonic: `0` is still converted to string `"0"` which is truthy.

```ruby
3.2.2 :001 > puts 'asd' if '0' 
(irb):1: warning: string literal in condition
asd                                          
 => nil                                      
3.2.2 :002 > puts 'asd' if '1'
(irb):2: warning: string literal in condition
asd                                          
 => nil                                      
3.2.2 :003 > !!"0"
 => true 
3.2.2 :004 > !!"1"
 => true 

3.2.2 :005 > ActiveModel::Type::Boolean.new.cast(0)
 => false 
3.2.2 :006 > ActiveModel::Type::Boolean.new.cast('0')
 => false 
3.2.2 :007 > ActiveModel::Type::Boolean.new.cast('1')
 => true 
```

That being said, I still believe calling `ActiveModel::Type::Boolean.new.cast` is the best solution.

vfonic: Here's the warning that I'm seeing when setting `HIDE_THINGS: false` in my `config/application.yml`:

https://github.com/laserlemon/figaro/issues/248

```
WARNING: Use strings for Figaro configuration. false was converted to "false".
```

vfonic: Here's the list of all `FALSE_VALUES` constant:
https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html

vfonic: > Looks like we can just write `HIDE_THINGS:` (or just comment it out if needed).

I mean...yes, we can. :) But you also don't want to rely on your users knowing that. I was bitten by this and so I thought it would be good to fix it in the codebase to prevent any possible such user error, instead of documenting it in documentation or just hoping that user (developer) knows about this. Logging output can be noisy and it's easy to miss the WARNING, which doesn't break the server...

vfonic: > If this is something specific to Figaro, I'm not sure we have to cast booleans here just to make it work. I think we should go along with whatever Figaro wants us to do (just write `0`?) and adjust our docs accordingly.

As you can see from your comment, you would've faced the same bug, because `"0"` is truthy. I strongly believe this should be fixed in the bullet_train-core (or in figaro gem itself, would be even better). Bullet Train added Figaro dependency. The way Figaro gem works doesn't have to be visible to the user (developer). End user just needs to know that he can put ENV vars in `config/application.yml`.

gazayas: > I strongly believe this should be fixed in the bullet_train-core (or in figaro gem itself, would be even better).

Thanks for the detailed comments! Yeah, like you said, I think it would be better to leave any changes concerning HOW we handle the environment variables to the original library, and I'm still of the opinion that we should let Figaro be Figaro and just adjust our docs accordingly. I'll sit on this one and comment again if anything changes/comes up!

vfonic: Thanks @gazayas!

I looked into this further and was about to open an issue on Figaro's repo asking why this is the case.
Turns out ENV vars in _Linux_ are always strings. This means that Figaro is doing good, following the same functionality of ENV vars, converting the ENV vars to strings. ENV vars work the same way in Heroku because Heroku ENV vars are Linux ENV vars.

That being said, if Bullet Train has ENV vars which are non-string (boolean or number) by type, then I believe Bullet Train should take care of converting those ENV vars to their proper types.
For example from Puma gem:
https://github.com/puma/puma/blob/db060253336f9d4eca08cd3abc306c1576345d5f/lib/puma/configuration.rb#L224-L231

Here's how Webpacker gem converts bool ENV var from string to bool:
https://github.com/rails/webpacker/blob/3fd96bcbf495db5a24a46606465e9837fec232c1/lib/tasks/webpacker/compile.rake#L37
https://github.com/rails/webpacker/blob/3fd96bcbf495db5a24a46606465e9837fec232c1/CHANGELOG.md?plain=1#L753-L754
They don't have Rails/ActiveModel dependency so they do the conversion from string to bool manually.
They also have a bug here:
https://github.com/forem/forem/blob/be1e57d85ccc624bbb201249895a4b87737457d5/lib/tasks/assets.rake#L23C13-L23C33
Because `WEBPACKER_PRECOMPILE=no` doesn't behave the same across the whole gem, but that's unrelated to this discussion. :)

Essentially, it _is_ Bullet Train's responsibility to properly interpret these ENV vars. Or would you say that
`HIDE_THINGS= false bin/dev`
or
`HIDE_THINGS=$BT_HIDE_THINGS_FROM_SHARED_TEAM_CONFIG_I_DUNNO_JUST_MAKING_SHIT_UP bin/dev`
is user's (developer's) responsibility? The "you're holding it wrong":
https://www.businessinsider.com/steve-jobs-iphone-4-reception-problems-stop-holding-it-the-wrong-way-2010-6
:)

vfonic: Hey @gazayas did we reach any conclusion here?

jagthedrummer (approved):
This strikes me as a good change that will make things more intuitive for developers. @gazayas, after sitting on it for a while do you have anything else you want to add? (I agree that in theory it would better if an upstream gem handled this, but since they don't I think it's worth taking on this small bit of code to reduce friction for BT devs/users.)

gazayas:
>  (I agree that in theory it would better if an upstream gem handled this, but since they don't I think it's worth taking on this small bit of code to reduce friction for BT devs/users.)

@jagthedrummer This is where I currently stand so I'm still of the opinion that it shouldn't be our responsibility, but I'm totally okay with having the code if it won't negatively impact our code base and if it will be easier for Bullet Train developers to navigate. Thank you for weighing in!

jagthedrummer (approved):

PR #349, opened 2023-07-15
Add missing notices partial for user details edit page

### LINKED ISSUE #345: `/account/onboarding/user_details/xxx/edit` doesn't show errors when first and/or last name are blank but doesn't allow user to proceed onboarding
I wanted to finish onboarding without providing my last name, but that didn't work. The page just reloads and there's no error shown.

I realized the issue happens because `@user.details_provided?` is `false`, which adds `flash[:error]`, but that error is not shown in the UI for some reason and I cannot figure out why.

Here's where the user gets redirected back to user_details/edit page:
https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train/app/controllers/concerns/account/onboarding/user_details/controller_base.rb#L27-L34

I might look into this, can anyone give me some guidance please?

gazayas:
Fixes #345.

スクリーンショット 2023-07-15 10 13 38

### DIFF
--- a/bullet_train/app/views/account/onboarding/user_details/edit.html.erb
+++ b/bullet_train/app/views/account/onboarding/user_details/edit.html.erb
@@ -6,6 +6,7 @@
     <% within_fields_namespace(:self) do %>
       <%= form_for @user, url: account_onboarding_user_detail_path(@user), method: :put, html: {class: 'form'} do |f| %>
         <%= render 'account/shared/forms/errors', form: f %>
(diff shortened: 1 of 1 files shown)


jagthedrummer:
This might be something that's better handled in a different PR, but it seems weird to me that only "Your Team Name" has the red asterisk indicating that it's required even though the notice box says that _all_ of the fields are required.

jagthedrummer (approved):

PR #350, opened 2023-07-16
Resolve: Allow passing END annotated view tag

vfonic:
I tried passing END annotated view tag to `bin/resolve --interactive` and it failed.
The fix to enable this is really small so I thought it could be well worth it.
Sometimes in HTML there could be thousands of lines inbetween the BEGIN and END tags.
So whichever you find first should work for `bin/resolve`.


bullet_train/lib/tasks/bullet_train_tasks.rake
```diff
@@ -40,11 +40,12 @@ namespace :bullet_train do
       puts "\nOK, paste what you've got for us and hit <Return>!\n".blue
 
       input = $stdin.gets.strip
-      $stdin.getc while $stdin.ready?
 
       # Extract absolute paths from annotated views.
       if input =~ /<!-- BEGIN (.*) -->/
         input = $1
+      elsif input =~ /<!-- END (.*) -->/
```
gazayas, line 47: Works for me 👍

bullet_train/test/lib/tasks/bullet_train_tasks_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  after do
+    $stdin = @original_stdin
+    $stdout = @original_stdout
+  end
+
+  describe "when passed BEGIN absolute path from annotated view" do
+    it "resolves the path" do
+      inputs = [
+        "<!-- BEGIN /ruby-3.2.2/gems/bullet_train-themes-light-1.2.26/app/views/themes/light/_notices.html.erb -->",
```
gazayas, line 21: I don't want to rely on specific versions of Ruby or our gems for these tests, so can we somehow justify these with a regular expression, or grab the ruby version and current gem version?

```ruby
# Something like this for the Ruby version
File.open(".ruby-version").readlines.pop

# Something like this for the gem version
`bundle show bullet_train`.regexp_to_get_version_only
```

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -26,9 +26,8 @@ def run(eject: false, open: false, force: false, interactive: false)
         puts ""
 
         if interactive && !eject
-          puts "\nWould you like to eject the file into the local project? (y/n)\n"
+          print "\nWould you like to eject the file into the local project? (y/n): "
```
gazayas, line 29: Unless there's a specific reason for using `print` over `puts` here, I think we should keep it as `puts` because that's what we use across the board for a lot of our other scripts (i.e. - [bin/configure](https://github.com/bullet-train-co/bullet_train/blob/main/bin/configure)).

If we want to change something stylistic like this I think it should have its own PR and rationale written there.

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -61,9 +60,8 @@ def run(eject: false, open: false, force: false, interactive: false)
         end
 
         if interactive && !open
-          puts "\nWould you like to open `#{source_file[:absolute_path]}`? (y/n)\n"
+          print "\nWould you like to open `#{source_file[:absolute_path]}`? (y/n): "
           input = $stdin.gets
-          $stdin.getc while $stdin.ready?
```
gazayas, line 66: Can I ask your rationale for removing this line? Not against the change, just curious as to why.

gazayas (changes requested):
Yes, we definitely should have tests for this script, thanks for writing this!

bullet_train/test/lib/tasks/bullet_train_tasks_test.rb
vfonic: I'm pretty sure I could strip that first part out. I actually did strip out some of the absolute path (my user account name). It's not a full absolute path. I'm not sure if that's a good idea. The _implementation_ ignores the good amount of the absolute path, but perhaps we should keep some kind of example absolute path nevertheless? 🤔 
For the sake of time, would it work if I removed ruby version as well since the absolute path doesn't have to be within the `ruby-<version>` directory. I'm using RVM so the gems are stored in `$HOME/.rvm/gems/ruby-<version>/gems`?

bullet_train/lib/bullet_train/resolver.rb
vfonic: Great question! The mocked `StringIO` doesn't have `ready?` method so it was failing.

From [this SO post](https://stackoverflow.com/questions/15690902/how-to-clean-stdin-in-ruby), `$stdin.getc while $stdin.ready?` can be used to make sure all characters passed to STDIN are being read.

So...I should probably add it back here where user is copy-pasting:
https://github.com/bullet-train-co/bullet_train-core/pull/350/files/0c1a30879f9ff9d82006b9af46ba10899f0900d3#diff-169057f5b52e3c8d1a1f9510a4ee51d00c3974ae4b89567cd707bede5b73322fL42

I'll see how I can mock `$stdin.ready?`.

vfonic:
@gazayas just replied to all your comments. Please have a look. I'll probably have some time on Wednesday or Thursday to make changes.

bullet_train/lib/bullet_train/resolver.rb
gazayas: I actually personally like the one-line setup you proposed:
```
> Would you like to eject the file into the local project? (y/n):  y
```

Also, totally valid about the `ask` method. We should be consistent with that too. Hopefully we can open up another PR or an issue addressing that soon!

bullet_train/lib/bullet_train/resolver.rb
gazayas: Great, I wasn't sure about this portion myself so I'm glad you found the post.

bullet_train/test/lib/tasks/bullet_train_tasks_test.rb
gazayas: > For the sake of time, would it work if I removed ruby version as well since the absolute path doesn't have to be within the ruby-<version> directory. I'm using RVM so the gems are stored in $HOME/.rvm/gems/ruby-<version>/gems?

Yes, I thought about this as well. Without prying too deep into it, I think we'll actually need to use an absolute path because we're running the test from inside `bullet_train-core/bullet_train` as opposed to a newly cloned repository itself where the task is usually run, but I'd have to check though.

If we go with an absolute path (which will most likely be more sustainable in the long run), it might take some work but you could potentially grab the path like this:
```
# Finds the installation directory for all the gems
> gem environment | grep "\- INSTALLATION DIRECTORY" | sed -E 's/- INSTALLATION DIRECTORY: //'

# You'd need to pull the version number and build the project path
# but this will at least give you the gem since `bundle show` won't work in bullet_train-core/bullet_train
> gem list | grep bullet_train-themes-light
#=> bullet_train-themes-light (1.2.26, 1.2.25, 1.2.24)
```

Then you could build the string like this:

```ruby
"<!-- BEGIN #{absolute_path} -->"
```

vfonic:
@gazayas I reverted those changes to puts/print. I also added back all of those while `.ready?` and used the actual absolute path from the documentation: https://bullettrain.co/docs/indirection

I think this might be an overkill:

```
# Finds the installation directory for all the gems
> gem environment | grep "\- INSTALLATION DIRECTORY" | sed -E 's/- INSTALLATION DIRECTORY: //'

# You'd need to pull the version number and build the project path
# but this will at least give you the gem since `bundle show` won't work in bullet_train-core/bullet_train
> gem list | grep bullet_train-themes-light
#=> bullet_train-themes-light (1.2.26, 1.2.25, 1.2.24)
```

...and it wasn't clear to me why would we need this and how exactly it's working or how it would work in the scope of Minitest. @gazayas feel free to take over if you feel like that's needed. I'll leave this PR as-is. I believe it's done and ready to be merged.

gazayas:
@vfonic Thanks again for all of these changes. The more I'm looking/thinking over it, the more I think these tests belong in the starter repository at [resolver_system_test.rb](https://github.com/bullet-train-co/bullet_train/blob/main/test/system/resolver_system_test.rb). I have an idea for how to solve it, so I'll try to make a PR and tag you over there.

jagthedrummer:
@vfonic @gazayas I just wanted to check in and see what you all are thinking about this one. Should we merge this, or are these tests better located in a different repo? I see https://github.com/bullet-train-co/bullet_train/pull/816 but it doesn't look like it's testing as much as this PR does.

vfonic:
I'm not 100% sure. I'm not that familiar with the project. But _I guess_ having some overlapping tests is fine.

gazayas:
@jagthedrummer We can go ahead and merge it, the other PR should be good to go as well :+1:

PR #352, opened 2023-07-18
Fix resolver issue for GitHub branches

gazayas:
@kaspth

The regular expression was working when resolving files in public releases of gems like so:
```
bullet_train-1.2.26
```

However, when running a matching branch between the starting repository and `bullet_train-core` in CircleCI, this string was showing up in the resolver:
```
bullet_train-core-b00a02bd513c/bullet_train/lib/bullet_train/resolver.rb
```

I went ahead and updated the regular expression so it handles both situations.


kaspth (approved):
@gazayas thanks!

PR #357, opened 2023-07-20
Docs: Usage: Include correct concern name; Fix typo

vfonic:
I don't know how to test this, I don't have access to Bullet Train Pro, but there seems to be discrepancy between mentioned module name
and the actual module name included in the code snippet.


vfonic:
@gazayas do you have a moment to have a quick look at this?

jagthedrummer (approved):
I think you're right @vfonic. Thanks!

PR #359, opened 2023-07-27
Rescue readonly-related error in `ExampleBot` module

### LINKED ISSUE #358: Can't nest models deeper than two levels
Running `bin/super-scaffold crud` outputs the following help text.

```
🚅  usage: bin/super-scaffold crud <Model> <ParentModel[s]> <attribute:type> <attribute:type> ...

E.g. a Team has many Sites with some attributes:
  rails g model Site team:references name:string url:text
  bin/super-scaffold crud Site Team name:text_field url:text_area

E.g. a Section belongs to a Page, which belongs to a Site, which belongs to a Team:
  rails g model Section page:references title:text body:text
  bin/super-scaffold crud Section Page,Site,Team title:text_area body:text_area

E.g. an Image belongs to either a Page or a Site:
  Doable! See https://bit.ly/2NvO8El for a step by step guide.

E.g. Pages belong to a Site and are sortable via drag-and-drop:
  rails g model Page site:references name:string path:text
  bin/super-scaffold crud Page Site,Team name:text_field path:text_area --sortable

🏆 Protip: Commit your other changes before running Super Scaffolding so it's easy to undo if you (or we) make any mistakes.
If you do that, you can reset to your last commit state by using `git checkout .` and `git clean -d -f` .

Give it a shot! Let us know if you have any trouble with it! ✌️
```

Super Scaffolding the `Section` model doesn't work here, and running tests after Super Scaffolding it produces an error.

## Commands for Example Models
```
rails g model Site team:references name:string url:text
bin/super-scaffold crud Site Team name:text_field url:text_area
rails g model Page site:references name:string
bin/super-scaffold crud Page Site,Team name:text_field
rails g model Section page:references title:string
bin/super-scaffold crud Section Page,Site,Team title:text_field
```

## Error
```
Error:
Api::OpenApiControllerTest#test_OpenAPI_document_is_valid:
ActionView::Template::Error: Cannot modify association 'Section#team' because it goes through more than one other association.
    app/views/api/v1/open_api/index.yaml.erb:60
    test/controllers/api/open_api_controller_test.rb:6:in `block in <class:OpenApiControllerTest>'
```

## Thoughts
I'm not quite sure how to fix this one right now, but I feel like this one is pretty high-priority if we're using an example in the help output.

gazayas:
Closes #358.

I tried to come up with some other solutions to fix this one because I didn't want to simply rescue the error, but I think the error is specific enough and the original code is clear enough to figure out what's going on here.

To be 100% honest I'm not entirely sure how much of an effect this will have on the Open API docs themselves, but I'm hoping to get the error rescue reviewed first, and I think I'll try to generate and compare some API docs in the meantime.

### DIFF
--- a/bullet_train-api/lib/bullet_train/api/example_bot.rb
+++ b/bullet_train-api/lib/bullet_train/api/example_bot.rb
@@ -72,7 +72,7 @@ module FactoryBot
           clone.send("#{name}=", associations)
           @tables_to_reset << name
         elsif %i[belongs_to has_one].include?(reflection.macro)
-          clone.send("#{name}=", instance.send(name).clone)
+          clone.send("#{name}=", instance.send(name).clone) rescue ActiveRecord::HasOneThroughNestedAssociationsAreReadonly
           @tables_to_reset << name.pluralize
         end
       end
--- a/bullet_train-api/lib/bullet_train/api/example_bot.rb
+++ b/bullet_train-api/lib/bullet_train/api/example_bot.rb
@@ -72,7 +72,11 @@ module FactoryBot
           clone.send("#{name}=", associations)
           @tables_to_reset << name
         elsif %i[belongs_to has_one].include?(reflection.macro)
-          clone.send("#{name}=", instance.send(name).clone) rescue ActiveRecord::HasOneThroughNestedAssociationsAreReadonly
+          begin
+            clone.send("#{name}=", instance.send(name).clone)
+          rescue
+            ActiveRecord::HasOneThroughNestedAssociationsAreReadonly
+          end
           @tables_to_reset << name.pluralize
         end
       end


gazayas:
For example as another solution, I figured we could get `reflection.source_reflection.active_record` and compare it with `reflection.options[:through]`. However, I found it will be difficult trying to constantize values for namespaced classes like `:absolutely_abstract_creative_concept`.

newstler:
I didn't go into the code yet, but is `instance.class.reflections.macro[:has_one]` really returns models that are 3 levels higher in the hierarchy? I doubt that should be, but if so, we probably should limit the levels of inheritance lookup some how.. I will try to dig in there deeper tomorrow.

gazayas:
Closing this one in favor of #387 since it provides more context.

PR #372, opened 2023-08-03
DO NOT MERGE: Just a test branch/PR for CI purposes

jagthedrummer:


jagthedrummer:
Confirmed that the system tests from the starter repo are run in CI when a PR is opened for `core`. I currently have no idea how https://github.com/bullet-train-co/bullet_train-core/pull/368 ever passed CI since it broke some tests in the starter repo.

PR #374, opened 2023-08-04
Refactor time zone selection

kaspth:
Removes unused `time_zone_name_to_id` and cuts needless `to_s` calls.


kaspth:
Actually, let's wait on this for now.

PR #377, opened 2023-08-04
Inject a theme color picker in Showcase

kaspth:
So it's a little easier to see them in practice, and they could be demoed on bullettrain.co/docs/showcase

The pickers work, the UI just isn't quite done yet. I'd be nice to have that drawer concept overlayed and maybe somehow using `position: sticky`. But my tailwind wrangling skills aren't quite there to make it happen, so if anybody else knows how!

Screenshot_2023-08-04_at_21 57 25


jagthedrummer:
@kaspth, I think what you showed in your screenshot is fine for now. What do you think?

If you think it's good to go can you look at the merge conflict?

kaspth:
@jagthedrummer I think the UI needs a little more fine tuning, but I can look at that.

jagthedrummer:
@kaspth, I just pushed a commit to this branch that makes the theme selector sticky.

CleanShot 2023-08-31 at 14 16 51

I see that there are some merge conflicts now. I'm not super familiar with all the showcase stuff but it looks like maybe we want all of both sides of the conflict?

kaspth:
Ok, I was able to get a version of this that works pretty much like I'd expect and I tuned the UI some too.

Screenshot 2023-09-01 at 18 39 08

kaspth:
@jagthedrummer I tuned the UI a bit more and I couldn't quite the JS you added working on my end. But I think this is good to go now, so I'll merge and then we can always tweak this if we find anything.

PR #378, opened 2023-08-04
Inject Light theme colors into Showcase

kaspth:
Let's the configured theme colors be used in Showcase automatically.

What we want before https://github.com/bullet-train-co/bullet_train-core/pull/377

Closes https://github.com/bullet-train-co/showcase/issues/68


kaspth:
There could potentially be clashes if apps also add a `showcase/engine/head` partial since that'll take precedent. I just don't have other ideas right now and it feels better that we have this.

PR #355, opened 2023-07-20
Introduce syntax highlighting

vfonic:
Couple of things to note:

1. This PR switches to commonmarker 1.0.0.pre9 which might not be stable enough for production use. I've checked all of the docs pages and it seems to work fine.
2. I had some Rust errors while upgrading the gem to 1.0.0.pre9, but it worked after I ran `rustup update`, which I found while searching the error message "error: failed to select a version for the requirement `clap = "^4.0"`" here:
   https://users.rust-lang.org/t/how-do-i-use-clap-4-0-0-rc-3/81865
3. I chose to use `InspiredGitHub` theme for syntax highlighting, but there are other themes. Here's the full list of available themes:
   https://github.com/gjtorikian/commonmarker#plugins
4. I didn't add ```sh to the shell code blocks, because the syntax coloring is tiny, but we can add it.

| Before | After |
| ----------- | ----------- |
| Screenshot 2023-07-19 at 22 17 02 | Screenshot 2023-07-19 at 22 16 57 |
| Screenshot 2023-07-19 at 23 08 58 | Screenshot 2023-07-19 at 23 08 19 |

Not really related, I've also added some dark mode styles to be able to read the docs easier at night:

```css
.prose {
  max-width: 76ch !important;
}
@media (prefers-color-scheme: dark) {
  .bg-white {
    background-color: #33415580;
  }
  .prose :not(:where([class~="not-prose"] *)) {
    color: #FFF;
  }
  .prose > p {
    color: #94a3b8;
  }
}
```

Not sure if anyone's interested in this? I could make a PR.

<details>
<summary>Here's how it looks</summary>

Screenshot 2023-07-19 at 23 15 25

</details>

PR #384, opened 2023-08-07
Introduce syntax highlighting

jagthedrummer:
I'm taking over https://github.com/bullet-train-co/bullet_train-core/pull/355 from @vfonic and since I can't push to his branch I'm starting a new branch with his work, plus a few modifications.


kaspth:
Come to think of it, it might be nice to move the `commonmarker` upgrade to just `bullet_train`'s Gemfile (instead of the `gemspec`), like in https://github.com/bullet-train-co/bullet_train-core/pull/355/files#diff-2e2559a7b01655e89a09351f8c04d2560b25cfc28b7918ff5473b26bea424f39R18

That way we don't force the `pre` upgrade onto users, but Bullet Train's public docs can still be updated.

We may need to add a `if Commonmarker::VERSION >= "1.x"` in the `markdown` helper though and scope the implementations.

jagthedrummer:
@kaspth I wondered about needing a different version but as far as I can tell the current version of that helper is using the `1.x` syntax. I'm a little surprised that it was working with versions < 1.

If you think there's a better way to handle the version bump I'm down for reviewing and merging a PR for it.

vfonic:
The helper was updated as you can see in this PR, if that's what you're referring to. So, before this PR, the code used commonmarker v <1 syntax and v0.x gem. After this PR the code uses v1. X syntax and v1.x (pre) gem version.

jagthedrummer:
Oh dang, you're right! I somehow missed that, or was looking in the wrong place. Yeah, we may need that conditional that you mentioned @kaspth. Or we leave the dependency update in the `.gemspec` and force it.

kaspth:
I'll make a PR tomorrow to move the version lock to the Gemfile, so we don't force the `pre`-releases on users. They may be using the `markdown` helper in production to render non-documentation content, and if we've got the version lock in the gemspec and anything goes wrong they can't downgrade. — I mean they can downgrade `bullet_train-core` but that may have others issues with it too.

PR #386, opened 2023-08-08
Fix html option settings for objects that have multiple values

gazayas:
I thought we would be ok with #317, but it turns out that merging it broke the tests.

What's happening here is that super select objects uses `html_options: {multiple: true}` whereas other objects use `multiple: true`:

```erb
<!-- From PartialTest's _form.html.erb partial after running the Super Scaffolding test setup script-->

<%= render 'shared/fields/buttons', method: :multiple_buttons_test, multiple: true %>

<%= render 'shared/fields/super_select', method: :multiple_super_select_test, html_options: {multiple: true} %>
```

It's a little difficult to tell what's going on in this portion of the transformer, but I hope this comment helps to clear things up a little bit.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb
@@ -841,8 +841,8 @@ class Scaffolding::Transformer
           field_options[:color_picker_options] = "t('#{child.pluralize.underscore}.fields.#{name}.options')"
         end
 
-        # When rendering a super_select element we simply use `multiple: true` as an option,
-        # but all other fields require `html_options {multiple: true}` to work.
+        # When rendering a super_select element we need to use `html_options: {multiple: true}`,
+        # but all other fields simply use `multiple: true` to work.
         if is_multiple
           if type == "super_select"
             field_options[:multiple] = "true"
--- a/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb
@@ -841,13 +841,21 @@ class Scaffolding::Transformer
           field_options[:color_picker_options] = "t('#{child.pluralize.underscore}.fields.#{name}.options')"
         end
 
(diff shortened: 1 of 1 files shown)


jagthedrummer:
@gazayas Which tests did this break? CI was green on the `main` branch of `core` after I merged this.

gazayas:
@jagthedrummer Yes, I think we were getting a false positive in CI, and the Super Scaffolding partial tests were failing. I'm assuming the tests were running against released versions as opposed to the most recent changes on GitHub. I say that because I ran `bin/hack --link` in the starter repository and THEN ran the Super Scaffolding Partial tests which failed.

The other one failing right now is related to `.daterangepicker`. I'm hoping we can merge https://github.com/bullet-train-co/bullet_train/pull/828 and #182, because I believe it will solve that issue as well. I saw the comment you left there about handling formats properly, so I'll try to work on that now.

jagthedrummer:
@gazayas Ah-ha! It all makes sense now. It's the what I found in https://github.com/bullet-train-co/bullet_train/pull/831. We were looking at the same thing, just coming at it from different directions.

So, just kinda thinking out loud here. Does it seem reasonable to do the following?

1. Merge this PR so that tests go green in #831.
2. Prepare a "revert" for this PR so we can get these changes back when we need them.
2. Merge #831.
3. Deal with https://github.com/bullet-train-co/bullet_train/pull/828 & https://github.com/bullet-train-co/bullet_train-core/pull/182 (being able to see their test status in the starter repo against both the released gems and `core/main`).
4. Merge the revert for this PR.

Also open to suggestions!

jagthedrummer (approved):

jagthedrummer:
Welp, merging this didn't fix the situation in https://github.com/bullet-train-co/bullet_train/pull/831. Now I'm suspecting that https://github.com/bullet-train-co/bullet_train-core/pull/334 has something to do with it. Unfortunately too much stuff has changed since we merged that and we can't just generate a revert PR for it. I'm going to try making a branch and undoing stuff in that PR by hand.

PR #387, opened 2023-08-09
rescue ExampleBot deep cloning error

jraedisch:
Calling e.g. address.team= throws an error, if address has_one team through person and person has_one team through company and company belongs_to team so the resulting error will be caught here and a warning logged.


jraedisch:
Linking to #358.

jagthedrummer (approved):
Thank you!

PR #389, opened 2023-08-09
Revert #334 - Themes localization

jagthedrummer:
This reverts #334 as it is causing test failures in the Minitest For Super Scaffolding tests in the starter repo when run against the `main` branch of `core`.

And they seem to be legitimate failures. With my starter repo linked to all of my local copies of everything in `core` I can reproduce the test failures. And when I enable Magic Test to pause the test just before the failure I can see that the date and date/time pickers are no longer working.

After preparing this PR my local tests are passing again, and the date and date/time pickers are working again. It also gets tests passing again in CI. https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/1981/workflows/4f3dcbe3-8ff4-4f55-8b13-9eea7f560494

@3pns I think the goals you outlined in #334 are great and I think we should adopt them as goals, but it seems the implementation needs a little more work. If you'd like to pick it back up that would be great. If you don't have time I'll put it on my (already large) list.

### DIFF
--- a/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_date_and_time_field.html.erb
+++ b/bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_date_and_time_field.html.erb
@@ -1,45 +1,43 @@
 <%
-stimulus_controller = 'fields--date'
-
 form ||= current_fields_form
 options ||= {}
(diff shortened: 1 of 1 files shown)


jagthedrummer:
Here's a PR to revert this revert: https://github.com/bullet-train-co/bullet_train-core/pull/390

PR #392, opened 2023-08-09
We should run the super_scaffolding_partial_test as well

### LINKED ISSUE #391: Starter repo tests aren't linking everyting
We currently link only the gems explicitly mentioned in the `Gemfile` of the starter repo, and anything that's a dependency that only shows up in `Gemfile.lock` ends up using the version that's in `Gemfile.lock`. This means that it's possible to introduce breaking changes in `core` that are missed by the test suite because the test suite isn't running against the latest state of `core`.

For instance: https://github.com/bullet-train-co/bullet_train-core/pull/390#issuecomment-1671932020

Here's a failing test run in the starter repo CI when linked against the branch from #390: https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/1982/workflows/fd4ed8da-6526-463c-ab5b-c15dc0b3d522

But the tests on #390 that were run in _this_ repo passed: https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train-core/1113/workflows/6d85dc23-51d7-45c8-8e48-018ec1e7b665

jagthedrummer:
Previously we could introduce changes in `core` that would break tests in the starter repo, but that wouldn't show up as failures in `core`.

This was due to the fact that we were running more tests in the starter repo than we were runnig here.

This PR makes it so that we run the extra tests so that we can spot test failures here.

Fixes #391

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,6 +139,9 @@ jobs:
       - run:
           name: 'Run Super Scaffolding Test'
           command: "bundle exec rails test test/system/super_scaffolding_test.rb"
+      - run:
+          name: 'Run Super Scaffolding Partial Test'
+          command: 'bundle exec rails test test/system/super_scaffolding_partial_test.rb'
 
 workflows:
   version: 2


.circleci/config.yml
```diff
@@ -139,6 +139,9 @@ jobs:
       - run:
           name: 'Run Super Scaffolding Test'
           command: "bundle exec rails test test/system/super_scaffolding_test.rb"
+      - run:
+          name: 'Run Super Scaffolding Partial Test'
+          command: 'bundle exec rails test test/system/super_scaffolding_partial_test.rb'
```
kaspth, line 144: I wonder if we should refactor these tests into a `test/system/super_scaffolding` directory on the starter repo end? So we could just get `bundle exec rails test test/system/super_scaffolding/` here.

Could do `test/system/super_scaffolding/super_scaffolding_test.rb` and `test/system/super_scaffolding/partial_test.rb` for the test names.

kaspth (commented):
Sweet!

.circleci/config.yml
jagthedrummer: @kaspth That sounds reasonable.

kaspth: @jagthedrummer do you want to look at doing it? I'm kinda up to my ears in open PRs right now.

jagthedrummer: @kaspth I'll open an issue for it and we can do it later. I'm also swamped with open stuff (which I'm trying to get closed). What we have currently is good enough for now.

jagthedrummer: https://github.com/bullet-train-co/bullet_train/issues/833

PR #393, opened 2023-08-09
Themes localization

3pns:
This PR was recovered from #334 after it was reverted due to new tests added to the repo CI failing due to this PR.

Original description:

The goal here is :

- allow using different Date and DateTime display formats in bullettrain fields.
- simplify Date and DateTime handling in the backend by using ISO8601 instead of a localized date/datetime format as rails and all modern languages and frameworks support iso8601 natively.

This is what I did :

- changed all themes date and datetime related form fields data format from an American localized Date and Datetime Strings to ISO8601, added a hidden display field that is connected to the datetimepicker
- removed existing assignments and scaffolding of assign_date_and_time and assign_date_and_time assignments as Ruby On Rails handle iso8601 parsing natively.
- add datetimepicker translations and fields formats
kept assign_date_and_time and assign_date for backward compatibility.


jagthedrummer:
@3pns, you might hold off on doing anything to this for a little while. @pascallaliberte pointed out to me earlier that our CI process currently depends on the JS packages that are published on npm and that it _does not_ run against the version of the JS that's in whatever branch of `core` is being tested. I'm currently working on resolving that issue, and given the amount of JS in this PR I wouldn't be surprised if that's at least a contributing factor.

Also worth mentioning, when I had recreated the failures previously I _did not_ have my local starter repo linked to my local `core` repo, so I was also testing against the released versions of the npm packages, which obviously wouldn't have any of your JS modifications.

I apologize for all the confusion around this!

3pns:
Well noted, thanks for the heads up @jagthedrummer !

jagthedrummer:
OK @3pns, I merged #396. And then out of curiosity I rebased my copy of this branch against `main`, pushed it, and opened a temporary PR (#397) to see what it did for CI. The scaffolding tests are still failing in the same way.

I spent a little time looking at it, and I think I have an idea of what's causing the trouble.

To trigger the date/datetime pickers the test is looking for an element with an `id` of `partial_test_date_test`, but this PR makes it so that the visible element now has an `id` of `partial_test_date_test_display`.

And, if I'm understanding everything correctly (which I may not be!), I think this is a breaking change that will cause problems for anyone who has ejected the `_date_and_time_field.html.erb` or `_date_field.html.erb` partials.

I'm kind of new to the management/admin side of BT so I'm not entirely sure what the project's stance is towards this kind of thing. I'll see what I can learn about that.

3pns:
The test suite now passes thanks to this other PR https://github.com/bullet-train-co/bullet_train/pull/845

I added back the strong_params variable to deliveries_controller.rb since it breaks super scaffolding for the controller as pointed out by @kaspth in the previous PR.

Regarding the impact on existing users 
- those who have already ejected their fields, it should still work for them as the related helpers in date_and_time_support.rb and date_support.rb are still present but deprecated.
- those who have not ejected their fields will have the deprecation message explaining to them what to do to upgrade, it should also work for them as the helpers now support both the legacy format and iso8601.

jagthedrummer:
OK, gonna merge this one and then immediately release version `1.3.2` of the `core` gems, then will merge the joint PR in the starter repo, and update the starter repo to `1.3.2`.

gazayas:
:rocket:

PR #396, opened 2023-08-10
Link npm packages in CI

jagthedrummer:
This makes it so that we're running tests against the JS provided by `core` instead of against the published versions of the packages on npm.


pascallaliberte (approved):

kaspth (commented):
🙌

PR #398, opened 2023-08-11
Contain commonmarker upgrade to just Bullet Train for now

kaspth:
As discussed in #384 within https://github.com/bullet-train-co/bullet_train-core/pull/384#issuecomment-1668523752

Bullet Train apps can still do `bundle update --pre commonmarker` to get the new version going but we don't force it on them now.

### DIFF
--- a/bullet_train/Gemfile
+++ b/bullet_train/Gemfile
@@ -11,5 +11,7 @@ gem "sprockets-rails"
 # TODO Have to specify this dependency here until our changes are in the original package.
 gem "active_hash", github: "bullet-train-co/active_hash"
 
+gem "commonmarker", ">= 1.0.0.pre10"
+
 # Start debugger with binding.b [https://github.com/ruby/debug]
 # gem "debug", ">= 1.0.0"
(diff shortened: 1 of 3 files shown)


kaspth:
@jagthedrummer I'm going to merge this, since I can't think of another way that we'd need to scope the `if` in the `markdown` helper.

jagthedrummer:
Looks great, @kaspth! Do you have any thoughts on how/when we should manage the forced upgrade? I'm guessing we at least want to wait until the `1.0` line is out of `pre`?

kaspth:
@jagthedrummer yeah, I'd probably wait until after 1.0 drops. Maybe even 1.0.1 or later.

I just watched for new releases on the repo, so I'll be notified and I can follow up on our end as needed.

kaspth:
> I'm guessing we at least want to wait until the `1.0` line is out of `pre`?

Another thing to point out with this is that bundler won't automatically pull `pre` versions during `bundle update`, that's why I'm adding `--pre` in the command in the PR description.

So that feels like an extra indication to wait until they're out of `pre`.

PR #400, opened 2023-08-11
Jeremy/fix breaking date change

jagthedrummer:
WIP to see if we can maintain some backwards compatibility.

Joint PR https://github.com/bullet-train-co/bullet_train/pull/844

### DIFF
--- a/bullet_train/app/helpers/account/dates_helper.rb
+++ b/bullet_train/app/helpers/account/dates_helper.rb
@@ -1,21 +1,51 @@
 module Account::DatesHelper
   def display_date(timestamp, custom_date_format = nil, format: :default, date_format: nil)
-    date_format ||= custom_date_format
     format = date_format if date_format
-    localize(local_time(timestamp).to_date, format: format) if timestamp
+
+    if format && format != :default
+      localize(local_time(timestamp).to_date, format: format) if timestamp
+    else
+      # e.g. October 11, 2018
+      if custom_date_format
+        local_time(timestamp).strftime(custom_date_format)
+      elsif local_time(timestamp).year == local_time(Time.now).year
+        local_time(timestamp).strftime("%B %-d")
+      else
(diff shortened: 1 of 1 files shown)


bullet_train/app/helpers/account/dates_helper.rb
```diff
@@ -1,17 +1,54 @@
 module Account::DatesHelper
-  def display_date(timestamp, format: :default, date_format: nil)
+  def display_date(timestamp, custom_date_format = nil, format: :default, date_format: nil)
```
jagthedrummer, line 2: I added back `custom_date_format` to preserve the call signature to that we don't cause errors if someone is calling this method directly with the old syntax.

bullet_train/app/helpers/account/dates_helper.rb
```diff
@@ -1,17 +1,54 @@
...
+    return nil unless timestamp
     format = date_format if date_format
-    localize(local_time(timestamp).to_date, format: format) if timestamp
+
+    if format && format == :default
```
jagthedrummer, line 6: And this is an attempt to preserve the old behavior if someone doesn't explicitly include a `format` or `date_format`. I suspect that this might interfere with the idea that people could use localization strings to customize things, but I'm not entirely sure how all of those pieces fit together.

jagthedrummer:
@gazayas I'd love to hear your thoughts on this one. It's basically a partial revert of #182 with the goal of preserving/restoring the `Yesterday at ...` default formatting that we used to have.

gazayas:
@jagthedrummer Works for me! I was able to confirm that the different formats work locally. I'm not currently sure how we'll compensate for #393 right now, but I think this is fine for handling old and new format customization.

gazayas (approved):

bullet_train/app/helpers/account/dates_helper.rb
kaspth: Good call!

PR #461, opened 2023-08-23
Add configuration for incoming webhook parent class

### LINKED ISSUE #454: Add a `parent_class` configuration option for `BulletTrain::IncomingWebhooks`.
In a downstream application, the default super class of any new incoming webhook model is `WorkspaceRecord` not `ApplicationRecord`. It would be great if this were configurable at a package level instead of having to maintain a local version of `app/models/webhooks/incoming/bullet_train_webhook.rb` (that becomes outdated when changes happen upstream.)

gazayas:
Closes #454.

Joint PR (Details written there)
- https://github.com/bullet-train-co/bullet_train/pull/899

### DIFF
--- a/bullet_train/lib/bullet_train/configuration.rb
+++ b/bullet_train/lib/bullet_train/configuration.rb
@@ -1,7 +1,7 @@
 module BulletTrain
   class Configuration
     include Singleton
-    attr_accessor :strong_passwords
+    attr_accessor :strong_passwords, :incoming_webhooks_parent_class_name
 
     @@config = nil
 
@@ -15,6 +15,10 @@ module BulletTrain
     class << self
       def strong_passwords
         @@config&.strong_passwords
+      end
+
+      def incoming_webhooks_parent_class_name
+        @@config&.incoming_webhooks_parent_class_name
       end
     end
   end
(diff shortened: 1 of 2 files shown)


bullet_train/lib/bullet_train/configuration.rb
```diff
@@ -16,6 +16,10 @@ class << self
       def strong_passwords
         @@config&.strong_passwords
       end
+
+      def incoming_webhooks_parent_class_name
+        @@config&.incoming_webhooks_parent_class_name
```
jagthedrummer, line 21:
```suggestion
        @@config&.incoming_webhooks_parent_class_name || "ApplicationRecord"
```

It might be nice to have a safe default here just in case someone doesn't have the magic line in their initializer. Or maybe we want to set a default in `initialize` instead? (Not sure why we set `@strong_passwords` there. We don't seem to use it.) What do you think, @gazayas?

gazayas: @jagthedrummer Good call, I went ahead and set it as a default value in `initialize`. The original convention of setting `strong_passwords` here was introduced in #23 and I just went along with it (although it wasn't working at the time so I submitted #208).

PR #478, opened 2023-08-25
Pass a custom path to bin/hack --link

### LINKED ISSUE #463: `--link-local` flag for `bin/hack`
If I already have the `core` repo on my machine it would be nice to be able to pass a path to it to `bin/hack` so that I can use that copy, instead of checking out a fresh copy.

Something like:
```
bin/hack --link-local ../bullet_train-core
```

gazayas:
Fixes #463.


jagthedrummer (approved):
Nice! Works like a charm.

PR #486, opened 2023-08-27
Caching _path_methods in automatic_paths_for

newstler:


jagthedrummer:
@newstler, I was looking at this and just kind of wondered what would happen if we used `FactoryBot.build` to create shell objects instead of using `FactoryBot.create` + clone + rollback.

It turned out that it seems to work. At least as far as anything that we have test coverage for.

I don't know much about the `-api` gem or about `ExampleBot`, so there may be reasons that the `.build` idea won't work. But I'd love to hear what you think about using the following two PRs instead of this one and it's companion.

https://github.com/bullet-train-co/bullet_train-core/pull/524
https://github.com/bullet-train-co/bullet_train/pull/965

jagthedrummer:
Or maybe we'd want to do both? Use `.build` instead of `.create` _and_ cache the various pieces as we're assembling the document? (How did you generate those `METHOD`/`TASK` benchmarks?)

newstler:
@jagthedrummer We used `.create`because we needed `id`s populated and callbacks to fire. We can of course fake ids, but it appeared we cannot fake callbacks when using `.build`, so that's why we ended up using `.create` with populating and then resetting the DB.

From your comment I think we probably need some more tests to be able to see this difference between `.create` and `.build`, will work on this.

I just used `Benchmark.bmbm` around those methods and around the whole task that writes openapi.yaml to a file.

jagthedrummer:
@newstler, thanks for the context.

I hadn't thought about callbacks at all. Is the reason that we want callbacks to fire _only_ so that we auto-create any related records that would normally be there, or are there instances where we'd want real side effects to happen (like queuing a Sidekiq job or sending an email or something)?

Sounds like maybe we should do the following:
1. Merge these caching PRs
2. Beef up our tests to make sure that we test for everything that we might miss if we switched to `.build`
3. Revisit the `.create -> .build` PRs that I linked above

Does that sound reasonable to you?

newstler:
I don't really remember what were the callbacks there, but as I know BT does a lot of tricky stuff especially with teams and users.

Yep, I think your proposal is the best we can do right now (because caching is being awaited for CF asap). I just created an issue https://github.com/bullet-train-co/bullet_train-core/issues/526 to write the tests and assigned it to myself.

jagthedrummer (approved):

PR #490, opened 2023-08-29
Features/gem tests

gazayas:
This is an attempt to run all of the tests for the gems in this repository.

Running `CI=true bundle exec rails test local/bullet_train-core/bullet_train` will work, but if we run `CI=true bundle exec rails test` from within `bullet_train-core/bullet_train`, it will fail because it doesn't have the context of the full application.

I figured we could test these gems from the context of a starter repository application because we already have the `Starter Repo Minitest` in place.

We would have to do a lot of editing to the tests to make them pass, but as of now this is the only solution I can come up with to automate our core gem tests on CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
     executor: ruby/default
     steps:
       - checkout:
-        path: ~/project-gems
+          path: ~/project-gems
       - browser-tools/install-browser-tools:
           firefox-version: "112.0"
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,16 +28,27 @@ jobs:
     executor: ruby/default
     steps:
       - checkout
+        path: ~/project-gems
       - browser-tools/install-browser-tools:
           firefox-version: "112.0"
 
+      # TODO: This is a workaround to get `git clone` working, but it shouldn't be here.
+      # https://github.com/CircleCI-Public/browser-tools-orb/issues/62
+      - run:
+          command: rm LICENSE.chromedriver
+
+      - run: ../../bin/checkout-and-link-starter-repo
+
       - ruby/install-deps:
           clean-bundle: true
-      - run: "yarn install"
+      - node/install-packages:
+          pkg-manager: yarn
+      - run: yarn build:css
+      - run: yarn build
       - *wait_for_docker
       - run:
           name: Run unit tests
-          command: bundle exec rails test
+          command: ../../bin/test-core-gems
 
   'Local Standard Ruby':
     docker:
@@ -149,7 +160,7 @@ workflows:
   version: 2
   build:
     jobs:
-      # - 'Local Minitest'
       - 'Local Standard Ruby'
+      - 'Local Minitest'
       - 'Starter Repo Minitest'
       - 'Starter Repo Minitest for Super Scaffolding'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
       - image: circleci/redis
     executor: ruby/default
     steps:
-      - checkout
+      - checkout:
         path: ~/project-gems
       - browser-tools/install-browser-tools:
           firefox-version: "112.0"
--- a/bin/test-core-gems
+++ b/bin/test-core-gems
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+packages=(
+  "bullet_train"
+  "bullet_train-api"
+  "bullet_train-fields"
+  "bullet_train-has_uuid"
+  "bullet_train-incoming_webhooks"
+  "bullet_train-integrations"
+  "bullet_train-integrations-stripe"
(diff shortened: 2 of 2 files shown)


jagthedrummer:
🤘 Oh, excellent! This looks like a solid start. I wasn't quite sure what sort of wiring we'd need between the gems and the starter repo, but it sounds like you've figure that out.

Since I've been knee deep in CI stuff lately I don't mind picking this up if you'd like. No pressure!

gazayas:
> Since I've been knee deep in CI stuff lately I don't mind picking this up if you'd like. No pressure!

@jagthedrummer I'm down either way! Feel free to commit straight to this PR if anything :smiley: `../../bin/checkout-and-link-starter-repo` isn't quite being run at the moment, but I'd be really happy if these start to pass on CI.

I know there are some other PRs to prioritize, so I'll swing back to this after things settle down!

gazayas:
> We would have to do a lot of editing to the tests to make them pass, but as of now this is the only solution I can come up with to automate our core gem tests on CI.

I say this primarily because, in the tests, there are places where we try to edit files.

https://github.com/bullet-train-co/bullet_train-core/blob/0755711ac7bae4cd05fff417df5445086b78fa15/bullet_train-super_scaffolding/test/lib/scaffolding/routes_file_manipulator_test.rb#L157-L161



```ruby
```

Since we're now in the root directory now though, it tries to read from the wrong place and we get an error like this:
```
Error:
Scaffolding::RoutesFileManipulator::apply#test_0004_adds 'Teams::Log' under 'Team' appropriately:
Errno::ENOENT: No such file or directory @ rb_sysopen - test/lib/scaffolding/examples/example._rb
    local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb:11:in `readlines'
    local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb:11:in `initialize'
    local/bullet_train-core/bullet_train-super_scaffolding/test/lib/scaffolding/routes_file_manipulator_test.rb:158:in `new'
    local/bullet_train-core/bullet_train-super_scaffolding/test/lib/scaffolding/routes_file_manipulator_test.rb:158:in `block (3 levels) in <top (required)>'


rails test local/bullet_train-core/bullet_train-super_scaffolding/test/lib/scaffolding/routes_file_manipulator_test.rb:157
```

There were some errors related to factories as well, but this one in particular stood out to me.

gazayas:
Managed to get the `bullet_train` tests to run under the `Starter Repo Minitest` job, but I ultimately want this to be in the `Local Minitest` job. Having some issues with the working directory, will keep trying to find a fix.

image

gazayas:
Ah, it actually looks like it's only running the starter repository minitest. Maybe the `bullet_train-core` gems have to be INSIDE the starter repository for it to pick up on the tests (like in `local/bullet_train-core`)?

gazayas:
Ok, and the tests are running!!

We're getting some false positives (See the tests on node 7 for example) so I'll have to find a way to make the script exit with the proper status.

I also don't plan on keeping the name `bin/checkout-and-link-starter-repo-2`. I personally like having the setup be under `local/bullet_train-core` because it reflects what we do in `bin/hack --link`, so it might be good to edit our original `bin/checkout-and-link-starter-repo` script to just do that.

kaspth (approved):
Nice! Does this mean that apps will be running these tests too, though?

gazayas:
@kaspth I don't *think* so, as long as the CI test is in `bullet_train-core` and not the starter repository I think we should be okay. Is that what you meant?

gazayas:
Almost done with this one, just want to make sure the tests are running properly across the nodes. I'll also change the bin/checkout-and-link... script name before marking it as ready for review.

kaspth:
> I don't think so, as long as the CI test is in bullet_train-core and not the starter repository I think we should be okay. Is that what you meant?

Ah ok, I thought this would be running in the starter repo CI script. All good!

> I am concerned however about running the tests on 16 nodes. Do you think we're eating up too many resources? I'm not too sure about how much we have to spend on CI nodes, etc. per month, but I thought it would be better to run the tests in parallel since we have a decent amount of packages.

There's barely any tests in these packages, and we spend most of our time on CI setup. I think we could get around the same speed right now with just 1 node — we can always scale up later too.

kaspth (approved):
By the way, if this runs tests for `bullet_train-roles` we can remove the GitHub Actions based CI I setup for that too.

gazayas:
@kaspth Oddly enough the `bullet_train-roles` tests fail when running them from the starter repository :thinking: Mainly because it's throwing an error about a missing factory that exists in `bullet_train-roles` as opposed to the working directory.

I'm under the impression that we should run our tests within the context of the starter repository, so I'm not 100% sure how to handle the roles tests right now (along with the other ones I commented out). Will definitely get rid of the GitHub Actions CI setup if we can figure this one out though.

I also went ahead and made the tests run linearly.

jagthedrummer (approved):

PR #496, opened 2023-08-30
Use parent_association setting for outgoing webhook events

abradburne:
Use parent_association setting for API output rather than hardcoded team in outgoing webhook events


andrewculver (approved):

andrewculver:
This makes sense!

Issue #379, opened 2023-08-04
Required fields do not always display a red asterisk


gazayas:
In the screenshot in #349, you can see that all of the fields in the form are required, but not all of them have a red asterisk indicating that they are.

The original PR implementing the style change: #107

jagthedrummer:
I just looked into this a bit and I think this line is what causes this to be unpredictable:

https://github.com/bullet-train-co/bullet_train-core/blob/2b9dea3798d625943d12c1a959d817cac71122df/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb#L34

I think that's a good fallback, but we probably want to somehow allow the required status to be passed in.

For instance, on the user registration form the `password_confirmation` _is_ a required field, but there isn't a true validation on the `User` model for it. So it would be nice if we could pass in an option in this block that will tell the `field` partial that `password_confirmation` is required even though we don't have a validation for it.

https://github.com/bullet-train-co/bullet_train-core/blob/2b9dea3798d625943d12c1a959d817cac71122df/bullet_train/app/views/devise/registrations/new.html.erb#L16-L21

Something like:

```diff
  <%= render 'shared/fields/password_field', form: f,
    method: :password_confirmation,
+   required: true,
    other_options: {
        error: f.object.errors.full_messages_for(:password).first,
        hide_custom_error: true
    }
  %>
```

I'm not sure if it should be a top-level option, or inside the `other_options` hash. Just trying to illustrate the idea.

PR #499, opened 2023-08-30
Allow `required` to be passed in to field partials

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train-core/issues/379 (Kind of)

Currently we have kind of weird handling around required fields. This PR is an attempt to resolve that situation.

It's a pretty small change, but has potentially big impacts on applications. I'd love to get a bunch of eyes on this and hear what everyone thinks, so I'm going to tag a bunch of people. (Sorry! Feel free to tag in anyone else that might be interested.) @andrewculver @gazayas @kaspth @pascallaliberte @newstler @adampal @julianrubisch 

## The Problem

Currently we show the red asterisk on required fields **only** if the object represented by the form has a validation for `presence` on that field. For instance on the user registration form we don't show an asterisk for `password_confirmation` because Devise doesn't setup a `validates_presence` validation for that field.

CleanShot 2023-08-30 at 10 45 51

But the field really is required. If you fill in a password, but no confirmation and then submit the form, it causes an error.

CleanShot 2023-08-30 at 10 52 33

If a developer were to eject the devise form they could add `options: { required: true}` to the form:

```diff
  <%= render 'shared/fields/password_field', form: f,
    method: :password_confirmation,
+   options: { required: true },
    other_options: {
        error: f.object.errors.full_messages_for(:password).first,
        hide_custom_error: true
    }
  %>
```

That will have the effect of setting the `required` attribute on the password field that ends up getting rendered. That leads to a confusing experience where the label does not indicate that the field is required, but the browser does complain that it's not filled in. 

CleanShot 2023-08-30 at 10 42 29

(And if they didn't also add it to the main `password` field they would not see the browser complain about _that_ field being blank.)

## One possible fix

This PR makes it so that if `required` is passed into the `options` hash that the field partial that renders the asterisk will use that value first, and will fallback to the existing validation method if it's not passed in.

That makes it so that the red asterisk will show up, and then the browser "validation" matches the label.

CleanShot 2023-08-30 at 10 58 39

This PR _also_ makes it so that we pass that same `required` value in the `options` hash down into the rendered HTML element. 

## Possible problems

The biggest problem I see with this is that it dramatically changes what happens when someone tries to submit a form without filling in a required attribute.

Currently the form would actually be submitted and then we are in control of what the user sees in response.

CleanShot 2023-08-30 at 11 01 09

This PR makes it so that most browsers would refuse to even submit the form, and would show their own native messaging around empty required fields.

CleanShot 2023-08-30 at 11 00 32

## Other options

We could potentially tell people to put `required: true` in `other_options` instead of `options` so that it's not passed down to the rendered html element. ([See the docs for more info about this.](https://bullettrain.co/docs/field-partials#options-vs-other_options))

And then we'd modify this PR to look at `other_options` instead of `options`.

And we could potentially remove `required` from the `options` hash if it's passed there, just to be consistent. But I think that leads to...

## Accessibility concerns

I think that this PR is the "most correct" way of handling this because it ends up getting the `required` attribute into the rendered HTML of all fields that are actually required. That means that browsers and screen readers and other accessibility tools can "see" it. I _think_ that this is the best option from an accessibility standpoint because it relies on the HTML standard, whereas our current solution relies on people being able to _visually see_ the little red asterisk that we add via CSS.

But I realize that this is a big change from a user experience perspective, so I want to hear what everyone else thinks.

### DIFF
--- a/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
+++ b/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
@@ -15,6 +15,11 @@ options ||= {}
 options[:id] ||= form.field_id(method)
 # options[:disabled] ||= !field_editable?(form.object, method) if user_signed_in?
 options[:placeholder] ||= labels.placeholder if labels.placeholder
+
+if !options.key?(:required)
+  options[:required] = presence_validated?(form.object, method)
+end
+
 other_options ||= {}
 other_options[:help] = [other_options[:help], labels.help].compact.join(" ")
 
@@ -31,7 +36,7 @@ end
 
 %>
 
-<div class="<%= 'required' if presence_validated?(form.object, method) %>">
+<div class="<%= 'required' if options[:required] %>">
 
   <% # the label. %>
   <% unless other_options[:hide_label] == true %>
--- a/bullet_train/app/views/devise/registrations/new.html.erb
+++ b/bullet_train/app/views/devise/registrations/new.html.erb
@@ -17,7 +17,7 @@
            <%= render 'shared/fields/password_field', form: f, method: :password, options: {show_strength_indicator: true} %>
          </div>
          <div>
-           <%= render 'shared/fields/password_field', form: f, method: :password_confirmation, other_options: {error: f.object.errors.full_messages_for(:password).first, hide_custom_error: true} %>
+           <%= render 'shared/fields/password_field', form: f, method: :password_confirmation, options:{ required: true }, other_options: {error: f.object.errors.full_messages_for(:password).first, hide_custom_error: true} %>
          </div>
        </div>
 


kaspth (approved):
I'm in favor of this!

Re: possible problems. I think the same problem with HTML5 validations preventing Rails form submits exists whenever developers use HTML5 validations.

Since these are just about `required:` I don't think we're that worse off. The browser will instruct on missing fields so the user gets feedback inline (and our red dots help connote `required:` for the other fields). Then if they try to sign up with an existing email they get a different error style, but I don't think that's that bad of an experience. I think it's overall more helpful and that we get correct accessibility is great too!

jagthedrummer:
Here's an interesting post talking about how to get better control of client side validations: https://www.jorgemanrubia.com/2019/02/16/form-validations-with-html5-and-modern-rails/

pascallaliberte:
I'd be more in favour of:
* using `options[:required]`
* not passing down the `:required` attribute down to the field by default (deleting the key)
* if other_options[:use_native_required_attribute] is passed, it passes the `:required` attribute down to the field (I don't like the attribute name, but that's the idea...)
* document our stance on our server-side validation default
* for accessibility, when `options[:required]` is passed, regardless of whether other_options[:use_native_required_attribute] is passed, add `aria-required="true"` to the field. The `required` attribute on the form field enforces browser validation, but `aria-required="true"` only informs a screen reader that the field is required without enforcing validation (e.g. for a field labelled "Name", JAWS will apparently announce "Name edit required")

Reactions?

jagthedrummer:
@pascallaliberte, on first read I think that general direction is great! It would fix the missing asterisk issue and give consistency on that point, would improve our accessibility posture, wouldn't force a breaking change, and would give people a way to opt-in to (at least partial) client-side validation. 🤔

jagthedrummer:
OK, just kinda thinking out loud here. I love the general direction that @pascallaliberte proposed and am thinking how to implement it and make the developer experience as intuitive and consistent as possible.

I'm not sure that `options[:required]` is the right spot to handle the asterisk issue.

We document that `options` is always passed down to the underlying Rails field helper, and I don't think we want to start removing things from that hash that people have explicitly passed in. Additionally, if someone wanted to opt-in to _other_ client-side validations, they'd need to pass those in `options`.

So I think we want to leave it so that if someone _only_ wanted client-side validations they could just do `options: {required: true}` and that would be the end of it. (If they do that, we should also add the asterisk and the `aria-required` attribute.)

But if someone _doesn't_ want client-site `required` and they _only_ want to trigger the asterisk that sounds like "additional options for things like hiding labels, displaying specific error messages, etc", which we [document](https://bullettrain.co/docs/field-partials#options-vs-other_options) as being the purpose of `other_options`.

So... that leads me the think that if you _only_ want the asterisk you should do `other_options: {required: true}`.

I don't _love_ the idea of overloading `required` such that it could appear in either `options` or `other_options`, but I think that's the least bad option. Anything that I can think of to disambiguate the two just seems unnecessarily confusing and intuitive.

I think we could reasonably address the possible confusion by documenting `other_options[:required]` in [this section of the docs](https://bullettrain.co/docs/field-partials#globally-available-other_options-options). And then we might add a new section below that giving some examples of common things you'd pass to `options`, where we might specifically give a small example of how to do a couple of different kinds of client-side validations. And that would be a good place to mention that we recommend starting with (and defaulting to) server-side validations.

Does that sound reasonable?

kaspth:
> use other_options[:mark_required] as a way to dress up the visual that the field is required, show an asterisk

What does the `mark_` prefix give us? To me that seems more confusing on first pass. But maybe either `other_options[:required]` or `other_options[:mark_required]` are good enough when documented.

jagthedrummer:
I also think that the `mark_` prefix is confusing. I could see myself regularly thinking "I remember I have to put it in `other_options` and that it's named something _other_ than `required` but I don't remember exactly what".

gazayas:
Would it be possible to use `html_options` like what we have for the super_select partial?

If we at least write `html_options`, maybe that could help the developers understand that it's html specific, whereas `other_options` is a little too ambiguous. This would write `required` inline as opposed to `aria-required=true`, but I'm curious if this is another valid option.

### The Partial Call
```erb
<%= render 'shared/fields/super_select', method: :test_select, html_options: {multiple: true} %>
```

## The Result

image

As a side note, I was confused as to what the difference between `options`, `html_options`, and `other_options` were for a while, so I feel like we should document this, probably with an entire page on its own.

jagthedrummer:
@gazayas, if I'm reading [the docs](https://bullettrain.co/docs/field-partials#options-vs-other_options) right, it sounds like we use `options` as the thing you're describing as `html_options`. It might be good to standardize on one or the other, but I don't think we want to split things between those two (any more than we already have). I think it's probably a separate issue to reconcile those two names.

jagthedrummer:
@pascallaliberte @kaspth @gazayas I opened https://github.com/bullet-train-co/bullet_train-core/pull/513 to supersede this PR and in that one I made the changes discussed above. For now I'm leaving off the `mark_` prefix and have gone with just `other_options: {required: true}` in order to active just the asterisk and `aria-required="true"` while skipping client-side validation.

jagthedrummer:
Closing this in favor of #513.

PR #501, opened 2023-08-30
Get CircleCI running again

jagthedrummer:
Circle wasn't even getting to running the tests because it was failing with this error:

```
Installed version of Google Chrome is 116.0.5845.140 
116.0.5845.140 will be installed
curl: (22) The requested URL returned error: 404 

Exited with code exit status 22
```

This was due to Google releasing a new version of Chrome without releasing a corresponding version of chromedriver.

Some other issues for reference:
https://github.com/CircleCI-Public/browser-tools-orb/issues/90
https://github.com/CircleCI-Public/browser-tools-orb/issues/75

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 orbs:
   ruby: circleci/ruby@2.0.0
   node: circleci/node@5.1.0
-  browser-tools: circleci/browser-tools@1.4.1
+  browser-tools: circleci/browser-tools@1.4.3
 aliases:
   - &ruby_node_browsers_docker_image
       - image: cimg/ruby:3.2.2-browsers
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -63,11 +65,8 @@ jobs:
           path: ~/project
       - browser-tools/install-browser-tools:
           firefox-version: "112.0"
-
-      # TODO: This is a workaround to get `git clone` working, but it shouldn't be here.
-      # https://github.com/CircleCI-Public/browser-tools-orb/issues/62
-      - run:
-          command: rm LICENSE.chromedriver
```
kaspth, line 70: Oh we don't need to `rm` the LICENSE anymore? Nice!

kaspth:
@jagthedrummer thanks! this worked flawlessly over on https://github.com/bullet-train-co/nice_partials/pull/93 too

.circleci/config.yml
kaspth: Seems like we don't, sweet!

Issue #453, opened 2023-08-21
Add a scaffolding system test for new incoming webhooks endpoints.


andrewculver:
Similar to the tests we've got for other Super Scaffolding commands, we should test end-to-end that when someone runs `bin/super-scaffold incoming-webhooks` the resulting controller and model work as expected including being able to receive a webhook.

gazayas:
A couple of mental notes while I work on this.

1. The tunneling docs need to be updated since the `--subdomain` flag for the ngrok CLI tool is deprecated.
2. The help text after super scaffolding pluralizes the provider name when it should just appear as is. The last line in the following, for example, should have `foo_webhooks` as opposed to `foos_webhooks` to work:
```
> bin/super-scaffold crud incoming-webhooks Foo

Writing './app/models/webhooks/incoming/foo_webhook.rb'.
Fixing Standard Ruby on './app/models/webhooks/incoming/foo_webhook.rb'.
Writing './app/controllers/webhooks/incoming/foo_webhooks_controller.rb'.
Fixing Standard Ruby on './app/controllers/webhooks/incoming/foo_webhooks_controller.rb'.
Updating './app/models/webhooks/incoming/foo_webhook.rb'.
Updating 'config/routes.rb'.

1. To receive webhooks in your development environment, you'll need to configure a tunnel.
     See http://bullettrain.co/docs/tunneling for more information.

2. Once you have a tunnel running, you can configure the provider to deliver webhooks to:
     https://your-tunnel.ngrok.io/webhooks/incoming/foos_webhooks
```

https://github.com/bullet-train-co/bullet_train-core/blob/6f67c72da41f78c64e985c930ec8aca08c66347d/bullet_train-incoming_webhooks/lib/bullet_train/incoming_webhooks/scaffolders/incoming_webhooks_scaffolder.rb#L33

gazayas:
Struggling a bit with this one in actually trying to trigger webhook through the ngrok tunnel in the tests. Capybara has `Capybara.app_host`, but tinkering with that or anything similar doesn't seem to do the trick.

I propose creating a controller unit test as opposed to a system test. We would still use the `test/bin` setup file to super scaffold the webhook logic, but I think a controller test will be enough because we can:
1. Verify the newly scaffolded endpoint works
2. Verify the JSON response returns a status of "OK".

Here's what the test would look like:

```ruby
require "test_helper"

[...]

PR #502, opened 2023-08-31
Super Scaffold incoming webhooks test

gazayas:
Closes https://github.com/bullet-train-co/bullet_train-core/issues/453

Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/947

I was stuck on this one for a while for a couple of reasons:
1. I originally tried to write system tests but decided it would be more beneficial to write controller tests. Also, getting the webhook parameters right for this one took some fine-tuning.
2. Since we usually use `Scaffolding::CompletelyConcrete::TangibleThing` and such to transform strings, I initially set out to scaffold using those kinds of strings. It wasn't until I was working in the `bullet_train-incoming_webhooks` transformer itself that it dawned on me that we were already doing this with `Webhooks::Incoming::BulletTrainWebhook` :man_facepalming:

## Try it out
```
bin/super-scaffold incoming-webhooks SomeProvider
```

This will now scaffold a working controller test file.

### DIFF
--- a/bullet_train-incoming_webhooks/test/controllers/webhooks/incoming/bullet_train_webhooks_controller_test.rb
+++ b/bullet_train-incoming_webhooks/test/controllers/webhooks/incoming/bullet_train_webhooks_controller_test.rb
@@ -3,46 +3,40 @@ require "test_helper"
 class Webhooks::Incoming::BulletTrainWebhooksControllerTest < ActionDispatch::IntegrationTest
   include Devise::Test::IntegrationHelpers
 
-  # We can only run this test when Scaffolding things are enabled
-  # because the default webhook triggers come from `CreativeConcept`, etc.
-  unless scaffolding_things_disabled?
-    def setup
-      super
-      @user = FactoryBot.create(:onboarded_user)
-      sign_in @user
-      @team = @user.current_team
-    end
+  def setup
+    super
+    @user = FactoryBot.create(:onboarded_user)
+    @membership = @user.memberships.first
+    @team = @user.current_team
+  end
 
-    test "should get incoming webhook" do
-      creative_concept = Scaffolding::AbsolutelyAbstract::CreativeConcept.create(name: "Test Concept")
-
-      webhook_params = {
+  test "should get incoming webhook" do
+    webhook_params = {
+      data: {
         data: {
-          data: {
-            name: "Test",
-            team_id: {
-              id: @team.id,
-              slug: nil,
-              locale: nil,
-              time_zone: @team.time_zone,
-              created_at: creative_concept.created_at,
-              updated_at: creative_concept.updated_at,
-              being_destroyed: nil,
-            },
-            description: ""
+          name: "Test",
+          team_id: {
+            id: @team.id,
+            slug: nil,
+            locale: nil,
+            time_zone: @team.time_zone,
+            created_at: @membership.created_at,
+            updated_at: @membership.updated_at,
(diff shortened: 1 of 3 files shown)


bullet_train-incoming_webhooks/test/controllers/webhooks/incoming/bullet_train_webhooks_controller_test.rb
```diff
@@ -0,0 +1,48 @@
+require "test_helper"
+
+class Webhooks::Incoming::BulletTrainWebhooksControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
+
+  # We can only run this test when Scaffolding things are enabled
+  # because the default webhook triggers come from `CreativeConcept`, etc.
+  unless scaffolding_things_disabled?
```
jagthedrummer, line 8: @gazayas, I think we don't want this conditional here, if at all possible. After scaffolding an incoming webhook I'd want the tests for the controller to run, even if I'd set `ENV['HIDE_THINGS']` to `true`. (I don't know much about how incoming webhooks work behind the scenes, so apologies if this isn't really possible. Or easy.)

gazayas: @jagthedrummer After looking through the code for a bit and seeing what type of event types we have, I could probably pull this off with a Membership or Invitation. Will try to see what I can do.

gazayas:
@jagthedrummer I updated the tests with a Membership instead, but I'm not entirely satisfied with this. It's definitely hard to envision how we can accomplish this with a system test, but I think ultimately it makes the most sense to do it that way even if we're not sure HOW we're going to accomplish it.

In the meantime, I'm okay with merging this and then leaving the original issue open. I will need some time to just THINK about how to implement things with an ngrok tunnel though... There has to be a way to Super Scaffold the new webhook and ensure the actual flow of the process works.

I'm also not against requiring developers to setup an ngrok tunnel if we are making tests specifically for webhooks, and we can potentially add a warning message in `test/system/super_scaffolding_test.rb` letting them know that these tests aren't being run unless the tunnel is up.

jagthedrummer:
> It's definitely hard to envision how we can accomplish this with a system test, but I think ultimately it makes the most sense to do it that way even if we're not sure HOW we're going to accomplish it.

@gazayas can you help me understand why you think a system test is better than a controller test for this? I feel like I might be missing some context or something. I don't see any reason to try to test through a tunnel. That's just testing the tunnel setup, and that's really out of scope for this kind of test, in my opinion. I think the main thing we want to test is just "if a properly formed request lands at this endpoint, do we handle it correctly?" which I think this accomplishes. Am I missing something?

jagthedrummer:
@gazayas, thanks for the additional context! I think I'm with you now. I'm going to type out my new understanding mainly as a way of organizing my thoughts. I think you're already way ahead of me on this.

So, I think we have two separate but _extremely_ closely related things that we're trying to solve.

1. When we generate code into a project to handle an incoming webhook, we want to also generate tests for that code.
2. We also want to test the process by which we generate the code and the tests mentioned above.

I think this PR completely addresses the first point.

And the second point is what you're getting at with the idea that we need a system test. Something where in the test process we end up running a super scaffolding command for an incoming webhook and then testing that the resulting stuff works as expected, like we do for some of the other super scaffolding stuff.

Issue #395, opened 2023-08-10
Error when adding new boolean CRUD field


Meekohi:
This logic seems wrong for new CRUD fields. It's looking for the boolean in the original `create_students` migration instead of in the new `add_field_to_students` migration:

https://github.com/bullet-train-co/bullet_train-core/blob/8d083834c621fbc250681102d060acad75499fc7/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb#L1269-L1288

`old_line` and `new_line` are `nil` since the boolean is not in the original migration, which causes a crash:

```
Replacing in 'db/migrate/20230628195752_create_students.rb'.
rake aborted!
TypeError: no implicit conversion of nil into String
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/transformer.rb:574:in `gsub!'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/transformer.rb:574:in `replace_in_file'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/transformer.rb:1318:in `block in add_attributes_to_various_views'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/transformer.rb:663:in `each'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/transformer.rb:663:in `each_with_index'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/transformer.rb:663:in `add_attributes_to_various_views'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/bullet_train/super_scaffolding/scaffolders/crud_field_scaffolder.rb:32:in `run'
/Users/meekohi/.rvm/gems/ruby-3.2.2/gems/bullet_train-super_scaffolding-1.2.26/lib/scaffolding/script.rb:107:in `<main>'
[...]

PR #503, opened 2023-08-31
Check if model has foreign key to parent before Super Scaffolding

### LINKED ISSUE #475: `crud` scaffolding should warn if the model doesn't have an association to the parent resource.
I just accidentally scaffolded a resource that didn't have the `team:references` on it when I generated the model. We should be able to check for this reference since they have to pass in the name of the parent resource (in my case it was just `Team`.)

gazayas:
Closes #475.

# Super Scaffolding without `references`
```
> rails g model Project title:string
      invoke  active_record
      create    db/migrate/20230831123427_create_projects.rb
      create    app/models/project.rb
      invoke    test_unit
      create      test/models/project_test.rb
      invoke      factory_bot
      create        test/factories/projects.rb
> bin/super-scaffold crud Project Team title:text_field
rake aborted!
Project does not have a foreign key referencing Team.
Make sure you generate your model with a references type:
rails generate model Project team:references ...

/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb:46:in `run'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/script.rb:104:in `<main>'
<internal:/home/gazayas/.rbenv/versions/3.2.2/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
<internal:/home/gazayas/.rbenv/versions/3.2.2/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding.rb:25:in `run'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/tasks/bullet_train/super_scaffolding_tasks.rake:10:in `block (2 levels) in <main>'
Tasks: TOP => bullet_train:super_scaffolding
(See full trace by running task with --trace)
```

# Super Scaffolding with `references`
```
> rails g model Site team:references name:string
      invoke  active_record
      create    db/migrate/20230831123454_create_sites.rb
      create    app/models/site.rb
      invoke    test_unit
      create      test/models/site_test.rb
      invoke      factory_bot
      create        test/factories/sites.rb
> bin/super-scaffold crud Site Team name:text_field
Writing './app/controllers/account/sites_controller.rb'.
Fixing Standard Ruby on './app/controllers/account/sites_controller.rb'.
Writing './app/views/account/sites/_index.html.erb'.
Writing './app/views/account/sites/_form.html.erb'.
Writing './app/views/account/sites/_menu_item.html.erb'.
Writing './app/views/account/sites/index.html.erb'.
Writing './app/views/account/sites/_site.html.erb'.

...
```

### DIFF
--- a/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb
+++ b/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb
@@ -34,11 +34,29 @@ module BulletTrain
           parents = argv[1] ? argv[1].split(",") : []
           parents = parents.map(&:classify).uniq
           parent = parents.first
+          child_parts = child.split("::")
+          parent_parts = parent.split("::")
+
+          # Pop off however many spaces match.
+          child_parts_dup = child_parts.dup
+          parent_parts_dup = parent_parts.dup
+          parent_without_namespace = nil
+          child_parts_dup.each.with_index do |child_part, idx|
+            if child_part == parent_parts_dup[idx]
+              child_parts.shift
+              parent_parts.shift
+            else
+              parent_without_namespace = parent_parts.join("::")
+              break
+            end
+          end
+
+          # `tr` here compensates for namespaced models (i.e. - `Projects::Deliverable` to `projects/deliverable`).
+          parent_reference = parent_without_namespace.tableize.singularize.tr("/", "_")
+          tableized_child = child.tableize.tr("/", "_")
 
           # Pull the parent foreign key from the `create_table` call
           # if a migration with `add_reference` hasn't been created.
-          tableized_child = child.tableize.tr("/", "_") # Compensate for namespaced models (i.e. - `Projects::Deliverable` to `projects/deliverable`).
-          parent_reference = parent.tableize.singularize.tr("/", "_")
           migration_file_name = `grep "add_reference :#{tableized_child}, :#{parent_reference}" db/migrate/*`.split(":").shift
           migration_file_name ||= `grep "create_table :#{tableized_child}" db/migrate/*`.split(":").shift
           parent_t_references = "t.references :#{parent_reference}"
--- a/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb
+++ b/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb
@@ -37,10 +37,12 @@ module BulletTrain
 
           # Pull the parent foreign key from the `create_table` call
           # if a migration with `add_reference` hasn't been created.
-          migration_file_name = `grep "add_reference :#{child.tableize}, :#{parent.tableize.singularize}" db/migrate/*`.split(":").shift
-          migration_file_name ||= `grep "create_table :#{child.tableize}" db/migrate/*`.split(":").shift
-          parent_t_references = "t.references :#{parent.tableize.singularize}"
-          parent_add_reference = "add_reference :#{child.tableize}, :#{parent.tableize.singularize}"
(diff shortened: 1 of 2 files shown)


bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb
```diff
@@ -35,6 +35,20 @@ def run
           parents = parents.map(&:classify).uniq
           parent = parents.first
 
+          migration_file_name = `grep "create_table :#{child.tableize}" db/migrate/*`.split(":").shift
+          parent_foreign_key = nil
+          File.open(migration_file_name).readlines.each do |line|
+            parent_foreign_key = line.match?("t.references :#{parent.tableize.singularize}")
+            break if parent_foreign_key
+          end
```
jagthedrummer, line 43: I think we might want to alter this block somehow so that we don't assume that the needed reference was added at the time the table was created.

Like what if I did:

```
rails g model Project name:title
rails db:migrate
rails g migration AddTeamToProjects team:references
rails db:migrate
```

The column would be there, but this block would think it was missing and would refuse to generate a scaffold. (I have no idea how common the scenario I described might be.)

jagthedrummer: Oh, I just remembered that we tell people not to run migrations after generating their model. https://bullettrain.co/docs/super-scaffolding#1-basic-crud-scaffolding-with-crud

So maybe the concern I mentioned above is mostly invalid?

Should we add a flag to allow someone to say `--skip-parent-reference-check` or something?

gazayas: @jagthedrummer

> Oh, I just remembered that we tell people not to run migrations after generating their model. https://bullettrain.co/docs/super-scaffolding#1-basic-crud-scaffolding-with-crud

Yes, this is mainly what I had in mind. I primarily wanted to do something like `ModelName.column_names` to check for the reference, but since we don't run the migrations before Super Scaffolding I wasn't able to grab the reference that way.

However, I think your point is still valid. We could technically do this:
```
rails g model Project name:title
rails g migration AddTeamToProjects team:references
rails db:migrate
```

What I could do is check for the add_column reference like we do in this code for boolean buttons (we had a bug revolving around the same issue in #395):

https://github.com/bullet-train-co/bullet_train-core/blob/eb1758bdfedefdbdf41c93e009de952954a65d1c/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb#L1161-L1165

I'm not sure of any other scenarios that could cause trouble for the parent foreign key check, but I think we'll be safe if we at least add this part.

gazayas: @jagthedrummer I made some new editions, it currently looks like this. I decided to use `puts` instead of `raise` since it's easier to look at.

image

jagthedrummer:
@gazayas this was previously failing CI due to the chrome driver problem from last week. I rebased it and force pushed, and now it looks like there are some legit failures.

gazayas:
@jagthedrummer There were two issues:
1. References in migrations usually look like `t.references :projects_step` for namespaced models, but `parent.tabelize` was returning strings like `projects/step`, so I accounted for this with a `tr` call.
2. Children and parent models with the same namespace don't use the namespace in the `t.references` call. So, for a child like `Personality::Note` and a parent like `Personality::Disposition`, I popped off all matching namespaces and just returned `Disposition` to check for something like `t.references :disposition`.

jagthedrummer (approved):

PR #513, opened 2023-09-01
Introduce `other_options[:required]` for field partials

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train-core/issues/379

Supersedes #499, based on [refinements suggested by @pascallaliberte](https://github.com/bullet-train-co/bullet_train-core/pull/499#issuecomment-1701010343)

## The Problem

Currently we show the red asterisk on required fields **only** if the object represented by the form has a validation for `presence` on that field. For instance on the user registration form we don't show an asterisk for `password_confirmation` because Devise doesn't setup a `validates_presence` validation for that field.

CleanShot 2023-08-30 at 10 45 51

But the field really is required. If you fill in a password, but no confirmation and then submit the form, it causes an error.

CleanShot 2023-08-30 at 10 52 33

If a developer were to eject the devise form they could add `options: { required: true}` to the form:

```diff
  <%= render 'shared/fields/password_field', form: f,
    method: :password_confirmation,
+   options: { required: true },
    other_options: {
        error: f.object.errors.full_messages_for(:password).first,
        hide_custom_error: true
    }
  %>
```

That will have the effect of setting the `required` attribute on the password field that ends up getting rendered. That leads to a confusing experience where the label does not indicate that the field is required, but the browser does complain that it's not filled in. 

CleanShot 2023-08-30 at 10 42 29

(And if they didn't also add it to the main `password` field they would not see the browser complain about _that_ field being blank.)

## The fix

This PR makes it so that if `required` is passed into the `other_options` hash (instead of the `options` hash) that the field partial that renders the asterisk will use that value first, and will fallback to the existing validation method if it's not passed in. And it will not trigger a client-side presence check.

```diff
  <%= render 'shared/fields/password_field', form: f,
    method: :password_confirmation,
    other_options: {
+       required: true,
        error: f.object.errors.full_messages_for(:password).first,
        hide_custom_error: true
    }
  %>
```

That makes it so that the red asterisk will show up.

CleanShot 2023-09-01 at 10 16 36

If `required` is passed in `options`, indicating that the developer wants to trigger a client-side check, we'll copy that value into `other_options` so that the asterisk will show up (or not, if they pass `required: false`).

Any time that we are showing the asterisk that indicates a field is required, we also add an `aria-required="true"` attribute to the underlying field that we render. This indicates to browsers and screen-readers that the a value is required, without triggering browser-based validation for presence.

## Possible problems

I think these changes resolve the possible problems mentioned in #499. It will preserve all existing behavior, and just enhance some existing screens by adding `aria-required="true"` to form fields that are required.

There is some potential for confusion over the difference between `options: {required: true}` and `other_options: {required: true}`, but I've tried to address that by updating the docs.

### DIFF
--- a/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
+++ b/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
@@ -20,7 +20,7 @@ other_options ||= {}
 other_options[:help] = [other_options[:help], labels.help].compact.join(" ")
 
 if !other_options.key?(:required)
-  other_options[:required] = options.key?(:required) ? options[:required] : presence_validated?(form.object, method)
+  other_options[:required] = options.fetch(:required) { presence_validated?(form.object, method) }
 end
 
 if other_options[:required]
--- a/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
+++ b/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
@@ -16,12 +16,16 @@ options[:id] ||= form.field_id(method)
 # options[:disabled] ||= !field_editable?(form.object, method) if user_signed_in?
 options[:placeholder] ||= labels.placeholder if labels.placeholder
 
-if !options.key?(:required)
-  options[:required] = presence_validated?(form.object, method)
-end
-
 other_options ||= {}
 other_options[:help] = [other_options[:help], labels.help].compact.join(" ")
+
+if !other_options.key?(:required)
+  other_options[:required] = options.key?(:required) ? options[:required] : presence_validated?(form.object, method)
+end
+
+if other_options[:required]
+  options[:"aria-required"] = true
+end
 
 errors = [method, method.to_s.gsub(/_id$/, '').to_sym].uniq.map { |attribute| form.object.errors.full_messages_for(attribute) }.flatten
 has_errors = errors.any? || partial.error? || other_options[:error].present?
@@ -36,7 +40,7 @@ end
 
 %>
 
-<div class="<%= 'required' if options[:required] %>">
+<div class="<%= 'required' if other_options[:required] %>">
 
   <% # the label. %>
   <% unless other_options[:hide_label] == true %>
--- a/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
+++ b/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
@@ -15,6 +15,11 @@ options ||= {}
 options[:id] ||= form.field_id(method)
 # options[:disabled] ||= !field_editable?(form.object, method) if user_signed_in?
 options[:placeholder] ||= labels.placeholder if labels.placeholder
+
+if !options.key?(:required)
+  options[:required] = presence_validated?(form.object, method)
+end
+
 other_options ||= {}
 other_options[:help] = [other_options[:help], labels.help].compact.join(" ")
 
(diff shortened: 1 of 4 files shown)


bullet_train/docs/field-partials.md
```diff
@@ -63,6 +63,29 @@ For example, to suppress a label on any field, we can use the `hide_label` optio
...
+
+Usually you should stick to using `other_options[:required]` to mark fields as required (_only_ if they don't get marked automatically).
+
+If you pass `options[:required]` you'll be triggering a client-side validation, and many browsers would prvent the form from being submitted with out a value.
+
+The way that browsers communicate info about client-side validations varies from browser to browswer and is not easily controller or styled.
+
+For these reasons we generally recommend that you start with server-side validations and avoid using client-side validations for most simple forms.
+
+If you want to use client-side validations, and you want to be able to control the user experience [this article should give you a good starting place to work from.](https://www.jorgemanrubia.com/2019/02/16/form-validations-with-html5-and-modern-rails/)
```
kaspth, line 88: This seems like a fair bit of documentation for this relatively small quirk. I want to see if I can condense it a bit 😊

bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb
```diff
@@ -15,9 +15,18 @@ options ||= {}
 options[:id] ||= form.field_id(method)
 # options[:disabled] ||= !field_editable?(form.object, method) if user_signed_in?
 options[:placeholder] ||= labels.placeholder if labels.placeholder
+
 other_options ||= {}
 other_options[:help] = [other_options[:help], labels.help].compact.join(" ")
 
+if !other_options.key?(:required)
+  other_options[:required] = options.key?(:required) ? options[:required] : presence_validated?(form.object, method)
```
kaspth, line 23: We can shorten this a bit:

```suggestion
  other_options[:required] = options.fetch(:required) { presence_validated?(form.object, method) }
```

bullet_train/docs/field-partials.md
kaspth, line 88: Ok, here's a stab at it! I think it's important to stress our default presence validation so people know that they might not need to do anything, and that these options are an extra specialization.

If we link out to MDN for the `required` attribute, we can save having to explain how it works here.

The documentation focused a lot on client-side validation, but I don't think we really need all that here, we just need to be clear about our recommendation so users know what to do.

```suggestion
## `required`: through presence validation vs. `options: {required: true}` vs. `other_options:{required: true}`

By default, where there's a presence validation on the model attribute, we add an asterisk to indicate the field is required. For fields without a presence validation, you have options to pass the `:required` detail:

1. `other_options: {required: true}` adds the asterisk to the field manually.
2. `options: {required: true}` adds asterisk but also triggers client-side validation via the [`required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) attribute.

Since client-side validations vary from browser to browser, we recommend relying on server-side validation for most forms, and thus mostly using `other_options[:required]`.
```

jagthedrummer: I could see an argument that my original was too long, but I think this suggestion may have gone too sparse. Lemme think on this a bit. 🤔

kaspth: Yeah, totally! Trying to read it as a user, it just felt like I got a ton of stuff thrown me that didn't necessarily seem to help me decide between the options of the current problem I'm facing. But maybe I'm misunderstanding the scope of this documentation page and that it's meant to be more general purpose.

jagthedrummer: Let's go with the shorter version for now and update it if there's still confusion.

kaspth: Sounds good to me, and yeah, we can always tweak documentation!

PR #514, opened 2023-09-01
Make the showcase theme selector sticky

jagthedrummer:
Adding `fixed` to the theme selector make it stick to the bottom of the browser window instead of being stuck all the way at the bottom of the scroll.

But if we _just_ add the `fixed` class, then the theme selector will overlap the content at the very bottom of the page and we have no way to get to it.

CleanShot 2023-09-01 at 12 41 45

We can fix that by using a little bit of JS to calculate the height of the theme selector and then set that amount of padding on the bottom of the `main` container.

Doing that makes it so that there is a scrollbar if the theme selector will overlap any content at the bottom of the `main` container.

CleanShot 2023-09-01 at 12 38 56

And makes it so that you can scroll all the way down and see all of the content.

CleanShot 2023-09-01 at 12 39 08

### DIFF
--- a/bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
+++ b/bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
@@ -36,11 +36,13 @@
   document.addEventListener("DOMContentLoaded", () => {
     const node = document.getElementById("bullet_train_theme_selects").content.cloneNode(true)
     document.querySelector("main").appendChild(node)
+    const selectorNode = document.getElementById("bt-theme-selector")
+    document.querySelector("main").style.paddingBottom = selectorNode.offsetHeight + "px"
   })
 </script>
 
 <template id="bullet_train_theme_selects">
-  <section class="border-t w-full p-4 flex justify-end left-0 bottom-0">
+  <section class="border-t w-full p-4 flex justify-end left-0 bottom-0 fixed bg-white dark:sc-bg-neutral-900" id="bt-theme-selector">
     <h3 class="pt-2">Bullet Train Light theme options</h3>
 
     <div>
--- a/bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
+++ b/bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
@@ -33,12 +33,15 @@
   BulletTrain.theme = new ___theme()
   BulletTrain.theme.start()
 
-  document.addEventListener("DOMContentLoaded", () => {
+  function placeThemeSelector() {
     const node = document.getElementById("bullet_train_theme_selects").content.cloneNode(true)
     document.querySelector("main").appendChild(node)
     const selectorNode = document.getElementById("bt-theme-selector")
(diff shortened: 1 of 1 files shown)


kaspth:
Ok, something is up with my setup then, because I couldn't get this to look right locally in Safari. Let me pull this down and try again. Sorry for yanking it!

bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
```diff
@@ -36,11 +36,13 @@
...
     const node = document.getElementById("bullet_train_theme_selects").content.cloneNode(true)
     document.querySelector("main").appendChild(node)
+    const selectorNode = document.getElementById("bt-theme-selector")
+    document.querySelector("main").style.paddingBottom = selectorNode.offsetHeight + "px"
   })
 </script>
 
 <template id="bullet_train_theme_selects">
-  <section class="border-t w-full p-4 flex justify-end left-0 bottom-0">
+  <section class="border-t w-full p-4 flex justify-end left-0 bottom-0 fixed bg-white dark:sc-bg-neutral-900" id="bt-theme-selector">
```
kaspth, line 45: Interesting, it didn't look to me like we needed `bg-white dark:sc-bg-neutral-900` because we got `bg-inherit` automatically from being in `main`. Also `sc-bg-neutral-900` isn't safe since that comes from Showcase's CSS build and would go away if Showcase stopped using it.

jagthedrummer: Without those that div is transparent and content shows through it.

CleanShot 2023-09-01 at 12 58 01

kaspth:
I also just found out that the container disappears when running with the application.js that uses Turbo when navigating a page.

jagthedrummer:
Also, I just noticed that the theme selector disappears if you click on any of the sidebar links. This happens both in this branch and in `main`. I suspect it's a turbo thing.

jagthedrummer:
lol, jinx!

kaspth:
> lol, jinx!

sweet 😂, let me pull your fix down!

bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
kaspth: Dang, wut. It seems like we don't really ship dark-mode colors, I tried `dark:bg-black` but that's kinda too much. I don't know if there's a way to transform colors in Tailwind, it would be nice if we didn't have to muck with the build or pollute application.css.

kaspth (approved):
I fixed my local setup and I see this working now! It's just that dark background-color that I don't know what to do with, feel free to merge if that's easier.

bullet_train-themes-light/app/views/showcase/engine/_head.html.erb
jagthedrummer: I couldn't even get `dark:bg-black` to do anything. It seemed to fallback to `bg-white`, which is no good with white text.

PR #515, opened 2023-09-01
WIP : Add an image field partial

### LINKED ISSUE #498: We need an `image` Field Partial
Related to #464. (I think the underlying details have changed a bit since that issue was opened, and the new details point to us needing a new feature more than needing to fix a bug.)

As discussed in #464, we're currently not rendering a Profile Photo on the "Update Your Profile" page when we use ActiveStorage to handle the upload instead of Cloudinary.

CleanShot 2023-08-30 at 09 59 15

The reason that shows up like it does is that when Cloudinary is not enabled we render a simple field partial.

https://github.com/bullet-train-co/bullet_train-core/blob/2b9dea3798d625943d12c1a959d817cac71122df/bullet_train/app/views/account/users/_form.html.erb#L17-L21

I think that we really need a new type of field partial that is just `image`. And then inside the related partials we can do the checks for whether cloudinary is enabled.

That would mean that on user form we could eliminate the conditional and just do:

```ruby
<%= render 'shared/fields/image', method: :profile_photo_id %>
```

I think that developers for the most part shouldn't ever directly see `cloudinary_enabled?`, and it shouldn't be in most high-level templates.

We'll probably want to preserve the `cloudinary_image` field partials, for backwards compatibility, but we could probably just have them render the new `image` partial stuff.

jagthedrummer:
Fixes #498


gazayas:
@jagthedrummer This is great! We should definitely be showing the image whether we're using Cloudinary or ActiveStorage, so this is definitely a step up from #275! Here are some thoughts:

From #498
> I think that we really need a new type of field partial that is just image. And then inside the related partials we can do the checks for whether cloudinary is enabled.
>
> That would mean that on user form we could eliminate the conditional and just do:
> ```ruby
> <%= render 'shared/fields/image', method: :profile_photo_id %>
> ```

I agree, and I think this is the direction we should go in.

> Is there any reason that we shouldn't abstract away the details about Cloudinary vs ActiveStorage? Or to ask it another way, are there reasons that we'd want developers to have to explicitly say _cloudinary_image or _active_storage_image instead of just _image?

I don't think we need a reason for developers to think about the partial itself, I just think we'll have to make it very clear in the documentation (related to your third question) that there are two different types, and that they have to generate their models with the proper type before super scaffolding.

In the [field partials docs](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train/docs/field-partials.md) developers super scaffold Cloudinary images with `cloudinary_image`, so it's pretty straight-forward that they need to setup Cloudinary and look for whatever is necessary in the documentation to get set up properly (["Field partials that integrate with third-party service providers"](https://bullettrain.co/docs/field-partials#field-partials-that-integrate-with-third-party-service-providers)).

What I mean is, imagine we want to super scaffold the following:
```
rails g model Project Team title:string profile:attachment
bin/super-scaffold crud Project Team title:text_field profile:image
```

Since using `attachment` is vanilla Rails, I could see developers completely missing the fact that they have the option between ActiveStorage or Cloudinary. If the developers don't check the documentation and aren't aware, to compensate we can check if the model attribute is an `attachment` or a `string` before trying to super scaffold the image. If we do that, we can write a message in yellow with [add_additonal_step](https://github.com/bullet-train-co/bullet_train-core/blob/b8fe5dd9b789871193119fdcdc2b090d6c7ef570/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb#L1259-L1261) telling the developer that they could use Cloudinary if they wanted to, or if they generated a model with a `string` that they need to add `CLOUDINARY_URL` to `config/application.yml`.
What do you think?

## Separate Partials and Stimulus Controllers
> With ActiveStorage image uploads being essentially a special case of _file_field uploads, should we try to combine all of that into a single partial and Stimulus controller, or should we explicitly separate them in both places? (Currently this PR has a separate partial, but is reusing the Stimulus controller.) I kind of lean towards separating both, just because my gut instinct is that down the road we may want to do more involved UI stuff around images that wouldn't ever apply to file uploads, and it seems potentially messy to have all of that stuff entangled in a single partial/controller.

I'm for separating all of the logic, although I think we can pass with just inserting a TODO for now and perhaps making it a separate PR if it seems like it'll be too much of a hassle for the scope of this one.

bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_active_storage_image.html.erb
```diff
@@ -0,0 +1,56 @@
+<%# TODO: This is mostly just copy pasta from the field_field partial.
```
gazayas, line 1: Pasta! :spaghetti:

gazayas:
I also saw the comment you added in `bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_active_storage_image.html.erb`, so I don't think another one needs to be added :+1:

gazayas:
We'll also need to register `image` as a [valid attribute type](https://github.com/bullet-train-co/bullet_train-core/blob/b8fe5dd9b789871193119fdcdc2b090d6c7ef570/bullet_train-super_scaffolding/lib/scaffolding.rb#L8-L28).

jagthedrummer:
> We'll also need to register `image` as a [valid attribute type](https://github.com/bullet-train-co/bullet_train-core/blob/b8fe5dd9b789871193119fdcdc2b090d6c7ef570/bullet_train-super_scaffolding/lib/scaffolding.rb#L8-L28).

Thanks for the pointer on this.

> Now that I'm looking over things, I have a hunch we're trying to use different commands to super scaffold things. May I ask what commands you had in mind to generate the model and scaffold it?

I honestly haven't gotten as far as thinking in any detail about how to scaffold this, so I'd appreciate any thoughts you have on it. I hadn't fully realized that the `rails g model` command (before super-scaffolding) would have to be different for Cloudinary vs ActiveStorage, but that makes sense. We'll definitely need some explicit call-outs in the docs to guide people in the right direction on that.

### Related aside

It's always seemed a little weird to me that super-scaffolding in BT is a two step process, vs the one step process of scaffolding in Rails.

For things like this `image` attribute question it would be nice if you could _only_ run the super-scaffolding command, without having to do `rails g model` first, and then BT would generate the correct model for you.

Like instead of doing:

```
rails g model Project team:references name:string logo:attachment
bin/super-scaffold crud Project Team name:text_field logo:image
```

It would be nice to be able to just do:

```
bin/super-scaffold crud Project Team name:text_field logo:image
```

In this theoretical world BT would generate the model for you and be able figure out that `name` should be a `string` because it's going to be edited as a `text_field`. And it could inspect `CLOUDINARY_URL` to determine whether to add a `logo_id` field to the table or whether to add `has_one_attachment :logo` to the model class.

(Definitely don't want to tackle any of this as part of this issue/PR, just kind of thinking out loud.)

gazayas:
@jagthedrummer

> It's always seemed a little weird to me that super-scaffolding in BT is a two step process, vs the one step process of scaffolding in Rails.
> 
> For things like this image attribute question it would be nice if you could only run the super-scaffolding command, without having to do rails g model first, and then BT would generate the correct model for you.

I had the same sentiment when I first hopped on Bullet Train, and I personally think it would be cool to Super Scaffold everything in one go. I'm sure Andrew has his reasons for separating the process, but I can't think of anything right now. I understand that it wouldn't fall under the scope of this PR, but it seems like it would be a good improvement to me. And like you said, developers wouldn't have to think about what data type to generate. With `image`, we could potentially pass an option:
```
> bin/super-scaffold crud Project Team name:text_field logo:image{cloudinary}
```

I'll keep it in the back of my mind while this PR progresses :smiley:

jagthedrummer:
@gazayas, I added some documentation about how to scaffold images (and did a small bit of clean up). https://github.com/bullet-train-co/bullet_train-core/pull/515/files#diff-cb892fa51938f0d70a696162c76e52d677065ff39b16c1170b092fac7d88ba1f

Is there anything else you think we need to do on this one before we can merge it?

bullet_train/docs/super-scaffolding.md
```diff
@@ -228,6 +228,46 @@ end
 
 Honestly, it's crazy that we got to the point where we can handle this particular use case automatically. It seems simple, but there is so much going on to make this feature work.
 
+### 7. Scaffolding image upload attributes
+
+Bullet Train comes with two different ways to handle image uploads.
+
+* Cloudinary - This option allows your app deployment to be simpler because you don't need to ship any image manipulation libraries. But it does introduce a dependence on a 3rd party service.
+* ActiveStorate - This option doesn't include reliance on a 3rd party service, but you do have to include image manipulation libararies in your deployment process.
```
gazayas, line 236: ActiveStorate > ActiveStorage

bullet_train-super_scaffolding/lib/scaffolding.rb
```diff
@@ -7,6 +7,7 @@ def self.mysql?
 
   def self.valid_attribute_type?(type)
     [
+      "active_storage_image",
```
gazayas, line 10: @jagthedrummer I don't think we'll need "active_storage_image" here (or "cloudinary_image" from this array either) since we want to use just `image` as a Super Scaffolding type.

For that reason, we should also register "image" as a partial type in `attribute.rb`:

https://github.com/bullet-train-co/bullet_train-core/blob/9983ae6944b40468ea5a7d460371d1626ae5cfe0/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb#L140-L142

We could change it to this (assuming that we don't want to use `cloudinary_image` for backwards compatability):
```ruby
when "image"
  options[:height] = 200
  "image"
```

Since it's not currently registered there, I get the following when trying to scaffold an ActiveStorage image:

```
> rails g model Project team:references title:string foo:attachment
> bin/super-scaffold crud Project Team title:text_field foo:image
```

```
rake aborted!
Invalid field type: image.
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb:166:in `partial_name'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb:656:in `block in add_attributes_to_various_views'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb:632:in `each'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb:632:in `each_with_index'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb:632:in `add_attributes_to_various_views'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb:1409:in `scaffold_crud'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb:91:in `run'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/scaffolding/script.rb:110:in `<main>'
<internal:/home/gazayas/.rbenv/versions/3.2.2/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
<internal:/home/gazayas/.rbenv/versions/3.2.2/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding.rb:26:in `run'
/home/gazayas/work/bt/bullet_train/local/bullet_train-core/bullet_train-super_scaffolding/lib/tasks/bullet_train/super_scaffolding_tasks.rake:10:in `block (2 levels) in <main>'
Tasks: TOP => bullet_train:super_scaffolding
(See full trace by running task with --trace)
```

gazayas (commented):
Besides the ActiveStorage issue I mentioned inline, I noticed that the `show` page currently doesn't work when trying to create a new object for both Cloudinary images and ActiveStorage images.

For the new `_image.html.erb` and `active_storage_image.html.erb` partials, we get the proper invocation in the forms with something like the following:
```erb
<%= render 'shared/fields/image', method: :foo %>
```

However, the show page uses [a completely different image partial](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-themes/app/views/themes/base/attributes/_image.html.erb) because the invocation looks like this:
```erb
<%= render 'shared/attributes/image', attribute: :foo %>
```

I think we'll need to fix the themes image partial before merging this one.

bullet_train-super_scaffolding/lib/scaffolding.rb
gazayas: I'm not sure if I explained all of that too well, so feel free to ask away if something doesn't make sense.

jagthedrummer: Made perfect sense, @gazayas. Thanks for the assist!

I made the changes you suggested and did a little more clean up. Now scaffolding an image with the (new) documentation is working as expected for both Cloudinary and ActiveStorage.

jagthedrummer:
@gazayas, anything else you think we need to do on this one?

gazayas:
@jagthedrummer I think this PR is ready to merge! There is one thing that I don't think we need to change because I think we can safely assume that developers will use either Cloudinary or ActiveStorage images exclusively, but I'll mention it just in case.

Say we have two models where one has an ActiveStorage image and the other has a Cloudinary image:

```
rails g model Project team:references title:string profile:attachment
bin/super-scaffold crud Project Team title:text_field profile:image

rails g model Site team:references title:string photo:string
bin/super-scaffold crud Site Team title:text_field photo:image
```

If Cloudinary is enabled, the ActiveStorage image doesn't show (and vice versa):
image


If we wanted to, we could just check the object type instead of whether Cloudinary is enabled or not (from `bullet_train-themes/app/views/themes/base/attributes/_image.html.erb`)
```erb
<% if object.public_send(attribute).is_a?(ActiveStorage::Attachment) %>
  <%= image_tag photo_url_for_active_storage_attachment(object.public_send(attribute), 
<% else %>
  <%= cloudinary_image_tag object.public_send(attribute), options %>
options) %>
<% end %>
```

However, I think there would be more to edit after that to make it work properly, also like I mentioned earlier I think we can safely assume that developers will exclusively use one or the other, and I personally don't think it's currently worth the effort for the beneficial payoff that this PR offers.

If we want to we can make a new issue, but I'm currently fine with merging things as is!

jagthedrummer:
@gazayas, I agree that it doesn't seem like we need to support mixed use of Cloudinary and ActiveSupport. If someone presents a real use case for that we can always address it later.

🤘 🚀

PR #516, opened 2023-09-01
Some more refactoring and testing of BulletTrain::LoadsAndAuthorizesResource

aaricpittman:
This PR does the following:

* `BulletTrain::LoadsAndAuthorizesResource#regex_to_remove_controller_namespace` to a class method to match expectations in `BulletTrain::LoadsAndAuthorizesResource.account_load_and_authorize_resource`.
* Simplifies `BulletTrain::LoadsAndAuthorizesResource.model_namespace_from_controller_namespace`.
* Simplifies and adds some checks against undefined dependencies `BulletTrain::LoadsAndAuthorizesResource#load_team`.
* Adds some tests around these methods.

### DIFF
--- a/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
+++ b/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
@@ -2,17 +2,12 @@ module BulletTrain::LoadsAndAuthorizesResource
   extend ActiveSupport::Concern
 
   class_methods do
+    # Returns an array of module names based on the classes namespace minus regex_to_remove_controller_namespace
     def model_namespace_from_controller_namespace
-      controller_class_name =
-        if regex_to_remove_controller_namespace
-          name.gsub(regex_to_remove_controller_namespace, "")
-        else
-          name
-        end
-      namespace = controller_class_name.split("::")
-      # Remove "::ThingsController"
-      namespace.pop
-      namespace
+      name
+        .gsub(regex_to_remove_controller_namespace || //, "")
+        .split("::")
+        .slice(0...-1) # drops actual class name
     end
 
     def regex_to_remove_controller_namespace
--- a/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
+++ b/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
@@ -15,6 +15,10 @@ module BulletTrain::LoadsAndAuthorizesResource
       namespace
     end
 
+    def regex_to_remove_controller_namespace
+      raise "This is a template method that needs to be implemented by controllers including LoadsAndAuthorizesResource."
+    end
+
     # this is one of the few pieces of 'magical' functionality that bullet train implements
     # for you in your controllers beyond that is provided by the underlying gems that we've
     # tied together. we've taken the liberty of doing this because it's heavily based on
@@ -200,10 +204,6 @@ module BulletTrain::LoadsAndAuthorizesResource
           )
       end
     end
-  end
-
-  def regex_to_remove_controller_namespace
-    raise "This is a template method that needs to be implemented by controllers including LoadsAndAuthorizesResource."
   end
 
   def load_team
--- a/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
+++ b/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
@@ -7,10 +7,12 @@ module BulletTrain::LoadsAndAuthorizesResource
       name
         .gsub(regex_to_remove_controller_namespace || //, "")
         .split("::")
-        .slice(0...-1) # drops actual class name
+        .tap(&:pop) # drops actual class name
     end
 
     def regex_to_remove_controller_namespace
+      return super if defined?(super)
+
       raise "This is a template method that needs to be implemented by controllers including LoadsAndAuthorizesResource."
     end
 
@@ -202,17 +204,13 @@ module BulletTrain::LoadsAndAuthorizesResource
   end
 
   def load_team
-    # Not all objects that need to be authorized belong to a team,
-    # so we give @team a nil value if no association is found.
-    begin
-      # Sometimes `@team` has already been populated by earlier `before_action` steps.
-      @team ||= @child_object&.team || @parent_object&.team
-    rescue NoMethodError
-      @team = nil
-    end
+    @team ||= @child_object&.try(:team) || @parent_object&.try(:team)
 
-    # Update current attributes.
-    Current.team = @team
+    return unless @team
+
+    if defined?(Current) && Current.respond_to?(:team=)
+      Current.team = @team
+    end
 
     # If the currently loaded team is saved to the database, make that the user's new current team.
     if @team.try(:persisted?)
--- a/bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
+++ b/bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
@@ -1,15 +1,22 @@
 require "test_helper"
 
+# Has to be required here because these tests get run in the context of another application in another repo.
+require "minitest/mock"
 class BulletTrain::LoadsAndAuthorizeResourceTest < ActiveSupport::TestCase
-  class TestClass
+  class TestControllerClass < ActionController::Base
     include BulletTrain::LoadsAndAuthorizesResource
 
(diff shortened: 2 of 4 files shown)


jagthedrummer (changes requested):
Hey @aaricpittman, this seems to break the starter repo tests that we run in CI. Looks like it breaks things very early in the init process and we don't ever get around to actually running tests. https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train-core/1588/workflows/dcd633a8-a835-4e84-922a-4fe46a30867a

aaricpittman:
@jagthedrummer Sorry, I saw the failures but hadn't had a chance to look into them. I was wondering if this would happen. It is failing because the abstract [regex_to_remove_controller_namespace](https://github.com/bullet-train-co/bullet_train-core/blob/7162f6574439e70f1f8c1c011e4d37ce461b094f/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb#L13-L15) method is getting invoked now because it is defined at the appropriate level.

My guess is that it is an `include` order issue. I'll look into why and get it fixed.

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
```diff
@@ -1,23 +1,87 @@
...
-  test "it defines .model_namespace_from_controller_namespace" do
-    assert TestClass.respond_to?(:model_namespace_from_controller_namespace)
+  test "model_namespace_from_controller_namespace returns an array of modules names based on the classes namespace minus regex_to_remove_controller_namespace" do
+    assert_equal ["BulletTrain", "LoadsAndAuthorizeResourceTest"], TestControllerClass.model_namespace_from_controller_namespace
+    assert_equal ["Users"], Users::TestControllerClass.model_namespace_from_controller_namespace
   end
 
   test "it defines .account_load_and_authorize_resource" do
-    assert TestClass.respond_to?(:account_load_and_authorize_resource)
+    assert TestControllerClass.respond_to?(:account_load_and_authorize_resource)
```
kaspth, line 32: There's also a specific assertion for respond tos we could use in the file:

```suggestion
    assert_respond_to TestControllerClass, :account_load_and_authorize_resource
```

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
```diff
@@ -1,23 +1,87 @@
...
-    assert TestClass.new.respond_to?(:load_team)
+    assert TestControllerClass.new.respond_to?(:load_team)
+  end
+
+  test "#load_team returns nil child_object and parent_object are nil" do
+    subject = TestControllerClass.new
+    subject.child_object = nil
+    subject.parent_object = nil
+
+    assert_nil subject.load_team
```
kaspth, line 48: Typically these sorts of `load_` controller `before_action` aren't meant to have a specific return. I think it'd be clearer if we just detected the intended side-effect directly:

```suggestion
    assert_nil subject.instance_variable_get(:@team)
```

bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
```diff
@@ -202,22 +201,14 @@ def account_load_and_authorize_resource(model, options, old_options = {})
...
-    # Update current attributes.
-    Current.team = @team
+    return unless @team
+
+    if defined?(Current) && Current.respond_to?(:team=)
+      Current.team = @team
+    end
 
     # If the currently loaded team is saved to the database, make that the user's new current team.
     if @team.try(:persisted?)
```
kaspth, line 214: Since we've inserted the `return unless @team` above, we can remove the `try` here:

```suggestion
    if @team.persisted?
```

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
```diff
@@ -1,23 +1,87 @@
...
+    end
+    Object.const_set(:Current, current_class)
+
+    team = OpenStruct.new
+    subject = TestControllerClass.new
+    subject.child_object = OpenStruct.new(team: team)
+
+    team.stub(:try, nil) do
+      subject.load_team
+    end
```
kaspth, line 63: Is there a reason we can't pass in a regular Team here? What are we guarding against?

```suggestion
    team = Team.new
    subject = TestControllerClass.new
    subject.child_object = OpenStruct.new(team: team)

    subject.load_team
```

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
```diff
@@ -1,23 +1,87 @@
...
+    subject.parent_object = nil
+
+    assert_nil subject.load_team
+  end
+
+  test "#load_team sets Current attributes if defined" do
+    current_class = Class.new(ActiveSupport::CurrentAttributes) do
+      attribute :team
+    end
+    Object.const_set(:Current, current_class)
```
kaspth, line 55: Will this override an actual `Current` constant in apps? I wonder if this is related to why starter repo tests fail.

I wonder if we'd be better off with something like this in `test_helper.rb`:

```ruby
# Define a fallback `Current` when we're running our tests isolated from the starter repo's `Current`.
unless defined?(::Current)
  class Current < ActiveSupport::CurrentAttributes
    attribute :team
  end
end
```

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
kaspth: Ah, is it because we're running our tests isolated from the https://github.com/bullet-train-co/bullet_train starter repo? And then that's the same for the Mock test below?

kaspth (commented):
Hey thanks for working on this! I had some questions and suggestions for the changes here.

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
jagthedrummer: This may be one of those things where writing tests directly in the gem is somewhat lacking compared to writing tests in the starter repo where we have a more robust BT implementation that we can rely on.

jagthedrummer:
@aaricpittman, if possible it might be nice to split this into some smaller PRs that each do one specific bit of refactoring. That way we could merge the bits that we can, and then deal with the tricky bits separately.

bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
aaricpittman: I'm not calling `try` for null safety. I'm calling it because there is nothing in this gem to guarantee that the value of `@team` responds to `persisted?`

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
aaricpittman: I've updated the test as @kaspth suggested.

I think having to write tests in another application or library because they won't work in the library where the code under-test is written is an indication of one or more software design problems.

`bullet_train-super_load_and_authorize_resource` has some dependency issues.

aaricpittman:
@jagthedrummer / @kaspth made all the suggested changes and fixed the code so the tests pass.

bullet_train-super_load_and_authorize_resource/test/controllers/concerns/bullet_train/loads_and_authorizes_resource_test.rb
aaricpittman: For example, the code of this gem has an undeclared dependency on `cancancan` or, at a minimum, that the class calling `account_load_and_authorize_resource` defines a method `load_and_authorize_resource`.

jagthedrummer: Yeah there are definitely some undeclared dependencies and circular dependencies through a few of the BT packages. I have it on my list to try to sort that out soon.

bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
kaspth: Seems like it'd be better to raise the `NoMethodError` exception with our expected interface to ease debugging, but this code path was already wrapped in `try` so we can keep it.

kaspth (approved):

Issue #507, opened 2023-08-31
`bin/hack --watch-js` assumes packages are in `./local`


jagthedrummer:
I realize I'm off the beaten path by trying to link to an existing copy of `core` instead of checking out a fresh copy in `./local`. I just wanted to document this:

```
$ ./bin/hack --watch-js  bullet_train-fields
Make sure your server is running before proceeding. When you're ready, press <Enter>

Linking JavaScript for bullet_train-fields

sh: line 0: cd: local/bullet_train-core/bullet_train-fields: No such file or directory

bullet_train-fields has been linked.
Preparing to watch changes.

  yarn run v1.22.19
error Command "watch" not found.
  info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Run `bin/hack --clean-js bullet_train-fields` to revert to using the original npm package in your application.
```

jagthedrummer:
Would this be any smoother if we used `yalc` to munge `package.json` the same way we do the `Gemfile`? It would be nice if linking the `core` repo did both gems and npm packages all in one shot.

PR #518, opened 2023-09-04
Point BT gems not originally in Gemfile to custom path with `bin/hack --link`, add roles gem

### LINKED ISSUE #506: `bin/hack --link` still assumes a path of `local` for BT gems not in `Gemfile`
I ran:

```
./bin/hack --link ../bullet_train-core
```

And the gems that were already in the `Gemfile` are linked correctly, like this:

```ruby
gem "bullet_train", path: "../bullet_train-core/bullet_train"
```

But the extra gems that aren't explicitly listed get linked to `./local` like this:

```ruby
gem "bullet_train-fields", path: "local/bullet_train-core/bullet_train-fields"
```

gazayas:
Closes #506

The addition to `framework_packages.yml` also ensures that `bullet_train-roles` is added to the Gemfile.

### DIFF
--- a/bullet_train/config/locales/en/framework_packages.yml
+++ b/bullet_train/config/locales/en/framework_packages.yml
@@ -20,6 +20,8 @@ en:
       git: "bullet-train-co/bullet_train-core"
     bullet_train-outgoing_webhooks:
       git: "bullet-train-co/bullet_train-core"
+    bullet_train-roles:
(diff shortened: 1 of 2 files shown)


gazayas:
#507 is a similar issue but it is more involved since we have to switch directories etc., so we should probably add a fix for it in another PR.

jagthedrummer (approved):

PR #519, opened 2023-09-04
Fix breadcrumbs for edit pages

abradburne:
This partial is included in all generated _breadcrumb partials, so all edit pages are always showing 'team settings', not just the team pages.

This fixes, so it's only shown when editing 'team', not any other model.


jagthedrummer (approved):
Before this PR:

CleanShot 2023-09-05 at 13 18 46

After:

CleanShot 2023-09-05 at 13 19 14

PR #521, opened 2023-09-05
Remove api v2 templates

abradburne:
This is confusing and causes unexpected results in apps.

The gem should only provide v1 versions - and the templates and partials in the api/v2 dir are very out of date from the current v1 versions.

If you are updating your application to a v2 or above, you would expect to provide the templates yourself, and not have the gem provide out-of-date versions.


jagthedrummer:
@abradburne I think tests are failing here because your branch is out of date with `main`. Can you please rebase this branch against `main` and then force push to your repo? (And also the branch for #522?) Once you've done that there shouldn't be any need to touch `bullet_train-super_scaffolding/.circleci/config.yml` so you can remove those changes from both PRs.

bullet_train-api/lib/bullet_train/api/version.rb
```diff
@@ -1,5 +1,5 @@
 module BulletTrain
   module Api
-    VERSION = "1.3.22"
+    VERSION = "1.3.23"
```
jagthedrummer, line 3: @abradburne, something is weird with this PR (and with #522 now). All of these version changes shouldn't be here. I'm not sure what would lead to that, but can you try rebasing against `main` again? If that doesn't do the trick let me know and we'll figure something out.

abradburne: um, weird. Let me take a look.

jagthedrummer:
Closing this in favor of https://github.com/bullet-train-co/bullet_train-core/pull/548

PR #524, opened 2023-09-05
Use `build` in ExampleBot instead of `create` + `rollback`

jagthedrummer:
Joint PR: https://github.com/bullet-train-co/bullet_train/pull/965

This PR makes it so that we use `FactoryBot.build` when generating example records in `ExampleBot`.

Doing this so that we don't have to actually create live records, clone them, rollback, then reset primary keys. Doing all of that work is slow so it's best to avoid doing it. (And having to reset primary keys for something that we expect to be a normal thing kind of gives me the willies.)

### DIFF
--- a/bullet_train-api/lib/bullet_train/api/example_bot.rb
+++ b/bullet_train-api/lib/bullet_train/api/example_bot.rb
@@ -2,7 +2,6 @@ require_relative "../../../app/helpers/api/open_api_helper"
 
 module FactoryBot
   module ExampleBot
-
     def example(model, **options)
       FactoryBot.build(factory(model), **options)
     end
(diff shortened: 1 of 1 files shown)


jagthedrummer:
@newstler, you mentioned that making this change would actually break some things that we don't have tests for. Can you provide an example of something that would break? I'd like to look into getting the needed tests in place, but I don't really know what I'm looking for.

newstler:
Hey @jagthedrummer, yep I remember about this, and I started writing tests for it. Just got busy with fixing bugs in `jbulder-schema` and `bullet_train-api`. How urgent is this?

jagthedrummer:
It's not super urgent, but I would like to keep it moving. Can you give me an idea of what you know about that breaks with this change? I literally have zero idea why this is problematic in any way.

newstler:
I also have to make some investigation, will try to do it this week.
Basically the OpenAPI result should be the same for both cases, my idea is to test it on some real project with a significant amount of models and inter-model relations, and see what's going on there with callbacks and write tests for it.

newstler:
@jagthedrummer Answered in the starter repo https://github.com/bullet-train-co/bullet_train/pull/965#issuecomment-1754068284

PR #525, opened 2023-09-06
Refactor loads and authorize resource

kaspth:
I'm trying to see if there's some fairly simple and safe refactoring we could do here. Just because this method is 140 lines long.

### DIFF
--- a/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
+++ b/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
@@ -51,7 +51,6 @@ module BulletTrain::LoadsAndAuthorizesResource
       end
 
       model_class = model_class_names.find(&:safe_constantize)
-      model_class_name = model_class.name
       unless model_class
         raise "Your 'account_load_and_authorize_resource' is broken. We tried #{model_class_names.join(" and ")}, but didn't find a valid class name."
       end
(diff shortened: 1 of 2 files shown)


bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
```diff
@@ -135,42 +95,35 @@ def account_load_and_authorize_resource(model, options, old_options = {})
...
+        end
       end
 
       if options[:polymorphic]
         prepend_before_action only: collection_actions do
-          possible_sources_of_parent =
-            through_as_symbols.map { |tas| "@#{tas}" }.join(" || ")
-          eval "@#{options[:polymorphic]} ||= #{possible_sources_of_parent}"
+          unless instance_variable_defined?("@#{options[:polymorphic]}")
+            parent = through_as_symbols.lazy.filter_map { instance_variable_get "@#{_1}" }.first
```
kaspth, line 117: Just to demonstrate how `lazy.filter_map {}.first` works. It's so we only loop through just the amount of calls we need to find and return a match:

```ruby
irb(main):004> [1, 2, 3, 4].filter_map { p _1; _1 if _1.even? }.first
1
2
3
4
=> 2
irb(main):005> [1, 2, 3, 4].lazy.filter_map { p _1; _1 if _1.even? }.first
1
2
=> 2
```

kaspth:
@jagthedrummer I've tried cut down on some the incidental complexity in `account_load_and_authorize_resource` while also removing the `eval` calls. What do you think? Are there certain commits in here that moves too much, that we shouldn't ship?

jagthedrummer:
@kaspth I don't know a whole lot about this class so it may take me a little time to give this a thorough review.

kaspth:
@jagthedrummer hey yeah, good point. I'll try to get back to this, this week.

jagthedrummer:
@kaspth, just wanted to check in on your thoughts on this one. I've been reading the current version of that method, and I think I have at least a little more context as to what all is going on this PR, and I don't see anything that's obviously concerning. But I do still think it would be nice to split each logical change into smaller PRs. I can take a stab at that if you're not able to get to it.

kaspth:
@jagthedrummer oh, I forgot about this! I'll take a look at splitting this up tomorrow.

kaspth:
@jagthedrummer ok, I've split all of these out into 5 different PRs. I've stacked them so part-2 merges into part-1 etc. So if you merge them in order, GitHub should point the PR next-in-line branch back to `main` (I can't remember if GitHub can detect a squashed merge and swap the branch so you may need to do that manually, or delete the `part-n` branch from the previous PR).

This should let you sequence and ship these in several different releases if you want to slow the rollout down a bit.

PR #534, opened 2023-09-08
Address field: Add super-scaffolding support, document new field and document updating dependent-fields in a form

pascallaliberte:
Closes #533

This PR adds super-scaffolding support, documents the field and documents the self-updating form mechanism to be used elsewhere in forms.

* [x] Add the new `address_field` to the super-scaffolding mechanism
* [x] Document the new `address_field` compound field partial, its large support for country and region choices out of the box.
* [x] Add the address field to the showcase preview
* [x] Document the `attributes/address` partial and its display options (one_line)
* [x] Clarify the options available for the address_field (can it be read-only? will that render all sub-fields read-only?). See the showcase preview
* [x] Document the new mechanism for auto-updating dependent form fields (e.g. the states/provinces/region field updates based on `country_id` selected). Resolves https://github.com/bullet-train-co/bullet_train-fields/issues/42

PR #541, opened 2023-09-11
Add address_field to field partial help text

gazayas:
@pascallaliberte I know #534 is still a draft, but this CLI help text is relatively new so I just wanted to make sure it was covered. You can run this command to check it out in the terminal:

```
bin/super-scaffold --field-partials
```


jagthedrummer:
@gazayas, are you thinking we should get this in before #534?

gazayas:
@jagthedrummer I think it's fine either way, I just wanted to make sure we put the help text in!

We probably won't need `bin/super-scaffold --field-partials` if we implement https://github.com/bullet-train-co/bullet_train/issues/988 though.

jagthedrummer:
I think I'll wait until #534 is merged. Just so we're not showing people incorrect info.

PR #546, opened 2023-09-12
Require users to enter their current password to change their email

jagthedrummer:
This guards against account takeover in the "laptop left unlocked in a coffee shop" scenario.

This is basically what Devise does out of the box.

CleanShot 2023-09-12 at 12 19 12


bullet_train/config/locales/en/users.en.yml
```diff
@@ -5,20 +5,24 @@ en:
...
     edit:
       section: "%{users_possessive} Profile"
       profile:
         header: Update Your Profile
-        description: You can update the email address where you'll receive notifications and also update the way your name is displayed.
-      password:
-        header: Update Your Password
+        description: You can update the way your name is displayed and your time zone.
+      email_and_password:
+        header: Update Your Email Addresss & Password
```
gazayas, line 15: Addresss > Address

bullet_train/app/views/account/users/edit.html.erb
```diff
@@ -23,18 +23,35 @@
...
-                  <%= render 'shared/fields/password_field', method: :password_confirmation %>
+                  <%= render 'shared/fields/email_field', method: :email, options: {autofocus: true} %>
+
+                  <hr>
+
+                  <p><%= t('.email_and_password.new_password') %></p>
+
+                  <div class="grid grid-cols-1 gap-y gap-x sm:grid-cols-2">
+                    <div class="sm:col-span-1">
+                      <%= render 'shared/fields/password_field', method: :password, options: {show_strength_indicator: true}, other_options: { required: false } %>
```
gazayas, line 40: `other_options: { required: false }` in action! :100:

bullet_train/app/views/account/users/edit.html.erb
```diff
@@ -23,18 +23,35 @@
         <%= render 'account/users/oauth' %>
 
         <%= render 'account/shared/box', divider: true do |box| %>
-          <% box.title t('.password.header') %>
+          <% box.title t('.email_and_password.header') %>
+          <% box.description t('.email_and_password.description') %>
```
gazayas, line 27: Apologies for being really nitpicky, but could this be `email_or_password`? We can technically edit one or the other, so I feel like matching these keys with the other additions in this PR is appropriate. I'm open if this is commonly done in other web apps though.

gazayas (commented):
I like placing the email and password fields together, I think it makes more sense. Also confirmed that the mobile style is working fine.

I noticed that I wasn't getting an email notifying me after changing either my email address or password. Is this a devise thing in development that I'm missing?
```
When you change these credentials we'll email you a confirmation.
```

Looks good to me besides that.

bullet_train/app/controllers/concerns/account/users/controller_base.rb
```diff
@@ -24,15 +24,15 @@ def edit
   def show
   end
 
-  def updating_password?
```
kaspth, line 27: I wonder if we can yank this? I'm guessing apps won't be relying on this since they may not be overriding `update` which uses it, but I don't know if that's a safe bet.

kaspth (approved):
Overall looks good!

bullet_train/app/views/account/users/edit.html.erb
jagthedrummer: @gazayas I originally went with `email_or_password` but then realized that it reads weird on the page if the header says "Update Your Email Address or Password" when you can actually change both with the same form. I finally decided that for purposes of displaying them it should be "and" mostly because they live in the same form and so the form is for "email _and_ password". (Technically it's "email and/or password" but I don't think we want to be _that_ precise.)

bullet_train/app/controllers/concerns/account/users/controller_base.rb
jagthedrummer: 🤔 Hmm... That's a good question, @kaspth. If we leave it in, should it do what it's always done and just check for `:password` or should we make it an alias for `updating_password_or_email?`?

gazayas (approved):

bullet_train/app/controllers/concerns/account/users/controller_base.rb
jagthedrummer: OK, instead of yanking the method I'm leaving it as it was, but adding a deprecation message. If someone is using it they'll see something like this in their console:

```
DEPRECATION WARNING: #updating_password? is deprecated. \
  Use #updating_password_or_email? instead. \
  (called from block in update at /Users/jgreen/projects/bullet-train-co/bullet_train-core/bullet_train/app/controllers/concerns/account/users/controller_base.rb:43)
```

PR #550, opened 2023-09-13
Run tests for individual gems in GitHub Actions

jagthedrummer:


bullet_train-outgoing_webhooks/Gemfile
```diff
@@ -11,3 +11,7 @@ gem "sprockets-rails"
 gem "jbuilder"
 # Start debugger with binding.b [https://github.com/ruby/debug]
 # gem "debug", ">= 1.0.0"
+
+gem "bullet_train", path: "../bullet_train"
+gem "bullet_train-api", path: "../bullet_train-api"
+gem "bullet_train-super_scaffolding", path: "../bullet_train-super_scaffolding"
```
gazayas, line 17: @jagthedrummer Since these are pointing to local paths, I'm assuming they're going to be removed after the tests work? I remember when I tried working on the same issue and there were a lot of cross-dependencies to make tests pass (with just CI and not GitHub actions).

This looks like a lot of work! Hats off for tackling this one :clap:

jagthedrummer: Yeah, this turned out to be more involved than I had hoped it would be. I was hoping it would be a quick little experiment that would Just Work, but that didn't turn out to be the case.

jagthedrummer:
Closing in favor of #569

PR #532, opened 2023-09-07
Documentation about doing a stepwise upgrade

jagthedrummer:
Fixes #470 

Related to #529 & https://github.com/bullet-train-co/bullet_train/pull/968

## Notes for reviewers:

This is currently a "forward looking" PR that is making some assumptions about a future state of affairs. I'm assuming that we'll end up merging #529 & https://github.com/bullet-train-co/bullet_train/pull/968, and I'm assuming that those will mark the `1.4.0` release.

Once those are merged (and once we're a little ways down the road, to `1.4.1` and beyond) we'll have a few scenarios that we should cover as far as upgrading:

* Apps currently on `1.2.x` (or earlier) could either
    * upgrade to `1.3.0` and then do stepwise upgrades through `1.3.x` until they get to `1.4.0`
    * upgrade directly to `1.4.0` and do stepwise from there
    * upgrade directly to `latest`
* Apps currently on `1.3.x` could either
    * do stepwise upgrades through `1.3.x` until they get to `1.4.0`
    * upgrade directly to `1.4.0` and do stepwise from there
    * upgrade directly to `latest`
* Apps currently on `1.4.x` could either
    * do stepwise upgrades through `1.4.x`
    * upgrade directly to `latest`

I've attempted to cover all of these scenarios. All of the scenarios are pretty similar, but in the places where they differ they're pretty big differences. I opted to duplicate some info in order to have each scenario documented in a way that can be read through top to bottom and is internally consistent. I thought that a bit of duplication would be better than trying to cram everything into one giant list that would then have a bunch of "If you're on x.x.x and moving to y.y.y do this, but if you're on y.y.y and moving to z.z.z then do that".

PR #558, opened 2023-09-18
Update the upgrading documentation concerning ejected views

gazayas:
I didn't write a review on #532, but I read through the new documentation to upgrade a personal Bullet Train application I'm working on to 1.4.3, and the process went smoothly 🚀

One thing I noticed was that if you're upgrading to a new version and you have an ejected theme, no merge conflicts arise for those ejected themes because they're originally from `bullet_train-core` and not the starter repository. I struggled for a while trying to find out why my tests weren't passing although everything was working fine on the starter repository. I eventually found out it was because my ejected files hadn't been updated (specifically `_date_and_time.html.erb` and `_date.html.erb` since we updated the daterangepicker logic recently), so I had to manually copy and paste the contents from `bullet_train-core` to get my tests working again.

Since I usually work on the starter repository itself day to day it wasn't difficult for me to compare the tests there with my Bullet Train app, but I think most developers will only create a new app and not have a local setup of just the raw starter repository on their machine. For that reason they'll only have the context of their personal app, look at the tests and perhaps simply think that the new version is broken, or that they took a wrong step in resolving merge conflicts. I added this comment to the upgrading documentation to guide developers to review `bullet_train-core` just in case if they ejected files which may be outdated.

I also think we should write the comment at the top of ejected files more specifically. For example:

```
> bin/resolve shared/attributes/date --eject
```

```
<% # What currently gets written to ejected files %>
<% # Ejected from `bullet_train-themes-1.4.3`. %>

<% # What we should write %>
<%# Ejected from bullet_train-themes-1.4.3/app/views/themes/base/attributes/_date.html.erb #>
```

It should also be noted that we don't write **ANY** comment at all to ejected views when ejecting the files for a new theme:
```
> rake bullet_train:themes:light:eject[foo]
```

Lastly, I also fixed some typos along the way.

### DIFF
--- a/bullet_train/docs/upgrades.md
+++ b/bullet_train/docs/upgrades.md
@@ -68,7 +68,7 @@ git checkout -b updating-bullet-train-1.4.1
 
 ### 5. Merge in the newest stuff from Bullet Train and resolve any merge conflicts.
 
-Each version of the starter repo is tagged, so you can merge in the tag from the upstread repo.
+Each version of the starter repo is tagged, so you can merge in the tag from the upstream repo.
 
 ```
 git merge v1.4.1
@@ -116,5 +116,5 @@ git branch -d updating-bullet-train-1.4.1
 ```
 
 Alternatively, if you're using GitHub, you can push the `updating-bullet-train-1.4.1` branch up and create a
-PR from it and let your CI integration do it's thing and then merge in the PR and delete the branch there.
+PR from it and let your CI integration do its thing and then merge in the PR and delete the branch there.
 (That's what we typically do.)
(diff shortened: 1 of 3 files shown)


olbrich:
@gazayas When upgrading recently, I simply re-ran the `rake bullet_train:themes:light:eject[foo]` command.  This brought in a bunch of files that didn't exist when I first ejected and ended up showing diffs for the ones that changed.  Perhaps we should recommend re-ejecting the theme and any previously ejected views.

One downside to this approach is that if there are files that get renamed or deleted from the theme, those files will remain in the ejected theme.  Maybe we should recommend deleting the ejected theme folder before re-ejecting.

gazayas:
> Perhaps we should recommend re-ejecting the theme and any previously ejected views.

Interesting approach, I hadn't thought of this before but showing the diffs this way could be helpful.

> One downside to this approach is that if there are files that get renamed or deleted from the theme, those files will remain in the ejected theme.

Yes, I'm afraid things will get jumbled like this when working with new versions.

Maybe we can create a new task that's similar to re-ejecting the theme, but tries to merge in the new file content to the existing files? Then we could add a line in the documentation to tell developers to run the task if they've ejected anything in the past.

jagthedrummer (approved):

PR #569, opened 2023-09-20
[CI] Run tests for individual gems in GitHub Actions

jagthedrummer:
In #550 I realized that I was making the same changes over and over to each individual gem and that instead we should probably ship some base controller/models in the main `bullet_train` gem. So this is an attempt at doing that.


jagthedrummer:
OK, I finally got this all working, but this situation is a little ridiculous for a single PR:

CleanShot 2023-11-15 at 12 13 33

There are a few main things that this PR does, and I'm going to try to split it into smaller PRs as best I can.

* Adds some missing dependencies to `.gemspec` files. These dependencies end up getting resolved when we're installing things within the starter repo, but then some gem tests fail when we run them outside of the starter repo context.
* Add some dependencies to various `Gemfile`s. These make it so that in dev/test each gem will use the peer version of the gems in the `core` repo.
* Adds some "missing" files like `Account::ApplicationController`. These are files that we ship in the starter repo, and which various gems expect to be present. So when we try to test the gems outside of the starter repo we need to provide those files in some way. We could potentially put a copy of them in `test/dummy` for each gem, but that seems unwieldy. Instead I'm adding them directly to the gem that seems appropriate and then the copy in the starter repo will be as if that file has been ejected.
* Some CI plumbing in the `.github` diretory
* A few random things that I had to change here and there to get things to work. These are probably the changes most likely to cause problems or to be things that we don't want to do, and it would be much easier to review them if they're not mixed in with all this other stuff.

jagthedrummer:
I handled the `.gemspec` and `Gemfile` dependencies in #667 and #668. Then rebased this branch and things aren't quite as bad:

CleanShot 2023-11-15 at 17 02 52

Tomorrow I'll see what I can do to split this up further. And I'm sure there are some bits that I need to clean up.

bullet_train-outgoing_webhooks/test/dummy/app/models/team.rb
```diff
@@ -0,0 +1,23 @@
+class Team < ApplicationRecord
+  include Teams::Base
+  include Webhooks::Outgoing::TeamSupport
```
jagthedrummer, line 3: Added this entire file just so that this `include` will be available to tests.

bullet_train-api/app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,3 @@
+class Api::V1::ApplicationController < ActionController::API
+  include Api::Controllers::Base
+end
```
jagthedrummer, line 3: This and the next couple of files are just providing a default implementation of files that we ship in the starter repo. We can think of the files that we ship in the starter repo as being "ejected" versions of these files. I added these in the gem itself so that we don't have to supply these files in `test/dummy` of every singe gem that ends up needing them.

bullet_train-incoming_webhooks/app/controllers/webhooks/incoming/bullet_train_webhooks_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Webhooks::Incoming::BulletTrainWebhooksController < Webhooks::Incoming::WebhooksController
   def create
-    Webhooks::Incoming::BulletTrainWebhook.create(data: JSON.parse(request.body.read)).process_async
+    Webhooks::Incoming::BulletTrainWebhook.create(data: JSON.parse(request.raw_post)).process_async
```
jagthedrummer, line 3: For some reason `request.body.read` is returning an empty string `''` which then causes `JSON.parse` to throw an error. Using `raw_post` seems to work. At least for tests...

bullet_train-incoming_webhooks/config/models/roles.yml
jagthedrummer, line 1: This is a "default implementation" of something we ship with the starter repo. Putting it in this specific spot feels kinda wrong, but that's what I had to do to get tests to pass.

bullet_train-outgoing_webhooks/config/models/webhooks/outgoing/event_types.yml
jagthedrummer, line 1: Another "default implementation" that feels like it's not in quite the right place, but it had to be here for tests to work.

bullet_train-outgoing_webhooks/test/dummy/config/database.yml
jagthedrummer, line 1: This and the next few files are just setting up the database for the test env of `bullet_train-outgoing_webhooks`.

bullet_train-outgoing_webhooks/test/integration/api/a1/webhooks/outgoing/endpoints_controller_test.rb
```diff
@@ -3,6 +3,9 @@
 class Api::V1::Webhooks::Outgoing::EndpointsControllerTest < ActionDispatch::IntegrationTest
   setup do
     @controller = Api::V1::Webhooks::Outgoing::EndpointsController
+
+    @team = Team.first
+    @endpoint = @team.webhooks_outgoing_endpoints.first
```
jagthedrummer, line 8: Grabbing a couple of objects for reference later in the tests. Doing this since we can't (and shouldn't try to) rely on hard coded ids.

bullet_train-outgoing_webhooks/test/setup/active_record.rb
```diff
@@ -1,36 +1,8 @@
...
-  create_table :teams, force: true do |t|
-    t.string :name, null: false
-    t.timestamps null: false
-  end
-
-  create_table :webhooks_outgoing_endpoints, force: true do |t|
-    t.references :team
-    t.string :url, null: false
-    t.string :name, null: false
-    t.integer :event_type_ids, array: true, default: []
```
jagthedrummer, line 20: `sqlite3` doesn't seem to support `array: true` in CI, so I just changed this gem to test against postgres which is what we target for deployment anyway.

bullet_train-outgoing_webhooks/test/integration/api/a1/webhooks/outgoing/endpoints_controller_test.rb
```diff
@@ -3,6 +3,17 @@
...
   setup do
     @controller = Api::V1::Webhooks::Outgoing::EndpointsController
+
+    5.times do |n|
+      Team.create! name: "Generic name #{n}"
+    end
+
+    5.times do |n|
+      Webhooks::Outgoing::Endpoint.create! team_id: Team.first.id, name: "Generic name #{n}", url: "http://example.com/webhook-#{n}"
+    end
```
jagthedrummer, line 13: Moved these into the `setup` method so that they're created within the transaction for running tests. When they were outside of the `setup` method in the `setup/active_record.rb` file the records created would stay in the DB between test runs.

bullet_train-scope_validator/bin/rails
jagthedrummer, line 1: Added a missing binstub.

bullet_train-super_scaffolding/test/dummy/app/controllers/account/application_controller.rb
jagthedrummer, line 1: Removed this so that we use the default implementation from the `core` gems.

bullet_train/app/controllers/account/application_controller.rb
jagthedrummer, line 1: Added a default implementation of this file so that it doesn't need to be in `test/dummy` for each gem.

bullet_train-api/test/dummy/app/controllers/application_controller.rb
jagthedrummer, line 1: I removed a bunch of `*/test/dummy/app/controllers/application_controller.rb` `*/test/dummy/app/models/application_record.rb` files so that gems will use the default implementation that we ship.

bullet_train/app/models/concerns/users/base.rb
```diff
@@ -2,6 +2,7 @@ module Users::Base
   extend ActiveSupport::Concern
 
   included do
+    extend Devise::Models
```
jagthedrummer, line 5: Explicitly including this so that things will work in tests when the rail ties don't do the thing.

bullet_train/lib/bullet_train/configuration.rb
```diff
@@ -24,7 +24,7 @@ def enable_bulk_invitations
       end
 
       def incoming_webhooks_parent_class_name
-        @@config&.incoming_webhooks_parent_class_name
+        @@config&.incoming_webhooks_parent_class_name || "ApplicationRecord"
```
jagthedrummer, line 27: Need to supply a default value here or we get a failure like this:

```
Error:
Webhooks::Incoming::BulletTrainWebhooksControllerTest#test_should_get_incoming_webhook:
NoMethodError: undefined method `constantize' for nil:NilClass
    /Users/jgreen/projects/bullet-train-co/bullet_train-core/bullet_train-incoming_webhooks/app/models/webhooks/incoming/bullet_train_webhook.rb:1:in `<top (required)>'
    <internal:/Users/jgreen/.asdf/installs/ruby/3.2.2/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
    <internal:/Users/jgreen/.asdf/installs/ruby/3.2.2/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
    /Users/jgreen/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/zeitwerk-2.6.12/lib/zeitwerk/kernel.rb:30:in `require'
    /Users/jgreen/projects/bullet-train-co/bullet_train-core/bullet_train-incoming_webhooks/app/controllers/webhooks/incoming/bullet_train_webhooks_controller.rb:3:in `create'
```

jagthedrummer:
OK, I'm not sure that there's much more I can do so split this up unless I split it into one PR for each gem that I'm turning on in CI. I can do that if this is too overwhelming to review as it is.

All together this gets it so that we run the in-gem test suite for each gem in GitHub Actions. Also it's now possible to run gem tests locally from within each gem instead of needing to run them from within the context of the starter repo. You can prepend the test command with `CI=1` to get the same loading behavior as in CI, which comes in handy if CI is complaining about something.

@gazayas @kaspth @pascallaliberte @andrewculver, I'd love to get some eyes on this whenever y'all have some time.

bullet_train-api/app/controllers/api/v1/application_controller.rb
gazayas: I remember this being really annoying because the tests complained about the files not being there. This seems like a good option!

I think it would be helpful to provide a comment in these files briefly explaining the situation so in the future someone on the Bullet Train team doesn't come across it and mistake it for a "duplicated" file in the starter repository.

bullet_train-incoming_webhooks/config/models/roles.yml
```diff
@@ -0,0 +1,40 @@
+# This file is just here to make tests pass when run from inside this gem directory.
+# Seems like maybe we need a better resolution method for this file?
+# It wasn't found if I put this file in test/dummy/config/models/
```
gazayas, line 3: It looks like we just use `./config/models/roles.yml` to find the file. Maybe we can use the `Dir.glob` or something similar to find the file more dynamically and prioritize the `test/dummy` directory when working in the test environment.

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb#L486-L487

bullet_train-outgoing_webhooks/test/setup/active_record.rb
```diff
@@ -1,42 +1,12 @@
...
-
-5.times do |n|
-  Team.create! name: "Generic name #{n}"
-end
-
-5.times do |n|
-  Webhooks::Outgoing::Endpoint.create! team_id: Team.first.id, name: "Generic name #{n}", url: "http://example.com/webhook-#{n}"
+# We declare a Team model here so that we can have it
+# include Webhooks::Outgoing::TeamSupport
+# which we don't ship enabled by default. Maybe we should?
```
gazayas, line 8: Interesting, I would imagine we should ship it by default, especially since all of the Developers > API/Webhooks logic is available on a new Bullet Train app out of the box.

gazayas (commented):
I confirmed that all of the tests were working for me locally. As I mentioned in one of my inline comments, the only request from me is to add comments to the files that would be considered as "ejected" in the starter repository. There seems to be a number of them, and I just don't want to get confused in the future if I forgot why we added those files.

This is a big PR, kudos for getting everything to work properly!

bullet_train-incoming_webhooks/config/models/roles.yml
jagthedrummer: Ah, I'd bet we want:

```ruby
role_file = File.join(Rails.root, "config", "models", "roles.yml")
```

bullet_train-outgoing_webhooks/test/setup/active_record.rb
jagthedrummer: I just double checked the starter repo and there we _do_ include `Webhooks::Outgoing::TeamSupport` by default. It's just the copy in the `core` repo that doesn't include it. I'll try adding it there and see what happens.

jagthedrummer: Nope, no luck. Turns out that the team model is in the `bullet_train` gem, and that gem doesn't depend on the `outgoing_webhooks` gem, so that module isn't always available. (Relatedly, we should probably take a look at the dependency graph that we've constructed for all of these gems and make sure that what we currently have is actually what we want. I'm not sure we have a fully accurate story about which gems are really a "core" part of the BT stack vs which ones just happen to live in the `core` repo. But that's for another issue/PR.)

bullet_train-incoming_webhooks/config/models/roles.yml
jagthedrummer: Aaaaand, it's not just there that we need to fix. These two lines also point to that file directly, and just throwing a `Rails.root` in there doesn't seem to work. I'm just going to leave all of that stuff alone and will open an issue about resolving those paths differently.

PR #570, opened 2023-09-21
Add Delete Team button

### LINKED ISSUE #231: Can't delete Team after creating new Membership via invite
Context: https://discord.com/channels/836637622432170028/836637623048601633/1091054108561133648

We don't have Team deletion in place by default, so I'm curious how we want to handle this one. If we want to add a "Delete Team" button on the "Switch Teams" page, I can start working on that.

gazayas:
This doesn't close #231 because that one deals with a foreign key issue with Memberships, but this at least sets us up so we can delete teams. If we're okay with this, I'll try to figure out the foreign key issue too.

Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/1023

### DIFF
--- a/bullet_train/app/controllers/concerns/account/teams/controller_base.rb
+++ b/bullet_train/app/controllers/concerns/account/teams/controller_base.rb
@@ -110,13 +110,20 @@ module Account::Teams::ControllerBase
 
   # # DELETE /teams/1
   # # DELETE /teams/1.json
-  # def destroy
-  #   @team.destroy
-  #   respond_to do |format|
-  #     format.html { redirect_to account_teams_url, notice: 'Team was successfully destroyed.' }
-  #     format.json { head :no_content }
-  #   end
-  # end
+  def destroy
+    if current_user.teams.size == 1
+      respond_to do |format|
+        format.html { redirect_to edit_account_team_url(@team), alert: t("account.teams.notifications.cannot_delete_last_team") }
+        format.json { head :no_content }
+      end
+    else
+      @team.destroy
+      respond_to do |format|
+        format.html { redirect_to account_teams_url, notice: t("account.teams.notifications.destroyed") }
+        format.json { head :no_content }
+      end
+    end
+  end
 
   private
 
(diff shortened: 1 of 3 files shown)


gazayas:
Also, although delete buttons are usually on the show page, I think most users will try to delete their team via the Team Settings page which can be found by the link in the navbar.

bullet_train/app/controllers/concerns/account/teams/controller_base.rb
```diff
@@ -110,13 +110,20 @@ def update
...
   # # DELETE /teams/1.json
-  # def destroy
-  #   @team.destroy
-  #   respond_to do |format|
-  #     format.html { redirect_to account_teams_url, notice: 'Team was successfully destroyed.' }
-  #     format.json { head :no_content }
-  #   end
-  # end
+  def destroy
+    if current_user.teams.size == 1
```
jagthedrummer, line 114: More of an aside about BT in general than a comment about this particular PR, but I wish that we didn't need this conditional. I understand why we do, given the current state of things. But it would be nice if there was a little better handling of the "you don't belong to any teams" case so that we didn't need to prevent this.

jagthedrummer (approved):
Looks great! (And the button is right where I'd expect it to be.)

bullet_train/app/controllers/concerns/account/teams/controller_base.rb
gazayas: @jagthedrummer I totally agree, and I didn't want to write it this way initially. At first I tried to make it a `before_action` callback, but it didn't seem right when I started to actually write it out.

I knew we had the following code in the code base which raises this `RemovingLastTeamAdminException` error, but I didn't flesh out how to do it with a team so I just pushed things as is. I’ll see if I can make it raise an error for Teams in a similar way.

https://github.com/bullet-train-co/bullet_train-core/blob/3aa6c0a76f70d846bb23b2746a6368f71eab4df1/bullet_train/app/controllers/concerns/account/memberships/controller_base.rb#L70-L84

PR #579, opened 2023-09-27
Eject showcase preview when ejecting partials

### LINKED ISSUE #576: When ejecting a theme, also eject the showcase previews
Showcase is useful when working on a theme, but it can be pretty annoying to have to manually copy the previews into the showcase directory for every partial.

On a related note, it would probably be a good idea to eject any showcase previews for a partial when it is ejected, even if the entire theme is not ejected.

gazayas:
Closes #576.

I know there's duplicate code here, but we don't currently use the `eject` logic that lives within `BulletTrain::Resolver` when ejecting an entire theme (the ejection logic [there](https://github.com/bullet-train-co/bullet_train-core/blob/4a5818ed1b10893ddf1ebd8198fdb91269610932/bullet_train-themes-light/lib/tasks/application.rb#L6) is all inline).

### DIFF
--- a/bullet_train-themes-light/lib/tasks/application.rb
+++ b/bullet_train-themes-light/lib/tasks/application.rb
@@ -36,6 +36,8 @@ module BulletTrain
           "bullet_train-themes-light" => "light"
         }.each do |gem, theme_name|
           gem_path = `bundle show --paths #{gem}`.chomp
+          showcase_partials = Dir.glob("#{gem_path}/app/views/showcase/**/*.html.erb")
+
           `find #{gem_path}/app/views/themes`.lines.map(&:chomp).each do |file_or_directory|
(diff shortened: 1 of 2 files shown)


jagthedrummer (approved):

bullet_train-themes-light/lib/tasks/application.rb
```diff
@@ -49,6 +51,28 @@ def self.eject_theme(theme_name, ejected_theme_name)
...
+            showcase_preview = nil
+            showcase_partials.each do |partial|
+              has_showcase_preview = partial.match?(/#{file_name}$/)
+              if has_showcase_preview
+                showcase_preview = partial
+                break
+              end
+            end
+
+            if has_showcase_preview
```
kaspth, line 67: I think you can replace this with a `find` call:

```ruby
showcase_preview = showcase_partials.find { _1.end_with?(file_name) }
if showcase_preview
```

kaspth (commented):
nice!

PR #580, opened 2023-09-27
Add exceptions initializer from starter repository

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/1043


jagthedrummer:
@gazayas is there a reason that these need to be in an initializer instead of each error living in it's own class somewhere? I'm thinking maybe `bullet_train/app/errors`? Or perhaps somewhere in `bullet_train/lib`?

gazayas:
@jagthedrummer No reason besides it was originally there in the first place, and I wasn't the one who originally made the file. I like the file being in `bullet_train/lib`, so I moved it over.

By the way the tests are failing in the starter repository because I accidentally named the branches differently...

gazayas:
Will look into the failure that's coming up here, might be legit. Seems like the `RemovingLastTeamException` is legit.

gazayas:
This one should be good to go. The tests are failing locally, but pass with the fix in https://github.com/bullet-train-co/bullet_train/pull/1044

PR #582, opened 2023-09-28
Refactor showcase partial lookup when ejecting files

gazayas:
This PR refactors the showcase partial lookup logic as per @kaspth's [comment](https://github.com/bullet-train-co/bullet_train-core/pull/579#discussion_r1339416134) in #579.

I also have a question after looking at the starter repository.

Say we eject a custom theme with `rake bullet_train:themes:light:eject[foo]`. All of the partials that show up in the application are ejected to `app/views/themes/foo`. However, when we eject the showcase files, they get ejected to `app/views/showcase` and don't show up within the `foo` theme. Is this what we want? Do we want the showcase partials to be theme-specific?

### DIFF
--- a/bullet_train-themes-light/lib/tasks/application.rb
+++ b/bullet_train-themes-light/lib/tasks/application.rb
@@ -54,17 +54,8 @@ module BulletTrain
 
             # Look for showcase preview.
             file_name = target_file_or_directory.split("/").last
-            has_showcase_preview = false
-            showcase_preview = nil
-            showcase_partials.each do |partial|
-              has_showcase_preview = partial.match?(/#{file_name}$/)
-              if has_showcase_preview
-                showcase_preview = partial
-                break
-              end
-            end
-
-            if has_showcase_preview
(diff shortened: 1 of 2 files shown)


jagthedrummer:
Nice simplification.

On the question about the location where we eject showcase partials it seems to me (with limited context) that putting them into a theme specific directory would make sense if the other files are going into theme specific directories. I bet @kaspth will have a much more informed opinion.

kaspth:
> Say we eject a custom theme with rake bullet_train:themes:light:eject[foo]. All of the partials that show up in the application are ejected to app/views/themes/foo. However, when we eject the showcase files, they get ejected to app/views/showcase and don't show up within the foo theme. Is this what we want? Do we want the showcase partials to be theme-specific?

We don't want them to show up as theme-specific in Showcase, that was a request from @andrewculver.

gazayas:
Got it, I'll leave it as is then. Thanks for the follow up guys!

PR #592, opened 2023-10-02
Remove now unneeded inline billing support code.

kaspth:
We inserted load hooks in https://github.com/bullet-train-co/bullet_train-core/pull/335, and then used them in https://github.com/bullet-train-pro/bullet_train-billing/pull/19

We've shipped releases with that new code, so we can remove our inline backing code from here.


jagthedrummer (approved):
👍 Thanks for circling back on this clean up.

PR #600, opened 2023-10-04
Bug fix for super scaffolding an address_field

jagthedrummer:
The PR for testing super scaffolding of Addresses (https://github.com/bullet-train-co/bullet_train/pull/986) landed right before the PR for automatically generating a migration during super scaffolding (https://github.com/bullet-train-co/bullet_train-core/pull/547). As a result a bug / test-failure crept in.

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2585/workflows/1d20bad0-bcad-45f3-96e3-9ac377eec600/jobs/13420?invite=true#step-115-1450_67

The `Address` model `belongs_to :addressable`, which means that the relation is stored on the `Address` side, and not on the owning model. So that means that when we super scaffold an address field it turns out that we don't need to represent that in the migration for the model that will own the address.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/script.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/script.rb
@@ -8,7 +8,7 @@ require "scaffolding/routes_file_manipulator"
 require_relative "../bullet_train/terminal_commands"
 
 FIELD_PARTIALS = {
-  address_field: "string",
+  address_field: nil,
   boolean: "boolean",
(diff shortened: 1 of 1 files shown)


jagthedrummer:
I'm going to go ahead and merge this to get tests back to green, but we can always revert and do something different if need be.

gazayas:
@jagthedrummer Yes, this makes sense to me! I'll keep an eye out on it in case there's something I'm overlooking.

PR #602, opened 2023-10-04
Rails generators for super scaffolding

jagthedrummer:
This adds a series of Rails generators that can serve as a front-end to super scaffolding. 

The idea is to make things just a _little_ closer to the normal Rails way in order to make it easier for new devs to acclimate to a Bullet Train environment.

When doing a bare `rails g` a section like this will now show up in the output:

```
SuperScaffold:
  super_scaffold
  super_scaffold:field
  super_scaffold:incoming_webhooks
  super_scaffold:join_model
  super_scaffold:oauth_provider
```

Here are the equivalent `bin/super-scaffold` commands for each of the new generators:

| rails g                                 | bin/super-scaffold                  |
|-----------------------------------------|-------------------------------------|
| rails g super_scaffold                  | bin/super-scaffold crud             |
| rails g super_scaffold:field            | bin/super-scaffold crud-field       |
| rails g super_scaffold:incoming_webhook | bin/super-scaffold incoming-webhook |
| rails g super_scaffold:join_model       | bin/super-scaffold join-model       |
| rails g super_scaffold:oauth_provider   | bin/super-scaffold oauth-provider   |

---

## Original PR description (for reference to the conversation below):

This is currently just a proposal for adding a set of Rails generators as front-ends to the existing super scaffolding commands.

So far this makes it so that you can super scaffold a model and UI by doing this:

```
rails g super_scaffold ...
```

Instead of (or in addition to) the existing way of:

```
bin/super-scaffold crud ...
```

My goal here is to make things just a _little_ closer to the normal Rails way in order to make it easier for new devs to acclimate to a Bullet Train environment.

To my mind `bin/super-scaffold crud` is very directly analogous to `rails g scaffold`, and I'd like for the generator command to be close in a way that's easy to remember. So I think `rails g super_scaffold` makes the most sense. I can see that there's an argument to be made for it being `rails g super_scaffold:crud` to preserve existing semantics, but I personally don't think it's a very strong argument. (If we want to preserve that we could have both.)

For the other super scaffolding commands, I think we could nest some or all of them. I think `rails g super_scaffold:field` "feels" right. But I'm less sure about `rails g super_scaffold:oauth_provider`, for instance.

I'm not so sure that `join-model`, `oauth-provider`, and `incoming-webhooks` are really related to "scaffolding" so much as they're just fairly normal "code generation", to make a possibly subtle distinction. I tend to think that "scaffolding" is specifically about generating the model & the UI simultaneously whereas just doing `rails g model` does not count as "scaffolding".

If we want to go this route I'll add generators for all of the other actions, and will add some legit tests to make sure things get passed through correctly.

Would love to get some feedback on the general idea (and the namespacing question), so I'm going to tag some folks. Y'all feel free to tag anyone else that you'd like to chime in. @andrewculver @gazayas @kaspth @pascallaliberte @newstler


bullet_train-super_scaffolding/lib/generators/super_scaffold/super_scaffold_generator.rb
```diff
@@ -0,0 +1,49 @@
...
+  remove_class_option :skip
+
+  def generate
+    # In order to pass the args to the rake task we have to join them
+    # into a single string with spaces and commas escaped. (Due to
+    # weird rake syntax requirements.) We add the name of the specific
+    # super_scaffolding command that we want to invoke to the beginning
+    # of the argument string.
+    new_args = (['crud'] + ARGV).join("\\ ").gsub(",", "\\,")
+    rake "bullet_train:super_scaffolding[#{new_args}]"
```
kaspth, line 47: Another option, although it's more work, is to flip the order so the rake task is an alternative invocation for the generators.

Kinda something like this?

```sh
#!/usr/bin/env ruby

require "shellwords"

command = "super_scaffold"

if %w[crud field join-model oauth-provider incoming-webhooks].include?(ARGV.first)
 command += ":#{ARGV.shift}"
end

system "bin/rails generate #{command} #{Shellwords.escape(ARGV)}"
```

Then we could retain the class options like you're talking about.

Another benefit to using generators is that we can leverage `create_file` which will automatically generate a diff in case there's already a file at that path.

kaspth:
I like this!

I'm wondering if `super` is getting in the way and we could lean on our somewhat common `bt` prefix?

Like this:

```sh
bin/rails g bt:scaffold
bin/rails g bt:scaffold:field

bin/rails g bt:oauth_provider
bin/rails g bt:join_model
bin/rails g bt:incoming_webhooks # Could also be `bt:webhooks:incoming` if one day add `bt:webhooks:outgoing`
```

I think that would get us out of the asymmetry you're mentioning where `super-scaffold join-model` doesn't feel quite right.

Does anybody if Rails lets you filter generators by a prefix, e.g. if we used `bt:` could we list all of our generators?

jagthedrummer:
> I'm wondering if `super` is getting in the way

I'm not sure that `super` is getting in the way as much as it's being overused. When it comes to scaffolding I _really_ like the name "Super Scaffolding" for the Bullet Train version because it's very descriptive and is super (ha!) easy to remember.

> Does anybody if Rails lets you filter generators by a prefix, e.g. if we used bt: could we list all of our generators?

I'm not aware of a way to filter generators beyond something like `rails g | grep bt:` or something, but a bare `rails g` does list generators grouped by namespace.

bullet_train-super_scaffolding/lib/generators/super_scaffold/super_scaffold_generator.rb
jagthedrummer: I think this is an option that we should eventually take, but I'm not sure we want to go straight for it. I think it would involve quite a large refactoring of the code generation process. Using Rails generator helpers like `create_file` would definitely buy us a lot of extra functionality, like diffs and automatic support for `--force` and `--pretend` and what not.

jagthedrummer: Or maybe we go ahead and start with the simple version that you've suggested here and just put off the use of `create_file` and friends for a later PR... 🤔

gazayas:
> I'm not so sure that join-model, oauth-provider, and incoming-webhooks are really related to "scaffolding" so much as they're just fairly normal "code generation", to make a possibly subtle distinction. I tend to think that "scaffolding" is specifically about generating the model & the UI simultaneously whereas just doing rails g model does not count as "scaffolding".

I agree with this paragraph the most, although I also agree with the following:

> When it comes to scaffolding I really like the name "Super Scaffolding" for the Bullet Train version because it's very descriptive and is super (ha!) easy to remember.

Maybe we could do something like this:
```
rails g super_scaffold ...
rails g super_scaffold:crud_field ...
rails g bt:oauth_provider
etc...
```

Yes, we'd be splitting the two, but considering that the word `scaffold` carries a specific meaning in the Rails world, this seems to be the most natural way to write things for me. Everything else that doesn't fit under the umbrella of Super Scaffolding can just go inside a `bt` namespace.

## CLI
While we're here, I think it's worth noting that we [explored creating a CLI](https://github.com/bullet-train-co/bullet_train-core/pull/36) in the past. It was closed due to [this reason](https://github.com/bullet-train-co/bullet_train-core/pull/36#issuecomment-1431003661):
> Going to skip this. Don't want it to be confusing for anyone and we're already so deep with docs and videos that have the bin/super-scaffold syntax.

gazayas:
I'm down!

jagthedrummer:
Personally I don't think the need for consistency outweighs simplicity in this case, so I'd be inclined to skip the `bt:` prefix for the `super_scaffold` generators for sure. And I could really take 'em or leave 'em for the others. 🤔

As for the previous CLI exploration, I don't think that skipping it due to documentation already existing is a very strong argument. Docs are super easy to change, and we're not (yet) proposing the removal of the `bin/super-scaffold` script.

jagthedrummer:
OK, @kaspth & @gazayas I rounded this out with the rest of the generators. I opted to use `bullet_train` instead of `bt` for the prefix for the non-scaffold generators for a couple of reasons.

1. We already have some tasks in the `bullet_train` namespace.
2. It seems less mysterious when doing a plain `rails g`.

When using `bullet_train` a plain `rails g` gives you this:

```
BulletTrain:
  bullet_train:incoming_webhooks
  bullet_train:join_model
  bullet_train:oauth_provider
```

But using `bt` gives you this:

```
Bt:
  bt:incoming_webhooks
  bt:join_model
  bt:oauth_provider
```

If people would prefer to use `bt` I don't have a strong preference and I'm OK with that. We may want to open some issues to move other tasks in to that same namespace. (It would be nice if we could use the `bt` namespace but have the tasks grouped under a `BulletTrain` heading in the plain `rails g` output.)

Just for reference here are all of the new tasks that are added by this PR:

```
BulletTrain:
  bullet_train:incoming_webhooks
  bullet_train:join_model
  bullet_train:oauth_provider

SuperScaffold:
  super_scaffold
  super_scaffold:field
```

I went ahead and duplicated some of the usage instructions into the various `USAGE` files, and I set up the generators to expect the arguments that the script expects. I think eventually we can get that usage/documentation stuff out of the script itself so that it remains a little more compact.

Here are some ideas for some possible follow-on PRs that we might want to do after this is merged. I'm not sure that we'd want to do all of them, just listing things that have come to mind.

* Update `bin/super-scaffold` to use the generators under the hood
* Update `bin/super-scaffold` to output a message saying that it's deprecated
* Update docs to use `rails g ...` instead of `bin/super-scaffold ...`
* Remove usage/documentation from the main scaffolding script
* Move code generation/manipulation stuff directly into the generators

gazayas:
> If people would prefer to use bt I don't have a strong preference and I'm OK with that.

Same. If anything, we should probably get the opinion of the developers. Should we take a poll on Discord?

If I'm understanding this correctly, for [these lines](https://github.com/bullet-train-co/bullet_train-core/pull/602/files#diff-4fee44829732c7d5383ef624fbad3bef2fa9526f4ab0649b321dce63670877e1R30-R31), I think this would be well worth looking into soon after so we don't duplicate logic for options in case it already exists somewhere in the transformer. Thinking specifically of this portion in the documentation: https://bullettrain.co/docs/super-scaffolding/options

jagthedrummer:
Thanks for pointing me to that page of documentation, @gazayas. I thought that there were more options available but I was having trouble tracking them down. I'm going to go ahead and add those options to the main `super_scaffold` generator. (That's the only one that those apply to, right?)

gazayas:
@jagthedrummer I actually think there are some flags that will work for other types of scaffolders.

For example, we can do this:
```
bin/super-scaffold crud Project Team title:text_field
bin/super-scaffold crud-field Project body:text_field --skip-locales
```

There are ones that are specifically for the main scaffolder though, so I guess we'll just have to sift through them and apply them correctly:
```
# A couple of examples for the main scaffolder
--namespace=customers
--sidebar="ti-world"
...
```

jagthedrummer:
So.... I'm seriously reconsidering my previously stated position about what counts as "super scaffolding" in Bullet Train.

In preparing the `USAGE` files for all of these generators I'm seeing that some of them actually generate _way_ more stuff than I thought they did.

`oauth_provider` generates models, controllers, views, routes, tests, and factories. So I think it's doing almost _everything_ that I think of as being part of "scaffolding". ✅ 

`incoming_webhooks` generates a model, controller, route, and tests. So not quite as much "stuff" but it does give you everything that you need to use that new model/feature. So, I think ✅ 

`join_model` is a little less clear to me. It generates a model and a migration. Which is pretty much exactly what `rails g model` does. But it _does_ also alter the models that it's joining, which does save you from needing to open those files and add a couple of `has_many` statements. So that's some extra functionality that's a level above most standard "code generation". And it seems to be used primarily for enabling a followup `rails g super_scaffold:field` command. So... maybe it's also "super scaffolding"? Even if only because it would be weird to be out by itself? ✅?

So maybe it should be this:

```
SuperScaffold:
  super_scaffold
  super_scaffold:field
  super_scaffold:incoming_webhooks
  super_scaffold:join_model
  super_scaffold:oauth_provider
```

bullet_train-super_scaffolding/lib/generators/super_scaffold/super_scaffold_generator.rb
kaspth: @jagthedrummer not sure how I've missed your messages in this thread, but everything you said sounds good to me! And 👍 to taking it slow if we want to convert how super scaffolding executes.

kaspth:
@jagthedrummer just using `super_scaffold` as the task namespace sounds good to me!

bullet_train-super_scaffolding/lib/generators/super_scaffold/USAGE
```diff
@@ -0,0 +1,36 @@
+Description:
+  Generate a production ready CRUD interface for a model. See the docs for more details:
+  https://bullettrain.co/docs/super-scaffolding
+
+  E.g. a Team has many Sites with some attributes:
+    rails g super_scaffold Site Team name:text_field url:text_area
+
+  E.g. a Section belongs to a Page, which belongs to a Site, which belongs to a Team:
+    rails gsuper_scaffold Section Page,Site,Team title:text_field body:text_area
```
kaspth, line 9:
```suggestion
    rails g super_scaffold Section Page,Site,Team title:text_field body:text_area
```

bullet_train-super_scaffolding/lib/generators/super_scaffold/join_model/join_model_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+  #
+  # Maye we have comparable options for some of these and should retain them and
+  # pass them through to the rake task?
+
+  remove_class_option :skip_namespace
+  remove_class_option :skip_collision_check
+  remove_class_option :force
+  remove_class_option :pretend
+  remove_class_option :quiet
+  remove_class_option :skip
```
kaspth, line 38: Should we define a `SuperScaffolding::Generators::Base < Rails::Generators::Base` class for some of all this config?

And potentially sharing the `class_option`s below too?

kaspth (approved):

bullet_train-super_scaffolding/lib/generators/super_scaffold/join_model/join_model_generator.rb
jagthedrummer: Yeah, we should definitely do something to DRY these up. Good suggestion.

PR #604, opened 2023-10-06
Stop Super Scaffolding if model already exists

### LINKED ISSUE #603: Raise error when Super Scaffolding if `rails g model` was previously executed
Now that #547 has been merged, we don't need to run `rails g model ...`. However, if someone DOES run it, we don't raise an error, and the Scaffolder continues to add/edit files:

```
> bin/super-scaffold crud Project Team title:text_field --navbar="ti.ti-world"
Generating Project model with 'bin/rails generate model Project team:references title:string'

The name 'Project' is either already used in your application or reserved by Ruby on Rails. Please choose an alternative or use --skip-collision-check or --force to skip this check and run this generator again.
Writing './app/controllers/account/projects_controller.rb'.
Fixing Standard Ruby on './app/controllers/account/projects_controller.rb'.
Writing './app/views/account/projects/_index.html.erb'.
Writing './app/views/account/projects/_form.html.erb'.
Writing './app/views/account/projects/_menu_item.html.erb'.
...
```

In trying to find a solution for this, I kept track of the thread and waited for it to stop:
```ruby
begin
  generator_thread = Thread.new do
    `#{generation_command}` unless @options["skip-migration-generation"]
  end
  generator_thread.join # Wait for the thread to finish.
rescue => e
  # Handle error...
end
```

However, this doesn't actually rescue the error, and I can't quite seem to grab the error message which says the model already exists.

The original Rails error message can be found [here](https://github.com/rails/rails/blob/2d271a4b901e7a7dae8ba4e5e83edfebdee278e8/railties/lib/rails/generators/base.rb#L279-L281). Although I would like to call `class_collisions`, it's in a private method.

Would we be okay with just scanning our instances of ActiveRecord and seeing if the model exists already?

gazayas:
Closes #603.

I tried using `spawn`, `IO.popen` and `trap` as @kaspth suggested in the issue, but couldn't quite get it to work. What I decided to do was wait for the process to end, then compare `git status` before running `bin/rails generate model` and after running it.

Now we get the following if the model exists:
```
> rails g model Foo team:references title:string
      invoke  active_record
      create    db/migrate/20231006051223_create_foos.rb
      create    app/models/foo.rb
      invoke    test_unit
      create      test/models/foo_test.rb
      invoke      factory_bot
      create        test/factories/foos.rb
> bin/super-scaffold crud Foo Team title:text_field
Generating Foo model with 'bin/rails generate model Foo team:references title:string'

The name 'Foo' is either already used in your application or reserved by Ruby on Rails. Please choose an alternative or use --skip-collision-check or --force to skip this check and run this generator again.
> 
```

If this doesn't suffice I can look into those methods I mentioned at the beginning again, but I wanted to at least push this.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/script.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/script.rb
@@ -60,7 +60,7 @@ end
 
 # All untracked files begin with a tab (i.e. - "\tapp/models/model.rb").
 def get_untracked_files(status_lines)
-  `git ls-files --other --directory --exclude-standard`.split("\n")
+  `git ls-files --other --exclude-standard`.split("\n")
 end
 
 def check_required_options_for_attributes(scaffolding_type, attributes, child, parent = nil)
--- a/bullet_train-super_scaffolding/lib/scaffolding/script.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/script.rb
@@ -50,6 +50,20 @@ def standard_protip
   puts "If you do that, you can reset to your last commit state by using `git checkout .` and `git clean -d -f` ."
 end
 
+def git_status
+  `git status`.split("\n")
+end
+
+def has_untracked_files?(status_lines)
+  status_lines.include?("Untracked files:")
+end
+
+# All untracked files begin with a tab (i.e. - "\tapp/models/model.rb").
+def get_untracked_files(status_lines)
+  idx = status_lines.index("Untracked files:")
+  status_lines[idx..].select { |lines| lines.match?(/\t/) }
+end
+
 def check_required_options_for_attributes(scaffolding_type, attributes, child, parent = nil)
   tableized_parent = nil
 
@@ -183,7 +197,19 @@ def check_required_options_for_attributes(scaffolding_type, attributes, child, p
       puts "Adding new fields to #{child} with '#{generation_command}'".green
     end
     puts ""
-    `#{generation_command}` unless @options["skip-migration-generation"]
+
+    unless @options["skip-migration-generation"]
+      untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
+      generation_thread = Thread.new { `#{generation_command}` }
+      generation_thread.join # Wait for the process to finish.
+
+      newly_untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
+      if (newly_untracked_files - untracked_files).size.zero?
+        # Rails will output an error if nothing was generated,
+        # so we don't have to write anything extra upon exiting.
+        exit 1
+      end
+    end
   end
 end
 
--- a/bullet_train-super_scaffolding/lib/scaffolding/script.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/script.rb
@@ -60,8 +60,7 @@ end
 
 # All untracked files begin with a tab (i.e. - "\tapp/models/model.rb").
 def get_untracked_files(status_lines)
-  idx = status_lines.index("Untracked files:")
-  status_lines[idx..].select { |lines| lines.match?(/\t/) }
+  `git ls-files --other --directory --exclude-standard`.split("\n")
 end
 
 def check_required_options_for_attributes(scaffolding_type, attributes, child, parent = nil)
--- a/bullet_train-super_scaffolding/lib/scaffolding/script.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/script.rb
@@ -204,8 +204,13 @@ def check_required_options_for_attributes(scaffolding_type, attributes, child, p
 
       newly_untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
       if (newly_untracked_files - untracked_files).size.zero?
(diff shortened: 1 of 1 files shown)


bullet_train-super_scaffolding/lib/scaffolding/script.rb
```diff
@@ -50,6 +50,20 @@ def standard_protip
...
+def has_untracked_files?(status_lines)
+  status_lines.include?("Untracked files:")
+end
+
+# All untracked files begin with a tab (i.e. - "\tapp/models/model.rb").
+def get_untracked_files(status_lines)
+  idx = status_lines.index("Untracked files:")
+  status_lines[idx..].select { |lines| lines.match?(/\t/) }
+end
+
```
kaspth, line 66: It looks like there's a way to get just the untracked files without needing the parsing https://stackoverflow.com/a/11024039

What if we did something like this?

```suggestion
def untracked_files_from_git
  `git ls-files --other --directory --exclude-standard`.split("\n")
end

```

bullet_train-super_scaffolding/lib/scaffolding/script.rb
```diff
@@ -183,7 +197,19 @@ def check_required_options_for_attributes(scaffolding_type, attributes, child, p
...
     puts ""
-    `#{generation_command}` unless @options["skip-migration-generation"]
+
+    unless @options["skip-migration-generation"]
+      untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
+      generation_thread = Thread.new { `#{generation_command}` }
+      generation_thread.join # Wait for the process to finish.
+
+      newly_untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
+      if (newly_untracked_files - untracked_files).size.zero?
```
kaspth, line 207: With the other change we could do something like this:

```suggestion
      untracked_files = untracked_files_from_git
      generation_thread = Thread.new { `#{generation_command}` }
      generation_thread.join # Wait for the process to finish.

      newly_untracked_files = untracked_files_from_git
      if (newly_untracked_files - untracked_files).none?
```

bullet_train-super_scaffolding/lib/scaffolding/script.rb
```diff
@@ -183,7 +197,19 @@ def check_required_options_for_attributes(scaffolding_type, attributes, child, p
       puts "Adding new fields to #{child} with '#{generation_command}'".green
     end
     puts ""
-    `#{generation_command}` unless @options["skip-migration-generation"]
+
+    unless @options["skip-migration-generation"]
+      untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
+      generation_thread = Thread.new { `#{generation_command}` }
+      generation_thread.join # Wait for the process to finish.
```
kaspth, line 204: I can't remember if this is where `system generation_command` or `spawn generation_command` is meant to be used. But this seems good.

kaspth (approved):
Nice! I left some suggestions where I could trim the implementation a bit. But feel free to merge if you don't agree with them 😄

bullet_train-super_scaffolding/lib/scaffolding/script.rb
gazayas: I definitely like using `git` directly better.
For some reason the command is returning ignored directories for me?
```
> git ls-files --other --directory --exclude-standard
.knapsack_pro/
.ruby-lsp/
app/models/project.rb
db/migrate/20231009090516_create_projects.rb
test/factories/projects.rb
test/models/project_test.rb
tmp/storage/
```

I think we can just get away with this:
```
> git ls-files --other --exclude-standard
app/models/project.rb
db/migrate/20231009090516_create_projects.rb
test/factories/projects.rb
test/models/project_test.rb
```

Just pushed the changes :+1:

bullet_train-super_scaffolding/lib/scaffolding/script.rb
```diff
@@ -183,7 +196,19 @@ def check_required_options_for_attributes(scaffolding_type, attributes, child, p
...
+
+      newly_untracked_files = has_untracked_files?(git_status) ? get_untracked_files(git_status) : []
+      if (newly_untracked_files - untracked_files).size.zero?
+        # Rails will output an error if nothing was generated,
+        # so we don't have to write anything extra upon exiting.
```
jagthedrummer, line 208: Is the Rails error message going to be useful in a Bullet Train context? I think the error that people would see is going to say:

> The name 'Foo' is either already used in your application or reserved by Ruby on Rails. Please choose an alternative or use --skip-collision-check or --force to skip this check and run this generator again.

But I don't think that adding `--skip-collision-check` or `--force` to the `bin/super-scaffold` command is going to work.

Maybe we need to add a message telling people that they could use `--skip-migration-generation`?

gazayas: That's a good point, I'm not sure if I'll be able to prevent the default rails message from being output, but I'll try to at least write a meaningful message.

gazayas: Just pushed https://github.com/bullet-train-co/bullet_train-core/pull/604/commits/39150dd17a6f7a48d75e300ff0237518d44b0a97

PR #609, opened 2023-10-10
Support Trilogy MySql Adapter

jwright:
This PR adds support for the `Trilogy` adapter for roles.

There are several cases where BT interprets the database connection name to check for a MySql database adapter in order to use different syntax to communicate with the database.


jagthedrummer (approved):
Thanks, @jwright!

PR #614, opened 2023-10-14
Show relative path when ejecting showcase partials

### LINKED ISSUE #613: When ejecting the showcase partials, the path is too specific
Relates to #262.

When ejecting a theme we get a comment like...

    <% # Ejected from bullet_train-themes-light-1.6.0/app/views/themes/light/attributes/_block.html.erb %>

When ejecting the showcase partials the comment is like...

    <% # Ejected from /Users/kolbrich/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/bullet_train-themes-light-1.6.0/app/views/showcase/previews/partials/_link_row.html.erb %>

The second one contains the entire full path, which probably won't be the same on a teammate's machine and will cause an unnecessary diff if someone else on the team ever re-ejects a theme (say, during an upgrade).

gazayas:
Fixes #613.

This is hard to test locally because released gem paths look like `bullet_train-themes-light-1.6.3` when local gems don't have the version number. This should work though.

### DIFF
--- a/bullet_train-themes-light/lib/tasks/application.rb
+++ b/bullet_train-themes-light/lib/tasks/application.rb
@@ -62,7 +62,7 @@ module BulletTrain
               FileUtils.mkdir_p(directory)
               FileUtils.touch(partial_relative_path)
               `cp #{showcase_preview} #{partial_relative_path}`
-              new_files[partial_relative_path] = showcase_preview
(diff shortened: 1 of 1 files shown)


gazayas:
Just confirmed this works for released gems:
```
<% # Ejected from bullet_train-themes-light-1.6.4/app/views/showcase/engine/_head.html.erb %>
```

PR #619, opened 2023-10-18
Delete unneeded session key in invite-only mode

gazayas:
Fixes the issue discussed on Discord.

Also, I looked into the session logic for general invitations as well, and I'm not seeing anything in particular that seems to need attention.

We delete the uuid here:

https://github.com/bullet-train-co/bullet_train-core/blob/74333901d4ea56baec8e0066a444471bbe650ecf/bullet_train/app/controllers/concerns/account/invitations/controller_base.rb#L58

And also here for omniauth callbacks:

https://github.com/bullet-train-co/bullet_train-core/blob/74333901d4ea56baec8e0066a444471bbe650ecf/bullet_train-integrations/app/controllers/concerns/account/oauth/omniauth_callbacks/controller_base.rb#L165-L166

The first one deletes the key upon account creation, so the invitation uuid stays in the session until then. During that time, users can only access the new registration page and the new sign in page.

## Tests

I'm getting a failing test for the "Update Profile" button so I'll see if I can fix that.

### DIFF
--- a/bullet_train/app/controllers/concerns/controllers/base.rb
+++ b/bullet_train/app/controllers/concerns/controllers/base.rb
@@ -82,6 +82,9 @@ module Controllers::Base
       unless helpers.invited?
         redirect_to [:account, :teams], notice: t("teams.notifications.invitation_only")
       end
+
+      # We only want the session to store the key on the redirect from InviteOnlySupport.
+      session[:invitation_key] = nil
     end
   end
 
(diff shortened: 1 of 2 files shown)


gazayas:
The test must have been a fluke, it's passing for me locally now.

jagthedrummer:
@gazayas, as we discussed in Discord I'm not sure this is something we need to change. My understanding of the invite-only feature is "you should be able to sign up only if you have received an invitation" and I think the current functionality achieves that. These changes would further tighten the restriction to being "you can only sign up by clicking directly from an invitation link", which I think may be a step too far.

gazayas:
@jagthedrummer No problem, I'm glad we sifted through it though, we can always come back to it if anything.

PR #622, opened 2023-10-19
Update to Masamune 2.0.2

gazayas:
Masamune now uses [Prism nodes](https://github.com/gazayas/masamune-ast/pull/60), so I feel much safer using this as an interface for other parts of Bullet Train.

Before getting to those other parts though, I wanted to update what we already have in place first. This PR takes care of that.


gazayas:
I saw that we have v1.2.1 in the starter repository as well, I'm assuming those will get updated when we update to the next bt core version though:

https://github.com/bullet-train-co/bullet_train/blob/43aa386dc07300f87658a04a368ee567bd020fbf/Gemfile.lock#L180-L189

PR #629, opened 2023-10-23
name keyword splats

abradburne:
ruby gives syntax error if you try and call a method with a kwsplat (rather than just accept args) - they need to be named

### DIFF
--- a/bullet_train-api/lib/bullet_train/api/example_bot.rb
+++ b/bullet_train-api/lib/bullet_train/api/example_bot.rb
@@ -4,13 +4,13 @@ module FactoryBot
   module ExampleBot
     attr_accessor :tables_to_reset
 
-    def example(model, **)
+    def example(model, **options)
       @tables_to_reset = [model.to_s.pluralize]
 
       object = nil
 
       ActiveRecord::Base.transaction do
-        instance = FactoryBot.create(factory(model), **)
+        instance = FactoryBot.create(factory(model), **options)
         object = deep_clone(instance)
 
         raise ActiveRecord::Rollback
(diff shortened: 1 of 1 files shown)


jagthedrummer:
I don't think I follow what you're saying @abradburne. Can you show some reproduction steps for the bug/error you're trying to fix?

abradburne:
You can't call a method using `**` as an argument (as attempted in https://github.com/bullet-train-co/bullet_train-core/pull/629/files#diff-48c35d5e0625ab7d90692433cffa38ee23b10c0d508569f04242f56d5d6bd29dL13)

Attempting to use the open api features in bullet train with this as it is results in the error:
```
SyntaxError: [...]/bullet_train-core/bullet_train-api/lib/bullet_train/api/example_bot.rb:13: syntax error, unexpected ')'
...yBot.create(factory(model), **)
...                              ^
[...]/bullet_train-core/bullet_train-api/lib/bullet_train/api/example_bot.rb:30: syntax error, unexpected ')'
...t(factory(model), quantity, **)
...                           
```

In order to pass 'the rest' of the args from calling this option on to the `FactoryBot.create` method, the keyword splat needs to be named (e.g. `**options`), not unnamed (`**`)

We can accept & ignore 'the rest' of options into the `def example(model, **)` method using the anonymous keyword splat, but we can't reference them when calling another method in `instance = FactoryBot.create(factory(model), **)`

This was changed in https://github.com/bullet-train-co/bullet_train-core/pull/300/commits/9784160b74dc2ee0b813fa77dddf93ef68301df8#diff-48c35d5e0625ab7d90692433cffa38ee23b10c0d508569f04242f56d5d6bd29d as 'fixing standard ruby', but doesn't appear to be any other reason for the change.

abradburne:
@jagthedrummer did this make sense?

olbrich:
This looks to be a new language feature in Ruby 3.2...  https://rubyreferences.github.io/rubychanges/3.2.html#anonymous-arguments-passing-improvements

jagthedrummer:
Sorry for the delay @abradburne, that does make sense. Thanks for the extra context, and for the fix!

abradburne:
@olbrich Ah, thanks for the info! 
Our project is still stuck pre 3.2 for the moment (wip!) which explains is, but good to know!

PR #653, opened 2023-11-09
Remove collaborators example

### LINKED ISSUE #648: Remove `Collaborators` example
It seems that we don't need the `Collaborators` example in `CreativeConcept` anymore:

Context
- https://discord.com/channels/836637622432170028/1131046208627032135/1131397933988843552
- https://discord.com/channels/836637622432170028/979662230792597534

gazayas:
Closes #648.

Joint PR
- https://github.com/bullet-train-co/bullet_train/pull/1142

We'll need to drop the table in the stater repository as well.
I'm seeing there are duplicate migrations, so maybe we can delete them from `bullet_train-core`, but I'm thinking this goes beyond the scope of Collaborators, so we should probably address it in another PR.

I think the remainder of the collaborators content in the `bullet_train-roles` tests and the documentation, so I'll have to sift through this one a little more before marking it as ready for review.


jagthedrummer:
This is my favorite kind of PR.

CleanShot 2023-11-13 at 09 29 31

PR #663, opened 2023-11-15
Add PADDLE_SECRET_KEY to billing_enabled? check

julianrubisch:
Preparing for Paddle Billing support

### DIFF
--- a/bullet_train/lib/bullet_train.rb
+++ b/bullet_train/lib/bullet_train.rb
@@ -81,7 +81,7 @@ def inbound_email_enabled?
 end
 
 def billing_enabled?
-  (ENV["STRIPE_SECRET_KEY"].present? || ENV["PADDLE_SECRET_KEY"]) && defined?(BulletTrain::Billing)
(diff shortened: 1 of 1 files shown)


bullet_train/lib/bullet_train.rb
```diff
@@ -81,7 +81,7 @@ def inbound_email_enabled?
 end
 
 def billing_enabled?
-  ENV["STRIPE_SECRET_KEY"].present? && defined?(BulletTrain::Billing)
+  (ENV["STRIPE_SECRET_KEY"].present? || ENV["PADDLE_SECRET_KEY"]) && defined?(BulletTrain::Billing)
```
jagthedrummer, line 84: Do we need `.present?` on the new var that we're checking for?

```suggestion
  (ENV["STRIPE_SECRET_KEY"].present? || ENV["PADDLE_SECRET_KEY"].present?) && defined?(BulletTrain::Billing)
```

julianrubisch: Thank you @jagthedrummer that was an oversight of mine. I think eventually we'll want to refactor this and delegate to the provider gems...

jagthedrummer: Agreed on figuring out how to delegate this to the various providers.

PR #667, opened 2023-11-15
[CI] Be more explicit about some dependencies

jagthedrummer:
In the course of [trying to be able to run all of the tests for individual gems in CI](https://github.com/bullet-train-co/bullet_train-core/pull/569) I've discovered that we don't specify all of our dependencies fully.

In the starter repo the problems don't appear because all of the things that we need end up being dependencies of something or other, but when we try to run tests for individual gems we'll get errors about unsatisfied dependencies.

There are only a few gems where we needed to declare a missing dependency in the `.gemspec`. And those are the only things in this PR that could have the chance to impact downsteam apps.

All of the changes to `Gemfile` and `Gemfile.lock` in each gem won't impact gem consumers at all. They just come into play in local development and CI. The changes in each `Gemfile` ensure that when developing or testing these gems that they'll use their peer gems from inside the `core` repo insted of using whatever version happens to be latest.

### DIFF
--- a/bullet_train-api/.gitignore
+++ b/bullet_train-api/.gitignore
@@ -8,7 +8,6 @@
 /test/dummy/log/*.log
 /test/dummy/storage/
 /test/dummy/tmp/
-Gemfile.lock
 yarn.lock
 /node_modules
 /.idea/
--- a/bullet_train-api/Gemfile
+++ b/bullet_train-api/Gemfile
@@ -8,8 +8,10 @@ gem "sqlite3"
 
 gem "sprockets-rails"
 
-gem "bullet_train"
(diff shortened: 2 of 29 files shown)


bullet_train-api/.gitignore
```diff
@@ -8,7 +8,6 @@
 /test/dummy/log/*.log
 /test/dummy/storage/
 /test/dummy/tmp/
-Gemfile.lock
```
jagthedrummer, line 11: [Bundler recommends including `Gemfile.lock` in the repo of gems.](https://bundler.io/guides/faq.html#using-gemfiles-inside-gems)

PR #672, opened 2023-11-16
Publish changed quantity whenever a membership is mutated

julianrubisch:
A minimal instrumentation to notify any subscribing payment provider gems of changed per-seat quantity.

### DIFF
--- a/bullet_train/app/models/concerns/memberships/base.rb
+++ b/bullet_train/app/models/concerns/memberships/base.rb
@@ -30,6 +30,10 @@ module Memberships::Base
       end
     end
 
+    after_create_commit :publish_changed_quantity
+    after_update_commit :publish_changed_quantity
+    after_destroy_commit :publish_changed_quantity
+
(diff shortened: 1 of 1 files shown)


bullet_train/app/models/concerns/memberships/base.rb
```diff
@@ -151,5 +155,9 @@ def remove_user_profile_photo
     user_profile_photo.purge
   end
 
+  def publish_changed_quantity
+    ActiveSupport::Notifications.instrument("quantity-changed", {team:})
```
jagthedrummer, line 159: The name of this event seems too generic to me. I think we should do something to indicate _which_ quantity has changed. Maybe something like:

```suggestion
    ActiveSupport::Notifications.instrument("memberships.quantity-changed", {team:})
```

jagthedrummer (changes requested):

bullet_train/app/models/concerns/memberships/base.rb
julianrubisch: Good point!

PR #675, opened 2023-11-16
[CI] Use local copies of the `bullet_train` gem when testing `-fields`

jagthedrummer:
This ensures that when we're running tests in CI for the `bullet_train-fields` gem that it's using the peer version of `bullet_train` instead of whichever released version happens to be in `Gemfile.lock`.

### DIFF
--- a/bullet_train-fields/Gemfile
+++ b/bullet_train-fields/Gemfile
@@ -8,6 +8,7 @@ gem "sqlite3"
 
 gem "sprockets-rails"
 
+gem "bullet_train", path: "../bullet_train"
 gem "bullet_train-api", path: "../bullet_train-api"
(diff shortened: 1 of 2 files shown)


bullet_train-fields/Gemfile
```diff
@@ -8,6 +8,7 @@ gem "sqlite3"
 
 gem "sprockets-rails"
 
+gem "bullet_train", path: "../bullet_train"
```
gazayas, line 11: I can see this being helpful for tests for individual core gems, I remember the whole paths thing being an issue :+1:

gazayas: That's right, it goes back to tbis: https://github.com/bullet-train-co/bullet_train-core/pull/550#discussion_r1325760879

PR #677, opened 2023-11-17
Don't use _commit callbacks when sending quantity updates

julianrubisch:
For whatever reason, I suddenly ran into issues - the transactional versions of the callbacks weren't triggered anymore when adding/removing a team member.

I think we should be fine with the ordinary versions though. Sorry for the hassle!

### DIFF
--- a/bullet_train/app/models/concerns/memberships/base.rb
+++ b/bullet_train/app/models/concerns/memberships/base.rb
@@ -30,9 +30,9 @@ module Memberships::Base
       end
     end
 
-    after_create_commit :publish_changed_quantity
-    after_update_commit :publish_changed_quantity
-    after_destroy_commit :publish_changed_quantity
+    after_create :publish_changed_quantity
+    after_update :publish_changed_quantity
+    after_destroy :publish_changed_quantity
 
     after_validation :remove_user_profile_photo, if: :user_profile_photo_removal?
 
--- a/bullet_train/app/models/concerns/memberships/base.rb
+++ b/bullet_train/app/models/concerns/memberships/base.rb
@@ -30,9 +30,7 @@ module Memberships::Base
       end
     end
 
-    after_create :publish_changed_quantity
-    after_update :publish_changed_quantity
-    after_destroy :publish_changed_quantity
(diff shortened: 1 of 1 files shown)


jagthedrummer:
@julianrubisch what's the scenario where they weren't firing? It didn't happen to be in a `test` env did it?

julianrubisch:
No, actually I didn't change anything since yesterday 🤔

bullet_train/app/models/concerns/memberships/base.rb
```diff
@@ -30,9 +30,9 @@ module Memberships::Base
       end
     end
 
-    after_create_commit :publish_changed_quantity
-    after_update_commit :publish_changed_quantity
-    after_destroy_commit :publish_changed_quantity
+    after_create :publish_changed_quantity
+    after_update :publish_changed_quantity
+    after_destroy :publish_changed_quantity
```
jagthedrummer, line 35: @julianrubisch this is really weird. I just ran some tests and the callback for `after_destroy_commit` is firing, but not the other two. I tried a generic `after_save_commit` and that didn't fire either. But just a plain `after_commit` _does_ fire. So I think we could replace all three callbacks with a single `after_commit` and we'd get what we want.

```suggestion
    after_commit :publish_changed_quantity
```

jagthedrummer: Looks like this is the problem: https://github.com/rails/rails/issues/29554

Each of the three `after_*_commit` callbacks are just aliases for `after_commit on: :*` so whichever is declared last wins. For some reason the Rails team doesn't see this as being a bug.

julianrubisch: Oh my, now I **remember**! Seems like every 5 years I have to stumble upon this ...

jagthedrummer: Yeah, I feel like I've run into this before as well, and never seem to remember. To me it seems like an _obvious_ problem that should be fixed in Rails itself. Not sure why the Rails teams doesn't think so.

PR #678, opened 2023-11-17
Use Prism nodes to find namespaces in RoutesFileManipulator

gazayas:
Trying piece by piece to migrate to Masamune, and I'm really liking that we can find block ends with Prism nodes. There's one model whose routes are not being properly scaffolded though, so I'll keep trying to debug this one.

I added a TODO, but I also noticed we use both strings and symbols in the `namespaces` array we're passing to this method, which is having an affect on the keys for the hash we're returning. We don't have to do it in this PR, but it might be beneficial to standardize on one or the other.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb
@@ -73,13 +73,20 @@ class Scaffolding::RoutesFileManipulator
     namespace_nodes.map do |node|
       # Get the first argument (a Symbol) without the colon from the list of arguments.
       name = node.arguments.child_nodes.first.unescaped
-      results[name] = node.line_number if namespaces.include?(name)
+
+      # TODO: For some reason we use both strings and symbols for namespaces.
+      # i.e. - ["account"] and [:v1].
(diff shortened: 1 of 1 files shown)


gazayas:
Okay, I had a hard time finding out what was wrong, but once I *did* find out, it reminded me of the `lines` setup that had bothered me in the past.

Basically, we're constantly mutating the `RoutesFileManipulator#lines` object, and I forget in what situation this was in, but I remember it being hard to keep track of the lines. Here are some of my thoughts from back in 2022: https://github.com/bullet-train-co/bullet_train-super_scaffolding/pull/32. Basically, if we explicitly pass `lines` to each method, we know that we have to be aware of what its current state is. Although we technically have to write more code by including `lines` each time, I think it will be easier to keep track of them.

With that being said, I'm MUCH happier with this setup. Before, we were getting two of the same strings in `namespaces` like `["projects", "projects"]` and we would pop each one. This was a little difficult to understand when prying inside because I figured there might be a reason for there being two of them (which I ultimately don't think there was). Now that we use Prism nodes, it's much more explicit as to what the intent and the flow of the code is. I'm hoping we can use Prism nodes more extensively in the code base.

jagthedrummer (approved):

PR #681, opened 2023-11-17
Fix: ActionText values for jbuilder-schema

### LINKED ISSUE #680: ActionText::RichText being sent unmodified to `jbuilder-schema`
We use [`ValuesTransformer`](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-api/lib/jbuilder/values_transformer.rb) for Jbuilder, but in schema values stay unchanged.

newstler:
Fixes #680 

@jagthedrummer Not sure where we should put this `require`, would love to hear an advice.


bullet_train-api/app/helpers/api/open_api_helper.rb
```diff
@@ -46,6 +46,10 @@ def automatic_components_for(model, locals: {})
       path = "app/views/api/#{@version}"
       paths = [path, "app/views"] + gem_paths.product(%W[/#{path} /app/views]).map(&:join)
 
+      # Transform values the same way we do for Jbuilder templates
+      require "jbuilder/schema/template"
```
jagthedrummer, line 50: @newstler, I think this `require` should be in the primary `schema.rb` file of the `jbuilder-schema` gem.

https://github.com/bullet-train-co/jbuilder-schema/blob/73a07d0d6c7f4628308f9f775ea4b9f486b8c38b/lib/jbuilder/schema.rb#L4

Unless there are reasons that people might want to use the gem but _not_ have the transformer immediately available for use.

bullet_train-api/bullet_train-api.gemspec
```diff
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "pagy_cursor"
   spec.add_dependency "rack-cors"
   spec.add_dependency "doorkeeper"
-  spec.add_dependency "jbuilder-schema", "2.6.2"
+  spec.add_dependency "jbuilder-schema", "2.6.4"
```
jagthedrummer, line 35: Now that we can [test `jbuilder-schema` against the starter repo](https://github.com/bullet-train-co/jbuilder-schema/pull/73) prior to releasing a new version I think we should try allowing the patch number to float again.
```suggestion
  spec.add_dependency "jbuilder-schema", "~> 2.6.4"
```

bullet_train-api/app/helpers/api/open_api_helper.rb
```diff
@@ -46,6 +46,10 @@ def automatic_components_for(model, locals: {})
...
+      Jbuilder::Schema::Template.prepend ValuesTransformer
```
jagthedrummer, line 51: I kinda think that this line should go in an initializer or something. I don't fully grok what it's doing, but I'm wondering if `jbuilder-schema` should just do this by default? (Apologies if that's a dumb question.)

jagthedrummer (changes requested):

newstler:
@jagthedrummer Seems we fixed the issue with Template requirement, if we get this PR merged we can check https://github.com/bullet-train-co/jbuilder-schema/pull/73 and if everything's ok merge it too! The changes you requested I don't know how to mark I applied them, sorry 😅
Also not sure why bundle setup is failing, as I added Gemfile.lock to the PR...

jagthedrummer:
@newstler, I think the failures are due to us not having all of our dependencies between the `core` gems defined as explicitly as we should. I'm going to be out for the next couple of days, but next week I'll work on getting those dependency issues resolved so that we can get this merged.

jagthedrummer:
Getting our entire dependency graph sorted out is going to be a bigish job. For now I just updated the `Gemfile.lock` in each of the `core` gems that needs it so we can merge this and get moving again. I'll try to tackle the dependency thing this week.

PR #684, opened 2023-11-20
Fix: Handle internationalized git in superscaffolding script

MatElGran:
Git commands can be internationalized and superscaffolding untracked file detection will fail because the labels don't match (eg, on my machine, git displays "Fichiers non suivis" instead of "Untracked files"), which causes the script to stop right after model generation.

I initially modified `git_status`and `has_untracked_files` to use the `--porcelain` option of `git status` which, according to the docs, is intended to be used by scripts and as such is not internationalized. but digging deeper, it seems these are not needed, as `get_untracked_files` already manages the "no untracked files" scenario by returning an empty array.

As a workaround until this or a better solution is merged, one can re-run the superscaffolding command with `--skip-migration-generation` and the script will successfully run this time

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/script.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/script.rb
@@ -50,16 +50,7 @@ def standard_protip
   puts "If you do that, you can reset to your last commit state by using `git checkout .` and `git clean -d -f` ."
 end
 
-def git_status
-  `git status`.split("\n")
-end
-
-def has_untracked_files?(status_lines)
-  status_lines.include?("Untracked files:")
-end
-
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
This seems OK to me. @gazayas, would appreciate your eyes on this since it's an area you know well.

gazayas:
This is a great fix, @MatElGran! I originally added in the check for "Untracked files:" because the git command in `get_untracked_files` wasn't there when I first submitted the PR, but now that we're grabbing the files that way I think this will suffice moving forward. Thanks!

> but digging deeper, it seems these are not needed, as get_untracked_files already manages the "no untracked files" scenario by returning an empty array.

Yes, I think if we're just returning this empty array we shouldn't need to add another check for these methods.

gazayas (approved):

PR #687, opened 2023-11-21
Fix docs explaining how to copy billing-usage migrations

julianrubisch:


bullet_train/docs/billing/usage.md
```diff
@@ -34,7 +34,7 @@ Use the following two commands on your shell to copy the required migrations int
 
 ```
 cp `bundle show --paths | grep bullet_train-billing | sort | head -n 1`/db/migrate/* db/migrate
-cp `bundle show --paths | grep bullet_train-billing-stripe | sort | head -n 1`/db/migrate/* db/migrate
+cp `bundle show --paths | grep bullet_train-billing-usage | sort | head -n 1`/db/migrate/* db/migrate
```
jagthedrummer, line 37: I don't think we want to remove the bit about copying migrations from the `billing-stripe` gem. People are going to need those if they're using Stripe.

jagthedrummer (changes requested):

bullet_train/docs/billing/usage.md
julianrubisch: Yeah, I think, though, that this needs restructuring in general. Right now, the parts about how billing works in general, the payment provider docs (I'm currently writing the Paddle docs) and how to use usage limiting are kind of all mixed together...

jagthedrummer: Oh, wait, I just realized that this is a `usage` specific page, and not a general `billing` page. Maybe removing `stripe` is OK since we mention getting those migrations on the Stripe page. 🤔

I definitely see what you mean about this being all mixed together.

jagthedrummer (approved):
All good, disregard previous comments.

bullet_train/docs/billing/usage.md
julianrubisch: Yep technically you can use usage billing without even a payment provider in place 🤷‍♂️

PR #697, opened 2023-11-23
Find namespaces in the `RoutesFileManipulator` with Prism nodes

### LINKED ISSUE #690: Route namespace resolution bug?
@gazayas I just rebased [the branch for running individual gem tests in GitHub CI](https://github.com/bullet-train-co/bullet_train-core/pull/569) and I started seeing a new failure there.

```
Failure:
Scaffolding::RoutesFileManipulatorTest#test_find_namespaces_at_the_root_level_returns_{"webhooks"=>46,_"incoming"=>47}_for_`webhooks`_under_`incoming` [/home/runner/work/bullet_train-core/bullet_train-core/bullet_train-super_scaffolding/test/lib/scaffolding/routes_file_manipulator_test.rb:50]:
--- expected
+++ actual
@@ -1 +1 @@
-{"webhooks"=>46, "incoming"=>47}
+{"webhooks"=>154, "incoming"=>47}


rails test test/lib/scaffolding/routes_file_manipulator_test.rb:48
```

It looks like it's related to https://github.com/bullet-train-co/bullet_train-core/pull/678

The test is failing in the same way locally for me on my branch, but if I revert that PR in my branch it starts passing again.

Unfortunately `bullet_train-super_scaffolding` is currently excluded from running the individual gem test in Circle, so that's how it snuck past us. Maybe it's a weird edge case since it's only triggering a failure in the gem tests, and not in any of the starter repo tests.

gazayas:
Closes #690.

The code looks simple now, but this was a really difficult one for me to figure out. One of the most important parts is how we shift elements off of the `namespaces` array to populate the `results` hash.

The new code is the same as the original in that it shifts the first element to `results`, and then leaves any leftover elements in `namespaces` if they don't match any namespaces in the routes file's lines.

I can't say I'm entirely sure *why* we need to shift off lines like this. I want to look over this file as a whole and apply Prism nodes where possible, so I hope to sit down with this file longer to get a better grasp of it.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb
@@ -54,17 +54,23 @@ class Scaffolding::RoutesFileManipulator
   def find_namespaces(namespaces, within = nil)
     namespaces = namespaces.dup
     results = {}
-    block_end = Scaffolding::BlockManipulator.find_block_end(starting_from: within, lines: lines) if within
-    lines.each_with_index do |line, line_number|
-      if within
-        next unless line_number > within
-        return results if line_number >= block_end
-      end
-      if line.include?("namespace :#{namespaces.first} do")
-        results[namespaces.shift] = line_number
-      end
-      return results unless namespaces.any?
+    reinstatiate_masamune_object
+
+    # `within` can refer to either a `resources`, `namespace`, `scope`, or `shallow` block.
+    blocks = @msmn.method_calls.select {|node| node.token_value.match?(/resources|namespace|scope|shallow/)}
+    namespace_nodes = blocks.select{|node| node.token_value.match?(/namespace/)}
+
+    if within
+      starting_block = blocks.find {|block| block.line_number - 1 == within}
+      block_range = (starting_block.location.start_line - 1)..(starting_block.location.end_line - 1)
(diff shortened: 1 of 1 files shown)


bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb
```diff
@@ -54,17 +54,23 @@ def divergent_parts
...
-      return results unless namespaces.any?
+    reinstantiate_masamune_object
+
+    # `within` can refer to either a `resources`, `namespace`, `scope`, or `shallow` block.
+    blocks = @msmn.method_calls.select { |node| node.token_value.match?(/resources|namespace|scope|shallow/) }
+    namespace_nodes = blocks.select { |node| node.token_value.match?(/namespace/) }
+
+    if within
+      starting_block = blocks.find { |block| block.line_number - 1 == within }
+      block_range = (starting_block.location.start_line - 1)..(starting_block.location.end_line - 1)
```
gazayas, line 65: I actually don't think we'll have to do `- 1` here because the block range is checking against `node.line_number`

jagthedrummer: Dang off-by-one errors! 🤘

jagthedrummer (approved):

PR #698, opened 2023-11-27
Resolve `roles.yml` file from `Rails.root`

### LINKED ISSUE #688: Better resolution for `config/models/roles.yml`
[In the process of trying to get tests running for individual gems](https://github.com/bullet-train-co/bullet_train-core/pull/569) I discovered that we try to resolve `config/models/roles.yml` relative to the current working directory. Which meant that I ended up putting what's essentially a test fixture at `bullet_train-incoming_webhooks/config/models/roles.yml` just to keep the tests happy.

Ideally we'd resolve that file relative to `Rails.root` (if we _have_ `Rails.root`) and the test file could then live in `bullet_train-incoming_webhooks/test/dummy/config/models/roles.yml`.

Here are at least a couple of spots where that resolution takes place, there may be more:

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-super_scaffolding/lib/scaffolding/transformer.rb#L486-L487

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-roles/lib/models/role.rb#L7-L8

It looks like there are a couple of spots in the `roles` gem where we tried to get the test suite to look at the right place, but those don't seem to work.

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-roles/test/dummy/config/application.rb#L34-L35

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-roles/test/dummy/config/initializers/bullet_train-roles.rb#L3-L6

gazayas:
Closes #688.

As a side note, I didn't add `#{Rails.root}` the following lines because it was throwing an error when trying to run migrations in the starter repository:

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-roles/lib/models/role.rb#L7-L8

https://github.com/bullet-train-co/bullet_train-core/blob/445b8e616199782d4de2de96b1e4e319656b08ff/bullet_train-roles/test/dummy/config/application.rb#L34-L35

`set_root_path` is an ActiveHash method, and it looks like it already looks for a `yml` file based off of the root path. This can be inferred from the following in the [documentation](https://github.com/active-hash/active_hash#activeyaml):

> By default, this class will look for a yml file named "countries.yml" in the same directory as the file. You can either change the directory it looks in, the filename it looks for, or both:
> ```ruby
> class Country < ActiveYaml::Base
>   set_root_path "/u/data"
>   set_filename "sample"
> end
> ```
> The above example will look for the file "/u/data/sample.yml".

### DIFF
--- a/bullet_train-roles/test/dummy/config/application.rb
+++ b/bullet_train-roles/test/dummy/config/application.rb
@@ -31,7 +31,7 @@ module Dummy
 
     config.factory_bot.definition_file_paths += [File.expand_path("../../factories", __dir__)] if defined?(FactoryBotRails)
 
-    Role.set_root_path("#{Rails.root}/config/models")
+    Role.set_root_path("config/models")
     Role.set_filename("roles")
   end
 end
--- a/bullet_train-roles/test/dummy/config/application.rb
+++ b/bullet_train-roles/test/dummy/config/application.rb
@@ -31,7 +31,7 @@ module Dummy
 
     config.factory_bot.definition_file_paths += [File.expand_path("../../factories", __dir__)] if defined?(FactoryBotRails)
 
-    Role.set_root_path("config/models")
+    Role.set_root_path("#{Rails.root}/config/models")
     Role.set_filename("roles")
   end
 end
--- a/bullet_train-roles/test/dummy/config/initializers/bullet_train-roles.rb
+++ b/bullet_train-roles/test/dummy/config/initializers/bullet_train-roles.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 Role.class_eval do
-  set_root_path "test/dummy/config/models"
+  set_root_path "#{Rails.root}/config/models"
   set_filename "roles"
 end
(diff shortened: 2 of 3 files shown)


gazayas:
Since we're currently out of CI credits I ran the super scaffolding tests locally and everything's working fine.

jagthedrummer:
Hmmm... this looks like it should work, but it doesn't seem to change anything. The tests are still passing with this change, but I'd expect them to fail since the fixture file would be in the wrong place. Just to confirm I pulled down this branch and tried moving the fixture file into the `dummy` app where it should go, and then the tests fail due to not being able to find the file.

```
Webhooks::Incoming::BulletTrainWebhooksControllerTest
  test_should_get_incoming_webhook                               ERROR (0.46s)
Minitest::UnexpectedError:         Errno::ENOENT: No such file or directory @ rb_sysopen - config/models/roles.yml
```

jagthedrummer:
@gazayas I took another look at this today and ended up opening https://github.com/bullet-train-co/bullet_train-core/pull/968 to take a slightly different approach. Could you give that one a look and let me know if it seems reasonable? (Basically it uses the engine in the `roles` gem to set the root path once, when rails initializes, instead of needing to add configs to each gem that needs a roles file for tests.)

jagthedrummer:
Closing in favor of https://github.com/bullet-train-co/bullet_train-core/pull/968

PR #700, opened 2023-11-28
[CI] Run starter repo tests in GitHub Actions

jagthedrummer:
This makes it so that we run the test suites from the starter repo in GitHub Actions.

CircleCI is giving us trouble (again!), so it may be time to ditch it for this repo.

CleanShot 2023-11-28 at 11 22 47

### DIFF
--- a/.github/workflows/_run_super_scaffolding_tests.yml
+++ b/.github/workflows/_run_super_scaffolding_tests.yml
@@ -46,7 +46,7 @@ jobs:
       BUNDLE_JOBS: 2
       BUNDLE_RETRY: 3
     steps:
-      - name: Checkout This Repo
+      - name: Checkout Core Repo
         uses: actions/checkout@v4
         with:
           path: tmp/core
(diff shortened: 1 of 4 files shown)


kaspth (approved):
Cool! This sounds good to me. GitHub Actions is as much magic incantations as CircleCI to me so I don't have opinions or advice on the code if need be.

.github/workflows/_run_super_scaffolding_tests.yml
```diff
@@ -0,0 +1,102 @@
...
+          bundler-cache: true
+
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: tmp/starter/.nvmrc
+          cache: 'yarn'
+          cache-dependency-path: tmp/starter/yarn.lock
+
+      - name: Link Core Repo
```
gazayas, line 73: Since we have `Checkout This Repo` on line 49, maybe we should change this to `Link This Repo` instead? Super minor nitpick, just thought I'd mention it.

gazayas (approved):

jagthedrummer:
Since CircleCI is automatically failing due to billing issues I disabled it on this repo in the same way as described here: https://github.com/bullet-train-co/bullet_train/pull/1206#issuecomment-1832187370

PR #701, opened 2023-11-28
[CI] Handle HIDE_THINGS differently in tests

jagthedrummer:
Joint PR: https://github.com/bullet-train-co/bullet_train/pull/1207

Since models are evaluated once at the beginning of a test run this conditional was making it impossible to toggle the value for `HIDE_THINGS` during testing. I think that removing the conditional should be safe since the `Scaffolding::AbsolutelyAbstract::CreativeConcept` model and table are always present.

### DIFF
--- a/bullet_train/app/models/concerns/teams/base.rb
+++ b/bullet_train/app/models/concerns/teams/base.rb
@@ -3,9 +3,9 @@ module Teams::Base
 
   included do
     # super scaffolding
-    #unless scaffolding_things_disabled?
-      has_many :scaffolding_absolutely_abstract_creative_concepts, class_name: "Scaffolding::AbsolutelyAbstract::CreativeConcept", dependent: :destroy, enable_cable_ready_updates: true
-    #end
+    # unless scaffolding_things_disabled?
+    has_many :scaffolding_absolutely_abstract_creative_concepts, class_name: "Scaffolding::AbsolutelyAbstract::CreativeConcept", dependent: :destroy, enable_cable_ready_updates: true
+    # end
 
     # added_by_id is a foreign_key to other Memberships on the same team,
     # so we nullify this to remove the constraint to delete the team.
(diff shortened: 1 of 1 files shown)


kaspth (approved):
I don't know what the implications are, or when users would want to disable scaffolding, but it's a hard dependency so this seems a little strange to disable it like this.

Alternative we could move the `has_many` setup in the super_scaffolding gem and use the `ActiveSupport.run_load_hooks :bullet_train_teams_base, self` load hook like this:

```ruby
ActiveSupport.on_load :bullet_train_teams_base do
  has_many :scaffolding_absolutely_abstract_creative_concepts, class_name: "Scaffolding::AbsolutelyAbstract::CreativeConcept", dependent: :destroy, enable_cable_ready_updates: true
end
```

Then we could gate the above behind `scaffolding_things_disabled?` earlier in the loading code, so we still don't need to run the tests twice.

jagthedrummer:
> I don't know what the implications are, or when users would want to disable scaffolding, but it's a hard dependency so this seems a little strange to disable it like this.

I don't follow, @kaspth. This PR isn't disabling anything, it's enabling a relationship at all times which had previously been only sometimes enabled because of the conditional.

I'm also not following how it would help us to move the `has_many`. If we moved it and also wrapped it in the conditional I think we'd have the same problem that I encountered before.

kaspth:
> I don't know what the implications are, or when users would want to disable scaffolding, but it's a hard dependency so this seems a little strange to disable it like this.

Ah sorry, I can see how this is ambiguous now. "this" was referring to `scaffolding_things_disabled?`, as I don't understand why people would want to disable that relationship? But that's beyond the scope of this PR.

Let's go for it here!

gazayas:
> I think that removing the conditional should be safe since the Scaffolding::AbsolutelyAbstract::CreativeConcept model and table are always present.

Agreed, I think it developers really didn't want it in their application they could eject the file and remove it, but I think it's minimal enough to just keep the association.

gazayas (approved):

PR #702, opened 2023-11-29
Remove unneeded migrations

gazayas:
These migrations can be found 1-for-1 in the starter repository, so I don't think we need them here in the core repository.

I did not delete migrations for `bullet_train-outgoing_webhooks` because we need those for the tests.

Besides that I almost skipped over `bullet_train-integrations-stripe` because I thought we would need the migrations, but I was thinking of `bullet_train-billing-stripe` (The following is from the [documentation](https://bullettrain.co/docs/billing/stripe#23-copy-database-migrations)).

> ## 2.3. Copy Database Migrations
> Use the following two commands on your shell to copy the required migrations into your local project:
> ```
> cp `bundle show --paths | grep bullet_train-billing | sort | head -n 1`/db/migrate/* db/migrate
> cp `bundle show --paths | grep bullet_train-billing-stripe | sort | head -n 1`/db/migrate/* db/migrate
> ```

Totally understandable if this is too much in one PR, just let me know if you want me to split it up or take another approach.


jagthedrummer (approved):
I didn't even realize these files where there. I think this seems safe.

@pascallaliberte @kaspth @andrewculver Do you guys have any concerns with this?

jagthedrummer:
🤔 Hmmm... that's a really good point about the migrations potentially being useful for non-starter-repo apps.

PR #709, opened 2023-11-30
Respect logo_height setting in workflow/box

### LINKED ISSUE #708: Fix use of `logo_height` theme property on workflow pages, make instructions clearer?
The `logo_height` property is available on the `BulletTrain::Themes` module.

Its value can be set in the `config/initializers/logo.rb` file, and the comments make it clear where the setting is used.

https://github.com/bullet-train-co/bullet_train/blob/4e8479901a21b14f8f1b888640becae49afd5075/config/initializers/logo.rb#L1-L6

However, on workflow pages (sign in, sign up, invitations, etc), the `h-12` class set on the `img` tag overrides the `width` and `height` properties for the logo. Changing `logo_height` has no effect on those pages.

Also, to make it clearer that you can also change the height of the logo in the account menu (and that by default it's not shown), maybe the comments can be updated with instructions on ejecting the menu/logo partial.

pascallaliberte:
Fixes #708 

Joint PR: https://github.com/bullet-train-co/bullet_train/pull/1217


jagthedrummer (approved):
Thanks for handling this, @pascallaliberte!

PR #710, opened 2023-12-01
Refine Integration

kaspth:
Integrating refine-rails into Bullet Train.

### DIFF
--- a/bullet_train-super_scaffolding/Gemfile
+++ b/bullet_train-super_scaffolding/Gemfile
@@ -11,5 +11,7 @@ gem "sprockets-rails"
 gem "bullet_train"
 gem "bullet_train-api"
 
+gem "refine-rails", github: "clickfunnels2/refine-rails"
+
 # Start debugger with binding.b [https://github.com/ruby/debug]
 # gem "debug", ">= 1.0.0"
--- a/bullet_train-super_scaffolding/app/models/scaffolding/completely_concrete/tangible_thing/assignment.rb
+++ b/bullet_train-super_scaffolding/app/models/scaffolding/completely_concrete/tangible_thing/assignment.rb
@@ -1,0 +1,24 @@
+class Scaffolding::CompletelyConcrete::TangibleThings::Assignment < ApplicationRecord
+  # 🚅 add concerns above.
+
+  # 🚅 add attribute accessors above.
+
+  belongs_to :tangible_thing, class_name: "Scaffolding::CompletelyConcrete::TangibleThing"
+  belongs_to :membership, class_name: "Membership"
+  # 🚅 add belongs_to associations above.
+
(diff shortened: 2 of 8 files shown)


bullet_train-super_scaffolding/app/models/scaffolding/completely_concrete/tangible_thing/assignment.rb
kaspth, line 1: This probably isn't something we should commit. I'll remove later.

bullet_train-super_scaffolding/Gemfile
```diff
@@ -11,5 +11,7 @@ gem "sprockets-rails"
 gem "bullet_train"
 gem "bullet_train-api"
 
+gem "refine-rails", github: "clickfunnels2/refine-rails"
```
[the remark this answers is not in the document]
kaspth: @jagthedrummer yeah, it's private for now, but it's in the process of being open-sourced, so it can't be in the `.gemspec` yet.  There's a version out that you can see via `gem install refine-rails`.

bullet_train-super_scaffolding/app/models/scaffolding/completely_concrete/tangible_thing/filter.rb
```diff
@@ -0,0 +1,36 @@
...
+      condition(:boolean_button_value, :option),
+      condition(:multiple_button_values, :option),
+      condition(:option_value, :option),
+      condition(:multiple_option_values, :option),
+      condition(:super_select_value, :option),
+      condition(:multiple_super_select_values, :option),
+
+      condition(:date_field_value, :date),
+      condition(:date_and_time_field_value, :datetime),
+      condition(:created_at, :datetime),
```
ssteffen, line 24: One thing I would recommend is we add in the clause manipulations for created_at and updated_at generators. The IS_SET and IS_NOT_SET clauses serve no purpose for these fields since they are always present so the scaffolding should probably automatically apply
`.without_clauses([DateWithTimeCondition::CLAUSE_SET, DateWithTimeCondition::CLAUSE_NOT_SET])` for simplicity

kaspth: @ssteffen good point! I just pushed 382ef94 which will use a column's `null: false` information to automatically reject the presence concerned clauses aka SET/NOT_SET.

I think I found a bug in Refine though, so I'll submit a PR shortly.

bullet_train-super_scaffolding/Gemfile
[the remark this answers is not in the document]
kaspth: Realized we could put it into the `.gemspec` since there's a gem version out.

PR #716, opened 2023-12-13
Don't use `prepend_before_action` when loading/authorizing resources

### LINKED ISSUE #531: Execution order issue with `load_and_authorize_resource`?
> We are having issues where account_load_and_authorize_resource is generating errors cause it is “skipping” authenticate_user!.
> By skipping what I really mean is that this method is injecting methods into the rails action chain BEFORE the authenticate_user! via prepend: true (other usages too).
> We execute code in before we authenticate a user cause `account_load_and_authorize_resource`   uses prepend actions

jagthedrummer:
Using `prepend_before_action` causes problems because developers lose the ability to control the order in which callbacks happen.

Maybe fixes #531? (Hard to be sure since I don't know how to reproduce whatever problem people are running into.)


jagthedrummer:
Closing in favor of #730

PR #717, opened 2023-12-14
Refactor loads and authorize resource part 1

kaspth:
* Makes some exception messages less wordy
* Simplifies the way we turn a single argument into an array
* Simplification of a conditional

From https://github.com/bullet-train-co/bullet_train-core/pull/525


jagthedrummer (approved):
Thanks for breaking these out, @kaspth! Much easier to see exactly what's going on.

jagthedrummer:
The CI failures are caused by an update to Chrome, which I fixed in the starter repo here: https://github.com/bullet-train-co/bullet_train/pull/1243

I'll go ahead and merge these if the only failure is that one.

PR #718, opened 2023-12-14
Refactor loads and authorizes resource part 2

kaspth:
From #525 

Hoist main options into keyword arguments instead of extracting them from the options hash.

Support both old and new call structures:

```
account_load_and_authorize_resource :project, through: :team # New
account_load_and_authorize_resource :project, :team # Old
```

The deprecate notice TODO was added over 2 years ago, but Bullet Train shifted away from deprecating the `through:` version. https://github.com/bullet-train-co/bullet_train-core/commit/c5ba33bc770ca69a2a05a84f656cd36ed8dac2d8


jagthedrummer:
Huh, after I merged https://github.com/bullet-train-co/bullet_train-core/pull/717 the base branch of this automatically changed to `main`, which seems good. But now this PR is showing that it wants to make changes that we already made in #717, which seems less good. It's almost like it thinks this PR is targeting a previous state of `main`.

PR #721, opened 2023-12-14
Refactor loads and authorize resource part 5

kaspth:
Replace `eval` with direct setting of variables and/or `instance_variable_set` as appropriate.

From #525

Also closes #525.

### DIFF
--- a/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
+++ b/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
@@ -106,14 +106,18 @@ module BulletTrain::LoadsAndAuthorizesResource
       end
 
       prepend_before_action only: member_actions do
-        possible_sources_of_parent = through_as_symbols.map { |tas| "@#{model}.#{tas}" }.join(" || ")
-        eval "#{instance_variable_name} ||= " + possible_sources_of_parent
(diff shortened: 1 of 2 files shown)


bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
```diff
@@ -94,35 +96,28 @@ def account_load_and_authorize_resource(model, positional_through = nil, through
...
       end
 
       prepend_before_action only: member_actions do
-        instance_variable_name = options[:polymorphic] || through_as_symbols[0]
-        possible_sources_of_parent =
-          through_as_symbols.map { |tas| "@#{model}.#{tas}" }.join(" || ")
-        eval_string =
-          "@#{instance_variable_name} ||= " + possible_sources_of_parent
-        eval eval_string
+        if (model_instance = instance_variable_get("@#{model}")) && !instance_variable_defined?(instance_variable_name)
```
jagthedrummer, line 109: @kaspth, was this intentionally an assignment inside of a conditional check? Or should this be `==`? Since we use `model_instance` below I'm guessing the assignment is intentional. If that's the case I think this would be nicer to read:

```suggestion
        model_instance = instance_variable_get("@#{model}")
        if model_instance && !instance_variable_defined?(instance_variable_name)
```

kaspth: It was! I just merged your suggestion in, since you're right it's clearer (I got a little confused trying to get what I was going for here at first).

PR #726, opened 2023-12-19
Allow local css overrides

pascallaliberte:
Resolves https://github.com/bullet-train-co/bullet_train/issues/567

Joint PR:
- (add PR here)

### DIFF
--- a/bullet_train-themes-light/app/assets/stylesheets/light.tailwind.css
+++ b/bullet_train-themes-light/app/assets/stylesheets/light.tailwind.css
@@ -2,4 +2,5 @@
 @import "tailwindcss/components";
 @import "tailwindcss/utilities";
 
-@import "light/application";
+@import "$THEMESTYLESHEETS/light/application.css"; /* alias used to support ejecting the file to local */
+


gazayas (commented):
I'm still looking over this one, but this is some really awesome stuff! I think we might have to edit some of our style-related rake tasks (like `bullet_train:themes:light:install`, etc.), but I think it would be really beneficial for developers to have local CSS overrides like this. Maybe we could even give them the option in `bin/configure` when they first start up an app to eject the main application style sheet. Either way, I like what's being done here!

jagthedrummer:
This is looking great @pascallaliberte!

Potentially naive question: Would there be any downsides to us "pre-ejecting" `app/assets/stylesheets/light.tailwind.css` so that it's already there and ready for customization? I'm thinking that it would be more intuitive for there to be something already there that can be used, instead of people needing to find the documentation. If there _are_ some big downsides to pre-ejecting, then maybe we can add a `README` in `app/assets/stylesheets` that point people to the relevant docs.

Either way this is going to remove a good deal of friction for people who want to use their own CSS.

pascallaliberte:
I've published a new PR #741 which supplants this one. Using the new postcss trick, I now recommend adding a default local stylesheet.

pascallaliberte:
Although this PR is tagged as merged, it's only because #741 (which was actually merged) was adding commits to this new line. We're good, see #741 for the real deal.

PR #730, opened 2024-01-02
Don't use `prepend_before_action` when loading/authorizing resources

### LINKED ISSUE #531: Execution order issue with `load_and_authorize_resource`?

jagthedrummer:
Using `prepend_before_action` causes problems because developers lose the ability to control the order in which callbacks happen.

Maybe fixes #531? (Hard to be sure since I don't know how to reproduce whatever problem people are running into.)

Supersedes #716

### DIFF
--- a/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
+++ b/bullet_train-super_load_and_authorize_resource/app/controllers/concerns/bullet_train/loads_and_authorizes_resource.rb
@@ -29,6 +29,9 @@ module BulletTrain::LoadsAndAuthorizesResource
     #
     # to help you understand the code below, usually `through` is `team`
     # and `model` is something like `project`.
+    #
+    # For instance, if the call to this method looks like this:
+    # account_load_and_authorize_resource :project, through: :team, through_association: :projects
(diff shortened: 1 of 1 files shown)


kaspth (approved):
This seems alright to me, though I don't know how to reason about whether there'd be other unforeseen repercussions to this.

I wonder if we can explain this further in the eventual release notes? Or can we support passing `prepend: true` to get the old behavior?

jagthedrummer:
Yeah, this definitely strikes me as the kind of change that could definitely have unforeseen consequences. Especially because I don't really understand whatever problem people were running into.

Maybe a `prepend` param, or a global config option _would_ be a good idea... 🤔

jagthedrummer:
> although it's a bit of do we want to carry this forward, versus just cutting it off and deal with the aftermath?

🤔 Interesting proposal on using an array. But I don't fully understand how it would work.

Honestly, I'm a little hesitant to merge either until we hear from someone who's experienced the problem that we're trying to fix. I still feel like I'm flying blind on this one.

jagthedrummer:
@seattlecyclist @aaricpittman Can y'all give this branch a try and see if it fixes the problem you mentioned in #531?

PR #731, opened 2024-01-04
`bin/resolve` for JavaScript and CSS files

### LINKED ISSUE #727: Allow resolving theme css and js through `bin/resolve`
Currently, the `bin/theme` script from the main repo, used primarily to help locate theme file paths for js and tailwindcss build steps, allows overriding theme files locally. It prefers a local path equivalent if found through this line:

https://github.com/bullet-train-co/bullet_train/blob/61891bd76dea6007cd1f5f0898ff68f4508f71de/bin/theme#L37

```rb
@candidate_paths = [Dir.pwd, gem_path].select(&:present?)
```

For example, if we'd like to eject `bullet_train-themes-light/app/assets/stylesheets/light.tailwind.css`, the main entry point to the theme's css, we currently can't eject it via `bin/resolve`.

```
% bin/resolve bullet_train-themes-light/app/assets/stylesheets/light.tailwind.css
We could not resolve the value you're looking for: bullet_train-themes-light/app/assets/stylesheets/light.tailwind.css

If you're looking for a partial, please try passing the partial string in either of the following two ways:
1. Without underscore and extention: bin/resolve shared/attributes/code
2. Literal path with package name: bin/resolve bullet_train-themes/app/views/themes/base/attributes/_code.html.erb

If you're looking for a locale, the key might not be implemented yet.
Try adding your own custom text for the key to your locale file and try again.
```

`bin/resolve` currently resolves:

* `file_path` => absolute file paths if given, although this is not so intuitive
* `class_path` and `locale_path`
* `partial_path` => for theme partials, detecting the top-most theme matching the path if given a `shared`-prefixed partial path

But no mechanism for the css and js files from a theme. It should cover resolving and ejecting these same file paths covered by `bin/theme`:

https://github.com/bullet-train-co/bullet_train/blob/61891bd76dea6007cd1f5f0898ff68f4508f71de/bin/theme#L24-L35

```rb
file_path, fallback_file_path = case file_type
when "tailwind-config"
  ["/tailwind.#{theme}.config.js", nil]
when "tailwind-mailer-config"
  ["/tailwind.mailer.#{theme}.config.js", "/tailwind.#{theme}.config.js"]
when "tailwind-stylesheet"
  ["/app/assets/stylesheets/#{theme}.tailwind.css", nil]
when "javascript"
  ["/app/javascript/application.#{theme}.js", nil]
else
  raise "Sorry, we couldn't recognize the file you're looking for: \"#{file_type}\""
end
```

gazayas:
Closes #727.

This one is a little tricky because these assets can exist in ejected themes, or in `bullet_train-themes-#{theme_name}` gems. I set things up so all the developer has to do is pass the file name. If we're okay with this, I can add some documentation and mark the PR as ready for review.

resolve js and css

### DIFF
--- a/bullet_train/lib/bullet_train/resolver.rb
+++ b/bullet_train/lib/bullet_train/resolver.rb
@@ -262,23 +262,13 @@ module BulletTrain
 
       # If the file name doesn't include the theme name (i.e. - light.tailwind.css),
       # we append the theme name to the path to match something like "app/assets/javascript/foo/".
-      append_theme_name = !file_name.match?("#{current_theme}.")
-      asset_globs = ["js", "css"].map do |extention|
-        path = "app/assets/"
-        if append_theme_name
-          path += case extention
-          when "js"
-            "javascript/"
-          when "css"
-            "stylesheets/"
-          end
-          path += "#{current_theme}/"
-        end
-        path + "**/*.#{extention}"
-      end
-
-      # Ensure we include JavaScript files that exist under the app's root directory.
-      asset_globs << "*.js"
+      asset_globs = [
+        "*.js", # Include JavaScript files under the app's root directory.
+        "app/assets/javascript/**/*.#{current_theme}.js",
+        "app/assets/javascript/#{current_theme}/**/*.js",
+        "app/assets/stylesheets/**/*.#{current_theme}.css",
+        "app/assets/stylesheets/#{current_theme}/**/*.css",
+      ]
 
       # Search for the file.
       files = Dir.glob(asset_globs).reject { |file| file.match?("/builds/") }
--- a/bullet_train/lib/bullet_train/resolver.rb
+++ b/bullet_train/lib/bullet_train/resolver.rb
@@ -281,14 +281,8 @@ module BulletTrain
         return nil unless gem_path
 
         # At this point we can be more generic since we're inside the gem.
-        asset_globs = ["**/*.js", "**/*.css"]
-
-        # Read the globs from the home directory.
-        Dir.chdir
-        asset_globs.map! { |glob| "#{gem_path}/#{glob}".gsub("/#{Dir.home}", "") }
-        files = Dir.glob(asset_globs)
-
-        absolute_file_path = files.find { |file| file.match?(/#{file_name}$/) }
+        files = Dir.glob(["#{gem_path}/**/*.js", "#{gem_path}/**/*.css"])
+        absolute_file_path = files.find { |file| file.end_with?(file_name) }
       end
 
       absolute_file_path
--- a/bullet_train/lib/bullet_train/resolver.rb
+++ b/bullet_train/lib/bullet_train/resolver.rb
@@ -113,7 +113,7 @@ module BulletTrain
         package_name: nil,
       }
 
-      result[:absolute_path] = file_path || class_path || locale_path || partial_path
+      result[:absolute_path] = file_path || class_path || locale_path || js_or_stylesheet_path || partial_path
 
       # If we get the partial resolver template itself, that means we couldn't find the file.
       if result[:absolute_path].match?("app/views/bullet_train/partial_resolver.html.erb") || result[:absolute_path].nil?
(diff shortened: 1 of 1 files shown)


bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,71 @@ def locale_path
...
+    end
+
     def ejected_theme?
-      current_theme_symbol = File.read("#{Rails.root}/app/helpers/application_helper.rb").split("\n").find { |str| str.match?(/\s+:.*/) }
-      current_theme = current_theme_symbol.delete(":").strip
       current_theme != "light" && Dir.exist?("#{Rails.root}/app/assets/stylesheets/#{current_theme}")
     end
+
+    def current_theme
+      msmn = Masamune::AbstractSyntaxTree.new(File.read("#{Rails.root}/app/helpers/application_helper.rb"))
```
kaspth, line 316: `Rails.root` is a `Pathname` so we can use `join` + `read` on it instead:

```suggestion
      msmn = Masamune::AbstractSyntaxTree.new(Rails.root.join("app/helpers/application_helper.rb").read)
```

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,71 @@ def locale_path
...
+          when "css"
+            "stylesheets/"
+          end
+          path += "#{current_theme}/"
+        end
+        path + "**/*.#{extention}"
+      end
+
+      # Ensure we include JavaScript files that exist under the app's root directory.
+      asset_globs << "*.js"
```
kaspth, line 281: I find this code pretty complex to read, could we replace it with something like this? (If I haven't gotten it right, I'd say that's another indicator that it's too complex and we should simplify). 

```suggestion
      asset_globs = [
        "*.js", # Include JavaScript files under the app's root directory.
        "app/assets/javascript/**/*.#{current_theme}.js",
        "app/assets/javascript/#{current_theme}/**/*.js",
        "app/assets/stylesheets/**/*.#{current_theme}.css",
        "app/assets/stylesheets/#{current_theme}/**/*.css",
      ]
```

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,71 @@ def locale_path
...
+
+      if absolute_file_path.nil? # then search for it in its respective gem.
+        gem_path = `bundle show bullet_train-themes-#{current_theme}`.chomp
+
+        # Fall back to `light` theme if no gem is available.
+        # (The developer might just be trying to find stylesheets that exist in `light`.)
+        if gem_path.empty?
+          gem_path = `bundle show bullet_train-themes-light`.chomp
+          return nil if gem_path.empty?
+        end
```
kaspth, line 295: I wonder if we should condense and unify these branches? I don't think the two \`\` calls would be that expensive if the latter needn't have run.

```suggestion
        # Check the current them but fall back to `light` theme if no gem is available.
        # (The developer might just be trying to find stylesheets that exist in `light`.)
        gem_path = [`bundle show bullet_train-themes-#{current_theme}`, `bundle show bullet_train-themes-light`].map(&:chomp).find(&:present?)
        return nil unless gem_path
```

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,71 @@ def locale_path
...
+
+        # At this point we can be more generic since we're inside the gem.
+        asset_globs = ["**/*.js", "**/*.css"]
+
+        # Read the globs from the home directory.
+        Dir.chdir
+        asset_globs.map! { |glob| "#{gem_path}/#{glob}".gsub("/#{Dir.home}", "") }
+        files = Dir.glob(asset_globs)
+
+        absolute_file_path = files.find { |file| file.match?(/#{file_name}$/) }
```
kaspth, line 305: Do we need `Dir.chdir`? Won't it be enough since we're passing an absolute file path? I wonder if we could do this instead:

```suggestion
        # At this point we can be more generic since we're reading from the gem.
        files = Dir.glob(["#{gem_path}/**/*.js", "#{gem_path}/**/*.css"])
        absolute_file_path = files.find { |file| file.match?(/#{file_name}$/) }
```

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,71 @@ def locale_path
...
+        end
+        path + "**/*.#{extention}"
+      end
+
+      # Ensure we include JavaScript files that exist under the app's root directory.
+      asset_globs << "*.js"
+
+      # Search for the file.
+      files = Dir.glob(asset_globs).reject { |file| file.match?("/builds/") }
+      absolute_file_path = files.find { |file| file.match?(/#{file_name}$/) }
```
kaspth, line 285: We can simplify this with a bit with `end_with?`:

```suggestion
      absolute_file_path = files.find { |file| file.end_with?(file_name) }
```

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,71 @@ def locale_path
...
+        # Read the globs from the home directory.
+        Dir.chdir
+        asset_globs.map! { |glob| "#{gem_path}/#{glob}".gsub("/#{Dir.home}", "") }
+        files = Dir.glob(asset_globs)
+
+        absolute_file_path = files.find { |file| file.match?(/#{file_name}$/) }
+      end
+
+      absolute_file_path
+    end
```
kaspth, line 309: It feels a bit strange to assign `absolute_file_path` and then reassign it much later and makes the method harder to trace. What if we hoisted it up instead, like this?

```suggestion
      absolute_file_path = files.find { |file| file.end_with?(file_name) }
      if absolute_file_path
        absolute_file_path
      else
        gem_path = `bundle show bullet_train-themes-#{current_theme}`.chomp

        # Fall back to `light` theme if no gem is available.
        # (The developer might just be trying to find stylesheets that exist in `light`.)
        if gem_path.empty?
          gem_path = `bundle show bullet_train-themes-light`.chomp
          return nil if gem_path.empty?
        end

        # At this point we can be more generic since we're inside the gem.
        asset_globs = ["**/*.js", "**/*.css"]

        # Read the globs from the home directory.
        Dir.chdir
        asset_globs.map! { |glob| "#{gem_path}/#{glob}".gsub("/#{Dir.home}", "") }
        files = Dir.glob(asset_globs)

        files.find { |file| file.end_with?(file_name) }
      end
```

kaspth (approved):
I wanted to leave some suggestions here, because I found this code a bit hard to grok. But other than that, it seems good!

bullet_train/lib/bullet_train/resolver.rb
gazayas: I forgot to use `Pathname` from back when we used it in https://github.com/bullet-train-co/bullet_train-core/pull/312#issuecomment-1708562523, I'll be sure to keep it in mind moving forward!

bullet_train/lib/bullet_train/resolver.rb
gazayas: Yup, this is much simpler, too much overthinking on my part.

gazayas:
@kaspth Thank you for looking over this one! I could've written this a lot better initially, sorry it was a jumbled one to read through. I went ahead and applied the changes, and everything is still working for me! 🙌

bullet_train/lib/bullet_train/resolver.rb
```diff
@@ -251,10 +251,47 @@ def locale_path
...
+      asset_globs = [
+        "*.js", # Include JavaScript files under the app's root directory.
+        "app/assets/javascript/**/*.#{current_theme}.js",
+        "app/assets/javascript/#{current_theme}/**/*.js",
+        "app/assets/stylesheets/**/*.#{current_theme}.css",
+        "app/assets/stylesheets/#{current_theme}/**/*.css",
+      ]
+
+      files = Dir.glob(asset_globs).reject { |file| file.match?("/builds/") }
+      absolute_file_path = files.find { |file| file.end_with?(file_name) }
```
kaspth, line 272: @gazayas I went ahead and applied the change to `end_with?` so we're symmetrical with the `end_with?` call on line 283.

kaspth:
@gazayas hey, all good! I'm really glad you wrote this, I think it's a good change 😊🙌

kaspth (approved):

jagthedrummer (approved):
Nice work, @gazayas!

PR #733, opened 2024-01-08
Introduce separate translations for API documentation

newstler:
This PR introduces separate translations for API: `api_title` and `api_description`. Those are to be used in OpenAPI documentation without affection of the UI.

- Adds those fields to Super Scaffolding;
- Updates translations in core gems;
- Adds rake task `bullet_train:api:create_translations` which updates existing locale files, and creates backup translation files in `tmp/backup_locales` if the original file was updated.

### DIFF
--- a/bullet_train-api/lib/tasks/bullet_train/api_tasks.rake
+++ b/bullet_train-api/lib/tasks/bullet_train/api_tasks.rake
@@ -152,11 +152,14 @@ namespace :bullet_train do
       # Define the root of the locales directory
       locales_root = Rails.root.join('config', 'locales').to_s
 
-      # Define the temporary directory
-      tmp_dir = Rails.root.join('tmp', 'locales')
-      FileUtils.mkdir_p(tmp_dir) unless Dir.exist?(tmp_dir)
+      # Define the backup directory
+      backup_dir = Rails.root.join('tmp', 'locales_backup')
+      FileUtils.mkdir_p(backup_dir) unless Dir.exist?(backup_dir)
 
-      Dir.glob("#{locales_root}/**/*.yml").each do |file|
+      original_files = Dir.glob("#{locales_root}/**/*.yml")
+      differing_files_count = 0
+
+      original_files.each do |file|
         puts "Processing #{file}..."
 
         lines = File.readlines(file)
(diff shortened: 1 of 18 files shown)


bullet_train-api/lib/tasks/bullet_train/api_tasks.rake
```diff
@@ -146,5 +146,103 @@ namespace :bullet_train do
         )
       end
     end
+
+    desc "Create `api_title` and `api_description` translations"
+    task create_translations: :environment do
```
jagthedrummer, line 151: @newstler Is this a task that's intended to be used by end users/developers? If so, we should probably document it somewhere. And maybe mention in the PR description that people should run it?

jagthedrummer (approved):
This looks good to me. Just not sure about the included rake task, and whether we need to document it.

bullet_train-api/lib/tasks/bullet_train/api_tasks.rake
jagthedrummer: I don't exactly know. Why and when would people need to use it? Maybe that can help us figure out where to surface it.

newstler: They only need to run it once, to update all the old translations they have. As there is no documentation for the bt-api gem, probably there should be a post-install massage? I have some thoughts about bt-api README, but right now seems post-install message should work just fine, I'll try to add it.

newstler: To be honest I would even run this task automatically after gem install, do you think this is an acceptable approach?

jagthedrummer: I'm not a _huge_ fan of the idea of running it automatically. _Maybe_ if it's extremely fast to run. But otherwise it's going to be taking time every time someone runs `bundle install` whether locally or in CI/CD.

I think a post-install message would be better than automatically running.

newstler: Added post-install message in `bullet_train-api/bullet_train-api.gemspec`.

I will try to create README for bt-api soon in a separate PR, and will mention this (and other) rake tasks there.

PR #736, opened 2024-01-09
Add Action Model generators

### LINKED ISSUE #735: Action Models scaffolding fails the valid attribute type check
We provide this example in the Action Models documentation.

```
bin/super-scaffold crud Project Team name:text_field
bin/super-scaffold action-model:targets-many Archive Project Team
```

There are two problems with this.
1. `bin/super-scaffold` is deprecated and we should update the documentation to use `rails generate super_scaffold` instead (I overlooked this in #699).
2. Running the second command fails the valid attribute type check.

```
> bin/super-scaffold action-model:targets-many Archive Project Team
DEPRECATION WARNING: `bin/super-scaffold` is deprecated. Please use `rails generate super_scaffold` instead.
You can see which options are available in the Super Scaffolding documentation: https://bullettrain.co/docs/super-scaffolding 
/home/gazayas/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bullet_train-super_scaffolding-1.6.24/lib/scaffolding/script.rb:103:in `block in check_required_options_for_attributes': You have entered an invalid attribute type: . General data types are used when creating new models, but Bullet Train uses field partials when Super Scaffolding, i.e. - `name:text_field` as opposed to `name:string`. Please refer to the Field Partial documentation to view which attribute types are available. (RuntimeError)
```

gazayas:
Closes #735.

This initial commit is just for targets-many. I'm leaving it as a draft for now in case we want to close this and move the logic to the Action Models repository. I'm okay with keeping it here though because we include some files here like the `SuperScaffoldBase` generator module which we don't have available in Action Models. We also don't have `bullet_train-super-scaffolding` as a dependency over there, so although I'm sure it's possible to move everything over, I'm not sure if it's worth it for the scope of this PR.

If we're okay with putting everything here, I'll go ahead and add the generators for the other types of actions.

As a side note, we'll also have to update https://github.com/bullet-train-co/bullet_train/blob/main/bin/super-scaffold.

### DIFF
--- a/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding.rb
+++ b/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding.rb
@@ -17,7 +17,8 @@ module BulletTrain
       "crud" => "BulletTrain::SuperScaffolding::Scaffolders::CrudScaffolder",
       "crud-field" => "BulletTrain::SuperScaffolding::Scaffolders::CrudFieldScaffolder",
       "join-model" => "BulletTrain::SuperScaffolding::Scaffolders::JoinModelScaffolder",
-      "oauth-provider" => "BulletTrain::SuperScaffolding::Scaffolders::OauthProviderScaffolder"
+      "oauth-provider" => "BulletTrain::SuperScaffolding::Scaffolders::OauthProviderScaffolder",
+      "action-models:targets-many" => "BulletTrain::ActionModels::Scaffolders::TargetsManyScaffolder"
     }
 
     class Runner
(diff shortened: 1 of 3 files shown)


gazayas:
Will mark as ready for review since I added some other scaffolders. Will work on `PerformsImport` and `PerformsExport` in the meantime, but the `targets-*` actions are all that we have currently documented.

gazayas:
Did some work on `PerformsImport` and `PerformsExport`, and I don't think we'll need the scaffolders for the time being for a few reasons.

1. They're not documented.
2. `PerformsImport` specifically acts very much like `TargetsOneParent` in that we're importing CSV data to database records, so I'm curious if the `PerformsImport` was eventually superseded by the latter during its development.
3. `PerformsExport` doesn't seem to be functioning out of the box, so I'm assuming it's not as high-priority as the other generators I added here.

jagthedrummer (approved):
Looks good to me, @gazayas! I agree that some of this seems like it should go in the Action Models repo, though some of it (the docs specifically) will always have to live in this repo. 🤔 I think for now it's OK to put them here, and then we can look at moving them in the future if there's a need to do so.

gazayas:
@jagthedrummer Sounds good 👍

PR #738, opened 2024-01-09
Feature: Allow custom properties for components

newstler:
This PR allows custom properties for components in Open-API schema:

```ruby
<%= automatic_components_for Team, attributes: {add: {password: {type: :string, description: "Test", required: true, example: "password-ABC"}}, remove: :time_zone}, parameters: {add: {password: {type: :string, description: "Test", required: true, example: "password-ABC"}}, remove: :time_zone} %>
<%= automatic_components_for User, attributes: {remove: [:email, :time_zone]}, parameters: {remove: [:email, :time_zone, :locale]} %>
<%= automatic_components_for Membership, attributes: {remove: :team_id}, parameters: {remove: :user_first_name} %>
<%= automatic_components_for Webhooks::Outgoing::Endpoint, attributes: {only: [:id, :team_id]}, parameters: {only: :url} %>
```

Acceptable keys: `add`, `remove` and `only` inside `attributes` and/or `parameters`

### DIFF
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -176,8 +176,8 @@ module Api
     def customize_component!(original, custom)
       custom = custom.deep_stringify_keys.deep_transform_values { |v| v.is_a?(Symbol) ? v.to_s : v }
 
-      if custom.key?("add")
-        custom["add"].each do |property, details|
+      if custom.key?("include")
+        custom["include"].each do |property, details|
           if details["required"]
             original["required"] << property
             details.delete("required")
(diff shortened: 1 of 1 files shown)


jagthedrummer (changes requested):
@newstler, should we document this somewhere so that people will know it's available? And are there some easy tests we can write for it?

newstler:
@jagthedrummer Here's the problem: bt-api gem doesn't have neither documentation nor tests :)
I'll try to add some simple tests for this exact feature now, but assuming this is quite a huge separate task to write tests for the whole gem.

jagthedrummer:
OK, yeah, if there aren't any kind of tests at all and we have to bootstrap the entire test suite then it's probably not for this PR. And docs would help a lot. I've honestly disabled all of the API stuff that I can because I don't know how it works, or even really how it's _supposed_ to work. Would be really nice to have some orientation docs for it.

PR #741, opened 2024-01-12
Document and support new default local stylesheet, ejecting theme style sheet

pascallaliberte:
Resolves https://github.com/bullet-train-co/bullet_train/issues/567.
Replaces #726.

* Adds a new `Style Sheets` documentation, with instructions on modifying theme partials or adding custom CSS (see joint PR for new default local style sheets)
* Introduces the command `rake bullet_train:themes:light:eject_css` for further modifying the theme style sheet
* Allows the theme style sheet to be ejected and have its `@import` rule still resolve. This uses a special postcss-import resolver and a `$ThemeStylesheetsDir` alias, defined in `postcss-import-config`, which the main repo's postcss.config.js (and mailer equivalent) imports.

This allows us to keep using the `bin/theme` script to pull all of the components of a theme (js, css and erb partials) and thereby allowing us to eject a theme, have ejected local versions of theme files, and define which theme to use for the app even if multiple theme gems are installed.

Joint PR:
- https://github.com/bullet-train-co/bullet_train/pull/1274


bullet_train-themes-light/postcss-import-config.js
```diff
@@ -0,0 +1,17 @@
+const path = require('path');
+const { execSync } = require("child_process");
+
+const themeStylesheet = execSync(`bundle exec bin/theme tailwind-stylesheet ${process.env.THEME} 2> /dev/null`).toString().trim()
+const themeRoot = execSync(`bundle show bullet_train-themes-${process.env.THEME} 2> /dev/null`).toString().trim()
+const themeStylesheetsDir = path.resolve(themeRoot, 'app/assets/stylesheets/');
```
gazayas, line 6: What do you think if we do this?

```js
const gemList = execSync(`bundle list --name-only 2> /dev/null`).toString().trim()
const gemName = gemList.split("\n").find((gem) => gem == `bullet_train-themes-${process.env.THEME}`)
let themeRoot = ""
if(gemName != undefined){
  const themeRoot = execSync(`bundle show bullet_train-themes-${process.env.THEME} 2> /dev/null`).toString().trim()
}
const themeStylesheetsDir = path.resolve(themeRoot, 'app/assets/stylesheets/');
```

I found that the child process would quit `bin/dev` if `bundle show` didn't find what we were looking for, so using `bundle list` seems like a good way to find the information we want.

bullet_train-themes-light/lib/tasks/application.rb
```diff
@@ -3,6 +3,14 @@
 module BulletTrain
   module Themes
     module Application
+      def self.eject_theme_main_css(theme_name)
+        theme_base_path = `bundle show --paths bullet_train-themes-#{theme_name}`.chomp
+
+        puts "Ejecting app/assets/stylesheets/#{theme_name}.tailwind.css."
+        `mkdir -p #{Rails.root}/app/assets/stylesheets`
```
kaspth, line 10: If we use `Rails.root.join`, we get back a `Pathname` that I think we can call `mkpath` on and get the same effect:

```suggestion
        Rails.root.join("app/assets/stylesheets").mkpath
```

bullet_train/app/views/layouts/docs.html.erb
```diff
@@ -111,6 +111,12 @@
                   <% end %>
                 <% end %>
 
+                <%= render 'account/shared/menu/item', url: '/docs/stylesheets', label: 'Style Sheets' do |p| %>
+                  <% p.icon do %>
+                  <i class="fa-brands fa-js ti ti-brush"></i>
```
kaspth, line 116: Not sure why the `<i>` just above and below here are misindented, but the other ones in the file are alright. How about we indent like the rest of them?

```suggestion
                    <i class="fa-brands fa-js ti ti-brush"></i>
```

bullet_train/docs/stylesheets.md
```diff
@@ -0,0 +1,18 @@
+# Style Sheets
+Bullet Train's stock UI theme, called “Light”, is built to use `tailwindcss` extensively.
```
kaspth, line 2: Should we link to https://github.com/bullet-train-co/bullet_train-core/tree/main/bullet_train-themes-light and https://github.com/bullet-train-co/bullet_train-core/tree/main/bullet_train-themes-tailwind_css and give a note about how the Light theme is built on the tailwindcss theme?

bullet_train/docs/stylesheets.md
```diff
@@ -0,0 +1,18 @@
...
+
+As such, there are two ways to update the styling of your app: either by modifying theme partials, or by adding your own custom CSS.
+
+## Modify Theme Partials
+
+Since `tailwindcss` is used, most style changes are done by ejecting theme partials into a local copy, and modifying the Tailwind classes within their HTML.
```
kaspth, line 8: I think we can clarify this a little bit.

```suggestion
Since `tailwindcss` is used, most style changes are done by ejecting theme partials into your app's `app/views` directory, and modifying the Tailwind classes within the `.html.erb` templates.
```

bullet_train/docs/stylesheets.md
```diff
@@ -0,0 +1,18 @@
...
+
+You can eject only the theme files you wish to override or you can eject the whole UI theme for customization. You can find more information in the [indirection documentation](indirection) about using `bin/resolve` to find the theme partials to eject. Or see the [themes documentation](themes) for details on using the "Light" UI theme as a starting point for creating your own.
+
+## Add custom CSS 
+
+If you need to add your own custom CSS, override the custom non-tailwind classes found in the `light` theme (custom classes are still used sparingly), or to import CSS from a third-party component, simply add to the `app/assets/stylesheets/application.custom.css` file.
```
kaspth, line 14: I think we can clean this paragraph a little bit, but to be honest I don't know quite what it's saying.

```suggestion
To add your own custom CSS, override the custom non-tailwind classes found in the `light` theme (custom classes are still used sparingly), or to import CSS from a third-party component, add to the `app/assets/stylesheets/application.custom.css` file.
```

bullet_train-themes-light/postcss-import-config.js
pascallaliberte: @gazayas Let me know what you think of this alternative I pushed: I've added `bin/theme stylesheets-dir {THEME}` to return the stylesheets path for the gem. In the case of the light theme, it would return `{gem_path}/app/assets/stylesheets/light/`

https://github.com/bullet-train-co/bullet_train-core/blob/aff1b3f7996700f5d03fd8a0610110c2647b2c10/bullet_train-themes-light/postcss-import-config.js#L3-L4

And so `light.tailwind.css` will just need to import:

https://github.com/bullet-train-co/bullet_train-core/blob/aff1b3f7996700f5d03fd8a0610110c2647b2c10/bullet_train-themes-light/app/assets/stylesheets/light.tailwind.css#L5

bullet_train/docs/stylesheets.md
pascallaliberte: I've clarified the opening paragraph a bit, but I wouldn't point to the gems here, or away to the Themes doc until the question "but how do I modify these files" pops up as they read the Modify Theme Partials section just below, which points to the Theme doc.

https://github.com/bullet-train-co/bullet_train-core/blob/c8ece1807012351f99980f785f12729f67d772a0/bullet_train/docs/stylesheets.md?plain=1#L1-L2

bullet_train/docs/stylesheets.md
pascallaliberte: Clarified the bottom paragraph also:

https://github.com/bullet-train-co/bullet_train-core/blob/6b60e20d709e657f4c8cb50d276aecd2c6c7efc7/bullet_train/docs/stylesheets.md?plain=1#L12-L14

bullet_train-themes-light/postcss-import-config.js
gazayas: @pascallaliberte LGTM! I confirmed that things are working properly against the default app, and when a custom theme is ejected.

bullet_train/docs/stylesheets.md
kaspth: Perfect!

bullet_train-themes-light/lib/tasks/application.rb
```diff
@@ -19,14 +27,17 @@ def self.eject_theme(theme_name, ejected_theme_name)
         %x(sed -i #{'""' if `echo $OSTYPE`.include?("darwin")} "s/#{theme_name}/#{ejected_theme_name}/g" #{Rails.root}/tailwind.mailer.#{ejected_theme_name}.config.js)
 
         puts "Ejecting stylesheets into `./app/assets/stylesheets/#{ejected_theme_name}`."
-        `mkdir #{Rails.root}/app/assets/stylesheets`
+        `mkdir -p #{Rails.root}/app/assets/stylesheets`
```
kaspth, line 30: We can do the `mkpath` trick here too:

```suggestion
        Rails.root.join("app/assets/stylesheets").mkpath
```

kaspth (approved):

jagthedrummer (approved):
Thanks for this, @pascallaliberte! This is going to remove one of the most common hurdles (IMO) that new BT devs face. 🚀

gazayas:
> This is going to remove one of the most common hurdles (IMO) that new BT devs face. 🚀

Agreed. Great work, @pascallaliberte!

kaspth:
Concurred! Really cool 🚀

PR #742, opened 2024-01-15
Feature: pass html_option for using user time_zone or browser time_zone in onboarding

cnorm35:
The current User detail onboarding form is passing in `@user.time_zone` as the selected value but the script at the bottom is overriding that value with the result of `jstz.determine().name()`

I think for most cases, that's a good default but with overriding that value, it doesn't allow for something like setting the User's time_zone from OAuth response data.

I'm guessing there's a reason for doing this within the script block on the page so I didn't try to move that to a stimulus controller, but if that's not the case, I'm happy to try to refactor to a stimulus controller.

### DIFF
--- a/bullet_train/app/views/account/onboarding/user_details/edit.html.erb
+++ b/bullet_train/app/views/account/onboarding/user_details/edit.html.erb
@@ -30,7 +30,8 @@
           <div class="sm:col-span-2">
             <%= render 'shared/fields/super_select', form: f, method: :time_zone,
               choices: time_zone_options_for_select(@user.time_zone, nil, ActiveSupport::TimeZone),
-              other_options: {search: true, required: true} %>
+              other_options: {search: true, required: true},
+              html_options: {use_browser_time_zone: true} %>
           </div>
         </div>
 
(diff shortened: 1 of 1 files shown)


gazayas:
Since this isn't changing how we set the time zone by default I think we could merge this if we document its use in the [Super Select documentation](https://github.com/bullet-train-co/bullet_train-core/blob/75bdbef029ee24312676c39d3a32e3b854a6376b/bullet_train/docs/field-partials/super-select.md?plain=1#L53). I think developers could benefit from this increased flexibility, especially if it's something in `bullet_train-core` which developers don't see on the surface, but could learn more about by ejecting the file and reading the documentation if they wanted to.

If we go the Stimulus controller route, we could potentially move `use_browser_with_time_zone` to `other_options` as opposed to `html_options` and trigger the stimulus action like we do with `search`:

https://github.com/bullet-train-co/bullet_train-core/blob/75bdbef029ee24312676c39d3a32e3b854a6376b/bullet_train-themes/app/views/themes/base/fields/_super_select.html.erb#L28

We've talked about eventually changing `other_options` to `bt_options` which I think is a clearer way of identifying what is Bullet Train-specific logic, so it may be preferable to use `other_options` for the time being, but either way, for now I'm okay with just documenting what's currently here in this PR.

cnorm35:
Thank you for taking a look @gazayas!  

I'll add some documentation to the super select field docs and also see if I can refactor this to use Stimulus.  I think using Stimulus values would let us use that value as an actual Boolean instead of checking the string value.

jagthedrummer (approved):
Looks great @cnorm35, thanks!

PR #744, opened 2024-01-16
Change hashids to sqids

gazayas:
I looked into https://github.com/bullet-train-co/bullet_train/issues/1232 a little bit and came across an update concerning hashids. According to the [documentation](https://github.com/peterhellberg/hashids.rb), hashids has been rebranded to `sqids`, so I used the new gem. The code itself was the same for the most part, but the object constructor was just a little different.

I think I'll try to implement the slug logic with `obfuscates_id` next.

### DIFF
--- a/bullet_train-obfuscates_id/Gemfile.lock
+++ b/bullet_train-obfuscates_id/Gemfile.lock
@@ -2,8 +2,8 @@ PATH
   remote: .
   specs:
     bullet_train-obfuscates_id (1.6.24)
-      hashids
       rails (>= 6.0.0)
+      sqids
 
 GEM
   remote: https://rubygems.org/
@@ -95,7 +95,6 @@ GEM
     erubi (1.12.0)
     globalid (1.2.1)
       activesupport (>= 6.1)
-    hashids (1.0.6)
     i18n (1.14.1)
       concurrent-ruby (~> 1.0)
     io-console (0.6.0)
@@ -130,6 +129,8 @@ GEM
       net-protocol
     nio4r (2.5.9)
     nokogiri (1.15.4-arm64-darwin)
+      racc (~> 1.4)
+    nokogiri (1.15.4-x86_64-darwin)
       racc (~> 1.4)
     nokogiri (1.15.4-x86_64-linux)
       racc (~> 1.4)
@@ -187,7 +188,9 @@ GEM
       actionpack (>= 5.2)
       activesupport (>= 5.2)
       sprockets (>= 3.0.0)
+    sqids (0.2.0)
     sqlite3 (1.6.8-arm64-darwin)
+    sqlite3 (1.6.8-x86_64-darwin)
     sqlite3 (1.6.8-x86_64-linux)
     stringio (3.0.9)
     thor (1.3.0)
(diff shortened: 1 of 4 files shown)


pascallaliberte:
This change is going to cause IDs to change.

1. `hashids` let us use a salt (we used "Default ID Obfuscation Salt for #{name}") but `sqids` no longer supports using one.
2. [`sqids` states that the algorithm has changed](https://github.com/sqids/sqids-ruby/commit/ca7b6bf324811488f176c3341091625bc952d9b1) and so the generated IDs have changed.

## What to do?

The README on the Hashids GitHub org, talking about the rebrand, [advises to use sqids only on new projects.](https://github.com/hashids#is-the-new-algorithm-a-drop-in-replacement-for-hashids)

> ## Is the new algorithm a drop-in replacement for Hashids?
> No. Sqids produces different IDs. We advise to use it only on new projects or if you are not using any pre-generated IDs.

Perhaps we can provide a grace period where both are available, and old links are directed to new ones automatically?

pascallaliberte:
Or maybe some way to store, once, the old hashids for all resource as a form of redirects table, as a prerequisite for installing the new version of the gem? A rake task to be run before the switchover is allowed.

gazayas:
> # Is the new algorithm a drop-in replacement for Hashids?
> No. Sqids produces different IDs. We advise to use it only on new projects or if you are not using any pre-generated IDs.

Good catch, @pascallaliberte! I missed that when I was looking around in the docs (I only caught what was on the initial READMEs).

> Or maybe some way to store, once, the old hashids for all resource as a form of redirects table, as a prerequisite for installing the new version of the gem? A rake task to be run before the switchover is allowed.

Just to make sure I understand what you mean, in this case, alongside with us giving a grace period for the dependency itself, developers would run the rake task to store the hashids which they could then choose to either drop later on or keep indefinitely, right?

Either way I think a grace period with redirects is reasonable.

gazayas:
Note to self: It may benefit us to have a request or system test in the starter repository with a string literal that checks if we're getting the correct ID:

```ruby
visit "/teams/KJkAvg"
assert_text "Testing Team"
```

pascallaliberte:
> Just to make sure I understand what you mean, in this case, alongside with us giving a grace period for the dependency itself, developers would run the rake task to store the hashids which they could then choose to either drop later on or keep indefinitely, right?

I'm not sure how I'd do it, because I'm not a fan of the potential mess a dual-gem grace period might create. So another approach is to have a migration task, maybe with a ENV var that if toggled to true switches gems, and that ENV var is set as part of a two-step process: rake task to inventory old ids for redirects (in a separate table with id and class_name), and then setting the ENV var switches on the `sqids` gem, turns off `hashids`, and enables the redirects.

jagthedrummer:
I'm not entirely convinced that we _need_ to move to `sqids`. `hashids` strikes me as one of those gems that's focused enough that it's just "done" and doesn't need any more work. As long as it's available to be installed on rubygems and works with the version of ruby that we're using I'm not sure that we need to rush to replace it.

gazayas:
> I'm not entirely convinced that we need to move to sqids. hashids strikes me as one of those gems that's focused enough that it's just "done" and doesn't need any more work. As long as it's available to be installed on rubygems and works with the version of ruby that we're using I'm not sure that we need to rush to replace it.

Sounds good to me! I'll close this one out for the time being, and we can refer back to this one if the need comes up.

PR #761, opened 2024-01-31
Ensure necessary dependencies are included in `bullet_train.gemspec`

### LINKED ISSUE #714: Add explicit dependency from `bullet_train` => `bullet_train-fields`
In the `bullet_train` gem we `require "bullet_train/fields"` in `lib/bullet_train.rb` but we don't have `bullet_train-fields` specified in the `.gemspec` file.

gazayas:
Closes #714.

I also added `bullet_train-super_scaffolding` because we have the following lines in the Membership controller base:

https://github.com/bullet-train-co/bullet_train-core/blob/dab52ad22c932a4c7003c8747ba1c668c39f2444/bullet_train/app/models/concerns/memberships/base.rb#L16-L17

I wanted to added `bullet_train-api` because of this line:

https://github.com/bullet-train-co/bullet_train-core/blob/dab52ad22c932a4c7003c8747ba1c668c39f2444/bullet_train/app/controllers/concerns/account/users/controller_base.rb#L16

However, `bullet_train-api` already depends on `bullet_train`:

https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-api/bullet_train-api.gemspec#L37

So, trying to include `bullet_train-api` was giving me this error:
```
> bundle install
Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem 'bullet_train' or gem 'bullet_train-api' and try again
```

# `Gemfile` and `bullet_train.gemspec`
We're including a lot of in-house gems in the `Gemfile` and not the `gemspec`. In some instances, we're including BOTH:

https://github.com/bullet-train-co/bullet_train-core/blob/dab52ad22c932a4c7003c8747ba1c668c39f2444/bullet_train/Gemfile#L14-L22

I'm curious if we could learn from the rails repository. It looks like they don't have Gemfiles for their repositories but include everything only in their gemspecs (which I'm under the impression that's what we SHOULD do). Would be glad to undertake that task if that's the direction we should head in.

## Active Record

https://github.com/rails/rails/blob/774c630068e69547fabd6bddedd62a712ae7e736/activerecord/activerecord.gemspec#L38-L39

## Action View
https://github.com/rails/rails/blob/774c630068e69547fabd6bddedd62a712ae7e736/actionview/actionview.gemspec#L36-L44


jagthedrummer:
> I feel like this is an issue that goes beyond the scope of #714, and perhaps we should re-frame it as how we include our dependencies in general.

Yeah, I think this is right. I think all of the BT dependencies are somewhat arbitrarily declared. Everything mostly works itself out by the time you have everything added in the `Gemfile` of the starter repo, but our declared dependencies don't seem to accurately match our expected/actual dependencies.

jagthedrummer:
Closing in favor of https://github.com/bullet-train-co/bullet_train-core/pull/970

PR #765, opened 2024-02-05
JavaScript: Remove calls to `$` (global jQuery definition)

pascallaliberte:
As a first step to remove our dependency on jQuery completely, we can at least remove our reliance on a global `window.$` definition.

This PR:
- Updates `super_select_controller.js` to remove reliance on `window.$`, import `jQuery` on `window.jQuery` if not already defined (required for select2)
- Updates `date_controller.js` to import `jQuery` just in the scope of the controller, without reliance on `window.$`
- Adds `jQuery` as a dependency in the `@bullet-train/fields` package
- Updates `color_picker_controller.js` to remove jQuery requirement
- Removes `reassignable.js`

Joint PR:
- https://github.com/bullet-train-co/bullet_train/pull/1319

Out of scope:

This PR doesn't replace `select2` or `daterangepicker`, and so jQuery will still be a requirement and part of the application bundle. `window.$` however no longer needs to be defined.


bullet_train-fields/app/javascript/controllers/fields/color_picker_controller.js
```diff
@@ -21,27 +21,23 @@ export default class extends Controller {
...
     this.teardownPluginInstance()
   }
 
+
+
   pickColor(event) {
     event.preventDefault()
 
     const targetEl = event.target
     const color = targetEl.dataset.color
```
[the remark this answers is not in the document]
kaspth: Alternatively we can use `getAttribute` to match the `setAttribute` in `showUserSelectedColor`?

```suggestion
    const color = event.target.getAttribute("data-color")
```

bullet_train-fields/app/javascript/controllers/fields/color_picker_controller.js
```diff
@@ -146,7 +141,12 @@ export default class extends Controller {
     this.pickr.destroy()
   }
 
+  get colorOptions () {
+    const colorButtons = this.colorOptionsTarget.querySelectorAll('button[data-color]')
+    return Array.prototype.slice.call(colorButtons).map((el) => { return el.dataset.color })
```
kaspth, line 146: I think we can use `Array.from` here, which is a little simpler. I can't remember if `...` is required though.

We can also cut `return` if we don't use wrapping braces in the anonymous function.

```suggestion
    return Array.from(...colorButtons).map((el) => el.dataset.color)
```

bullet_train-fields/app/javascript/controllers/fields/date_controller.js
```diff
@@ -29,13 +31,13 @@ export default class extends Controller {
     // don't submit the form, unless it originated from the cancel/clear button
     event.preventDefault()
 
-    $(this.fieldTarget).val('')
-    $(this.displayFieldTarget).val('')
+    jquery(this.fieldTarget).val('')
+    jquery(this.displayFieldTarget).val('')
```
kaspth, line 35: Can we use `value =` here like in the color picker controller?

```suggestion
    this.fieldTarget.value = ''
    this.displayFieldTarget.value = ''
```

bullet_train-fields/app/javascript/controllers/fields/date_controller.js
```diff
@@ -51,8 +53,8 @@ export default class extends Controller {
     )
     const displayVal = momentVal.format(format)
     const dataVal = this.includeTimeValue ? momentVal.toISOString(true) : momentVal.format('YYYY-MM-DD')
-    $(this.displayFieldTarget).val(displayVal)
-    $(this.fieldTarget).val(dataVal)
+    jquery(this.displayFieldTarget).val(displayVal)
+    jquery(this.fieldTarget).val(dataVal)
```
kaspth, line 57:
```suggestion
    this.displayFieldTarget.value = displayVal
    this.fieldTarget.value = dataVal
```

bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js
```diff
@@ -1,161 +1,182 @@
...
-
-    this.cleanupBeforeInit() // in case improperly torn down
-    this.pluginMainEl = this.selectTarget // required because this.selectTarget is unavailable on disconnect()
-    $(this.pluginMainEl).select2(options);
-
-    this.initReissuePluginEventsAsNativeEvents()
+    this.plugin = new TomSelect(this.selectTarget)
   }
 
   teardownPluginInstance() {
```
kaspth, line 21: I wonder if `initPluginInstance` and `teardownPluginInstance` add much when they're this simple now. But maybe that's for later!

jagthedrummer:
@pascallaliberte can you look into why tests are failing on this branch? I tried rebasing it to see if that would clear it up, but that didn't do the trick.

pascallaliberte:
@jagthedrummer I rebased the joint PR, which passed checks (I think?). I can never remember which tests should pass and which are okay not passing for -core and main repos respectively.

One of the issues I had spotted in this PR's checks was around the `postcss.config.js` and the use of `$ThemeStylesheetsDir`, which was merged only after this PR was submitted.

jagthedrummer:
Lookin' good, @pascallaliberte! All green here, and the ones failing in the starter repo are the ones running against the published gems (aka the one's we'd expect to fail in this situation). Thanks for tackling this!

PR #775, opened 2024-02-15
Make themes support single gem install, non-BT apps

pascallaliberte:
This PR:

* allows a non-Bullet Train app to include the Light theme
* adds bare-bones instructions for installing the Light theme on `rails new` and Jumpstart PRO rails projects
* moves required helpers, rake tasks and other requirements out of the `bullet_train` gem and into either the `bullet_train-themes` or `bullet_train-fields` gems
* adds `bullet_train-fields` as a dependency to `bullet_train-themes`
* makes the `phone_field` fall back to use `intl_tel_input_utils_path` from `cdn` if not defined
* defines `linked_gems` in `bullet_train-themes` for `bin/link` to work
* makes `current_team` (and `current_team_time_zone`) optional when found in theme partials
* makes `BulletTrain::Configuration` optional when found in theme partials
* renames `ThemeHelper` to `CurrentThemeHelper` to avoid conflict with Jumpstart Pro's own `ThemeHelper`
* uses `object.class.human_attribute_name` in `attribute` partial if locale string not defined for model field

Not in scope:

* rake task for automating installs
* mailer style sheets
* bin/resolve support
* local development utilities for non-BT projects
* scaffold overrides for Jumpstart PRO
* the views found in the `bullet_train`, `bullet_train-ongoing-webhooks`, `bullet_train-integrations-stripe`, `bullet_train-api` and `bullet_train-super_scaffolding` gems
* adapt `can?` calls in theme partials to other authorization gems

To-do:

- [x] Add instructions for `rails new` projects to make the style sheets work with the assets pipeline... something is off
- [x] Add an example locales for a model so that field and attribute partials can have those labels and options configurable
- [x] Pass CI tests
- [x] Run through instructions again on a Jumpstart PRO project
- [x] Fix importing `$ThemeStylesheetsDir/application.css` on `rails new` project
- [x] Fix esbuild.config.mjs on Jumpstart PRO to include theme javascript as endpoint
- [x] Run through instructions again on a `rails new` project


bullet_train-themes/app/helpers/current_theme_helper.rb
```diff
@@ -1,4 +1,4 @@
-module ThemeHelper
+module CurrentThemeHelper
```
jagthedrummer, line 1: Not a big deal, but it seems kind of odd to me that we're renaming things to accommodate a "competing" project. Is integration with JumpstartPro one of the explicit goals of this effort?

pascallaliberte: Yep, open to suggestions though

jagthedrummer: I guess I don't have much of a suggestion other than "let them sort it out" which probably isn't really helpful if the goal is for _us_ to sort it out.

pascallaliberte: I think the broader question is whether ThemeHelper is that good a name in the first place for this.

I can also keep it at ThemeHelper here and maybe add the step in the jumpstart pro instructions to rename their own ThemeHelper to something else?

jagthedrummer: Yeah, that's a good question. 🤔

PR #776, opened 2024-02-17
Enable bulk action select when one record is present to perform Action Model actions

gazayas:
Whenever Action Models is enabled, we give the user the option to select multiple records of the same model if two or more exist. This happens regardless if the model has an action model that performs actions on it, but I think we should only scaffold the following if the action model requires it:

```erb
<%= render 'shared/bulk_action_select' %>
```

I think that's a separate issue that would require more work, but for now with this PR, I just simply changed `many?` to `any?` because I think the `TargetsMany` action should target records >=1, but we currently have it set so `Select Multiple` is only visible when two or more records exist. Even if user intends to create many records in the future, this means that we can't perform the action at all when just one record is present. Currently if we wanted to target one record, we'd either have to create a new record to fulfill the `tangible_things.many?` requirement, or scaffold a `TargetsOne` action.


jagthedrummer (approved):
Sounds good.

PR #786, opened 2024-02-27
Check if billing_enabled? before enforcing billing requirements

julianrubisch:
instead of looking up `Billing::ControllerSupport`, which lead to false positives: 

when disabling billing in CI (via removing the `STRIPE_SECRET_KEY` or `PADDLE_SECRET_KEY` env variable), this would send me in an infinite redirect loop.

I can't think of any case where it would make sense _not_ to use the `billing_enabled?` helper method here, maybe I'm overlooking something?

### DIFF
--- a/bullet_train/app/controllers/concerns/account/controllers/base.rb
+++ b/bullet_train/app/controllers/concerns/account/controllers/base.rb
@@ -110,8 +110,7 @@ module Account::Controllers::Base
         end
       end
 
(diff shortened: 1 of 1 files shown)


kaspth (approved):
Can't think of why we shouldn't do this either 👍

gazayas:
Makes sense to me, we also check for `defined?(Billing::ControllerSupport)` in that method too.

gazayas (approved):

jagthedrummer (approved):
Makes sense to me.

PR #788, opened 2024-02-29
Add comment to method in `ClassNamesTransformer`

gazayas:
I was looking over the `ClassNamesTransformer` and wasn't quite sure what this method was doing, so I added an explanation so we won't have to dig in the future to find out what's going on.

I also wonder if there is some way to refactor `replacement_for` later on in the file since there are a lot of string literals there, but I think we can do that in another PR.


jagthedrummer (approved):
Nice.

PR #794, opened 2024-03-06
Move current theme definition to `ApplicationHelper` in light theme gem

gazayas:
Context: https://discord.com/channels/836637622432170028/1214513818093228032

This doesn't account for the use of `BulletTrain::Themes::Light` in the public layout, but it should take care of the `current_theme` definition issue.

### DIFF
--- a/bullet_train-themes-light/app/helpers/application_helper.rb
+++ b/bullet_train-themes-light/app/helpers/application_helper.rb
@@ -1,4 +1,4 @@
-module LightThemeHelper
+module ApplicationHelper
   # override in app/helpers/application_helper.rb
   def current_theme
     :light
--- a/bullet_train-themes-light/app/helpers/theme_helper.rb
+++ b/bullet_train-themes-light/app/helpers/theme_helper.rb
@@ -1,4 +1,4 @@
-module ApplicationHelper
+module ThemeHelper
   # override in app/helpers/application_helper.rb
   def current_theme
     :light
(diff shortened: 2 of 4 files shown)


pascallaliberte:
I'll have to check, but I think this will break compatibility with using themes in non-BT apps implementing ApplicationHelper locally.

gazayas:
Yup, just confirmed that that's the case for me.

I'm noticing that we store our themes in `BulletTrain::Themes`:
```
> rails c
Loading development environment (Rails 7.1.3.2)
irb(main):001> BulletTrain::Themes.themes
=> {:light=>#<BulletTrain::Themes::Light::Theme:0x00007f031fce3898>}
```

Maybe we should exclusively set/get themes via `BulletTrain::Themes` as opposed to making new modules for each theme or setting `current_theme` in the `ApplicationHelper`. That could also account for the [public layout](https://github.com/bullet-train-co/bullet_train-core/blob/4a02e01e04c79eff2d8451cf5b91152b9d15beef/bullet_train/app/views/layouts/public.html.erb#L2) issue, and we could use `BulletTrain::Themes.current_theme` in place of `BulletTrain::Themes::Light`. Setting the `current_theme` seems like something we should be doing in an initializer, especially considering we already have [theme.rb](https://github.com/bullet-train-co/bullet_train/blob/main/config/initializers/theme.rb) in place. Will give this more thought and see if I can swing back to it soon.

jagthedrummer:
Gonna convert this to a draft so I don't accidentally merge it before it's ready.

pascallaliberte (approved):
Let's move back to defining current_theme in `ThemeHelper` and add an extra step to the instructions for installing the theme in non-BT apps, for them to define current_theme locally.

PR #803, opened 2024-03-26
Add docs on how to add an address to a model

adampal:
Took me longer to figure this out that I'd like to admit so thought we should add it to the docs!


adampal:
Never mind. I must have been messing up my super-scaffolding command. Ran it again and was able to do everything with scaffolding. Not sure why it didn't work the first time I tried it.

PR #812, opened 2024-04-05
Fix "$ is not defined in sortable_controller"

pascallaliberte:
Fixes https://github.com/bullet-train-co/bullet_train/issues/1420

Changes `sortable_controller.js` to use `window.jQuery` instead of `$`, define it if missing.


jagthedrummer (approved):
Thanks for the quick fix, @pascallaliberte!

PR #825, opened 2024-05-06
Super scaffolding fix for `--sortable` with a multi word parent

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train/issues/1421


gazayas:
@jagthedrummer Nice, I confirmed that this properly finds the underscored parent in the migration and adds the sort order column after it.

I'm having trouble actually sorting the records though. Doesn't seem to be working on `main` either when just super scaffolding a simple model. This PR definitely accounts for the missing column though!

jagthedrummer:
> I'm having trouble actually sorting the records though.

Sounds like we might have another issue. I'll look into it.

jagthedrummer:
@gazayas I looked into the sorting thing and it does work for me, but it's not at all obvious that the list is sortable and that you can drag the list items around. We should probably make it more obvious. I'll open a ticket about that.

gazayas:
@jagthedrummer Looks like I just needed to run `yarn install`, it's working for me now :+1:

PR #830, opened 2024-05-15
date_and_time_field partial handles no timezone set on team

wesgarrison:
This started happening when I updated to tag v1.7.4

If there is no timezone set on a Team (my `db:reset` did not set one by default) then checking `ActiveSupport::TimeZone.find_tzinfo(current_team_time_zone)` fails with
```
TZInfo::InvalidTimezoneIdentifier at /account/model/:id/new 
Invalid identifier: nil
```

Trace:
image

Also, I don't see where `user_tz` is being used, but we do things like `<% if current_user&.time_zone != current_team_time_zone %>` when that should maybe be the `user_tz` calculated above. (I didn't know if that's part of this issue or not and I don't have time to dig in more now.)


jagthedrummer (approved):
Thanks @wesgarrison. The test failures here are related to the fact that we recently updated the starter repo to use `yarn` version 4. I've address that problem for the `core` repo here: https://github.com/bullet-train-co/bullet_train-core/pull/832

I'm gonna go ahead and merge this and will watch CI on `main`.

PR #843, opened 2024-06-03
Fix: Don't rely on jquery when detecting user timezone

jagthedrummer:
We recently remove the global `jQuery` alias `$` in an effort to work towards eliminationg jquery completely. This PR changes the timezone detection that happens when we edit user details to not rely on jQuery at all, global or otherwise.

Fixes: https://github.com/bullet-train-co/bullet_train/issues/1504

### DIFF
--- a/bullet_train/app/views/account/onboarding/user_details/edit.html.erb
+++ b/bullet_train/app/views/account/onboarding/user_details/edit.html.erb
@@ -48,7 +48,7 @@
 <% end %>
 
 <script type="text/javascript">
-  $(document).on('turbo:load', function() {
+  document.addEventListener('turbo:load', function() {
 
     // generate a mapping of js timezones compared to rails timezones.
     var jsTimezoneMapping = {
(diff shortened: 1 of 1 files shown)


bullet_train/app/views/account/onboarding/user_details/edit.html.erb
```diff
@@ -60,17 +60,16 @@
...
 
     if (useBrowserTimeZone == "true") {
-      // set the form accordingly.
-      var $option = $("#user_time_zone option[value=\"" + railsValue + "\"]")
-      $option.prop('selected', true);
-
-      // update the select2 as well. is there a better way to handle this?
-      // why don't _they_ handle this for us?
-      $("#select2-user_time_zone-container").attr('title', $option.text());
-      $("#select2-user_time_zone-container").text($option.text());
```
jagthedrummer, line 73: These lines don't seem to be necessary anymore. Setting the `value` attribute on the select element itself (`#user_time_zone`) now seems to make select2 automatically update.

PR #848, opened 2024-06-05
Fix association names when scaffolding join models if they don't match the table name

jagthedrummer:
Fixes the association naming problem in: https://github.com/bullet-train-co/bullet_train/issues/1508

I'm going to create a separate issue about the incorrect `valid_` method that's added on one side of the association.

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
@@ -58,7 +58,7 @@ class Scaffolding::Attribute
     name.split("_id").first
   end
 
-  def association_table_name
+  def association_name
     association_class_name.tableize
   end
 
--- a/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
@@ -60,7 +60,7 @@ class Scaffolding::Attribute
 
   def association_name
     plural_association_name = association_class_name.tableize
-    is_has_many? plural_association_name : plural_association_name.singularize
+    is_has_many? ? plural_association_name : plural_association_name.singularize
   end
 
   def class_name_matches?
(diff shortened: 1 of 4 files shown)


bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
```diff
@@ -58,6 +58,20 @@ def association_class_name
     name.split("_id").first
   end
 
+  def association_name
+    association_class_name.tableize
```
kaspth, line 62: `tableize` will also pluralize, which I don't know if we want for `belongs_to`/`has_one` associations. For those, we'd probably want `underscore.tr("/", "_")`.

bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
```diff
@@ -58,6 +58,20 @@ def association_class_name
...
+
+  def class_name_matches?
+    # if namespaces are involved, just don't...
+    # TODO: I'm not entirely sure that extracting this conditional was the right thing to do.
+    # Are there scenarios where we want to assume a match even when namespaces are involved?
+    if options[:class_name].include?("::")
+      return false
+    end
+    name_without_id.tableize == options[:class_name].tableize.tr("/", "_")
+  end
```
kaspth, line 73: What are the cases that this logic is for, I'm having a bit of a hard time understanding it?

kaspth:
Ok, there's a whole bunch of code that I haven't really looked at and that it'd take me some more time to dive into, so feel free to forward without me :)

bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
jagthedrummer: 🤔 Hmm, good point. Currently I think this method is only being called in places where we have a `has_many` or `has_many through:` relationship, but I can see how this would be confusing and potentially lead to problems down the road. I'll clean it up a bit.

bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
jagthedrummer: @kaspth This is used in situations where you're scaffolding a relationship with a name that doesn't match the name of the table/model you're relating to. See the main description in this issue for an example: https://github.com/bullet-train-co/bullet_train/issues/1508#issue-2328545278

PR #855, opened 2024-06-13
refactor: namespace custom js events with bulk-actions

jagthedrummer:
Joint PR: https://github.com/bullet-train-pro/bullet_train-action_models/pull/1


bullet_train/app/javascript/controllers/bulk_actions_controller.js
```diff
@@ -22,7 +22,7 @@ export default class extends Controller {
 
   updateActions() {
     this.actionTargets.forEach(actionTarget => {
-      actionTarget.dispatchEvent(new CustomEvent('update-ids', { detail: {
+      actionTarget.dispatchEvent(new CustomEvent('bulk-actions:update-ids', { detail: {
```
pascallaliberte, line 25: I recommend changing from:

```js
new CustomEvent('bulk-actions:update-ids'
```

To this:

```js
new CustomEvent(`${this.identifier}:update-ids`
```

We're dispatching CustomEvents here because we're dispatching them from an element that isn't `this.element`. If we were, we could have used the helper `this.dispatch("event-name")` which already prepends `${this.identifier}:` to the event-name. So we're mirroring the underlying convention here.

pascallaliberte (approved):
Suggestions above are nice to have, but otherwise looks good

bullet_train/app/javascript/controllers/bulk_actions_controller.js
jagthedrummer: Good suggestion, thanks!

PR #865, opened 2024-06-20
Bug fix for `devise-two-factor` missing `$` and some documentation robustification

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train/issues/1555
Fixes https://github.com/bullet-train-co/bullet_train/issues/1556

The files in `bullet_train/app/views/account/two_factors/` still rely on a global `jQuery` object.

I tried to change them from this pattern:

```javascript
jQuery("#two-factor").html("<%= j render partial: "devise/registrations/two_factor"%>");
```

To this pattern:

```javascript
document.querySelector("#two-factor").innerHtml = "<%= j render partial: "devise/registrations/two_factor"%>";
```

But after I did that clicking the "Enable" button in the MFA setup area just didn't do anything. Not sure why.

### DIFF
--- a/bullet_train/app/views/account/two_factors/create.js.erb
+++ b/bullet_train/app/views/account/two_factors/create.js.erb
@@ -1,1 +1,1 @@
-$("#two-factor").html("<%= j render partial: "devise/registrations/two_factor"%>");
+jQuery("#two-factor").html("<%= j render partial: "devise/registrations/two_factor"%>");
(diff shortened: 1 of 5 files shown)


pascallaliberte (commented):
Real quick, I'm thinking a slightly larger refactor is recommended. It's probably worth:

1. replacing`create.js.erb` and other `ujs` js responses with a turbo response. I'm thinking responding with a turbo-stream `update` (not `replace`) of `#two-factor`'s internals using the contents of the `devise/registrations/two_factor`
2. replacing the two `link_to`'s in `_two_factor.html.erb` with `button_to` to create small proper forms, which will return a turbo_stream as in step 1

jagthedrummer:
@pascallaliberte, those refactoring suggestions sound reasonable, but is there any reason not to go ahead and release this fix since things are currently completely broken?

pascallaliberte:
@jagthedrummer ah yes right, if the `jQuery` replacement works for now, then yes, looks good to me.

PR #875, opened 2024-07-03
WIP: Fix attachments in the Trix editor

jagthedrummer:
We were incorrectly sanitizing rich text content and were overzealously removing important elements that should be passed through.

### DIFF
--- a/bullet_train-fields/app/helpers/fields/html_editor_helper.rb
+++ b/bullet_train-fields/app/helpers/fields/html_editor_helper.rb
@@ -5,7 +5,7 @@ module Fields::HtmlEditorHelper
     return string unless string
     # TODO this is a hack to get around the fact that rails doesn't allow us to add any acceptable protocols.
     string = string.gsub("bullettrain://", TEMPORARY_REPLACEMENT)
-    string = sanitize(string, tags: %w[div br strong em b i del a h1 blockquote pre ul ol li], attributes: %w[href])
+    string = sanitize(string, tags: %w[div br strong em b i del a h1 blockquote pre ul ol li action-text-attachment figure figcaption img], attributes: %w[href sgid content-type url filename filesize width height presentation src class])
     # given the limited scope of what we're doing here, this string replace should work.
     # it should also use a lot less memory than nokogiri.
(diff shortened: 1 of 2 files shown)


bullet_train-fields/app/helpers/fields/html_editor_helper.rb
```diff
@@ -5,7 +5,7 @@ def html_sanitize(string)
     return string unless string
     # TODO this is a hack to get around the fact that rails doesn't allow us to add any acceptable protocols.
     string = string.gsub("bullettrain://", TEMPORARY_REPLACEMENT)
-    string = sanitize(string, tags: %w[div br strong em b i del a h1 blockquote pre ul ol li], attributes: %w[href])
+    string = sanitize(string, tags: %w[div br strong em b i del a h1 blockquote pre ul ol li action-text-attachment figure figcaption img], attributes: %w[href sgid content-type url filename filesize width height presentation src class])
```
jagthedrummer, line 8: I've added `class` here because the `action-text-attachment`, `figure`, and `figcaption` elements generated all include a `class`, so I figured it would be good to pass them through. Are there reasons we _shouldn't_ allow `class` in user-generated content?

gazayas:
@jagthedrummer I can't say off the top of my head why the image isn't showing up in the edit view, but it looks like there's no sanitization happening as far as I can tell.

The form calls the persisted object here:

https://github.com/bullet-train-co/bullet_train-core/blob/4aacb3036a944708d03583a700efdccd302760ab/bullet_train-themes/app/views/themes/base/fields/_trix_editor.html.erb#L26

The string produced here is similar to what you posted above which includes `data-trix-attachment`, `filename`, etc.

This is being passed to the `shared/fields/field` partial, where the invocation just looks like this:

https://github.com/bullet-train-co/bullet_train-core/blob/4aacb3036a944708d03583a700efdccd302760ab/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb#L85

Maybe we need to handle the strings at one of these two points?

jagthedrummer:
Looks like the problem with the missing images inside the editor is a Rails bug, not a Bullet Train bug. https://github.com/rails/rails/issues/52233

PR #882, opened 2024-07-11
Pass attribute down to _base partial

julianrubisch:
I would love to pass `attribute` down to the theme's `attributes/_base` partial.

That would allow for more versatile _strategies_ to be crafted.


jagthedrummer (approved):
👍 Looks good. I'll try to get this released later today. Thanks!

PR #887, opened 2024-07-12
Enable scaffolding of image fields with the `multiple` option for ActiveStorage

jagthedrummer:
This allows you to pass the `{multiple}` option when super scaffolding image fields **IF** you're using ActiveStorage and not Cloudinary.

If you have Cloudinary activated you'll see something like this:

```
$ rails g super_scaffold Widget Team name:text_field images:image{multiple} single_image:image
You have Cloudinary enabled and tried to scaffold an image field with the `multiple` option.
At this time we do not support multiple images in a single Cloudinary image attribute.
We hope to add support for it in the near future.
For now you could use individual named image attributes, or you might try disabling Cloudinary and using ActiveStorage.
```

Note: I pointed this PR into #886 because the `file_field` and `active_storage_image` field partials share the same Stimulus controller, so we need to update them simultaneously. When these are both ready to merge I plan to merge #886 first, then rebase this PR on `main` (if needed) and retarget it to `main` before merging it.

Fixes: https://github.com/bullet-train-co/bullet_train/issues/1598

### DIFF
--- a/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
+++ b/bullet_train-super_scaffolding/lib/scaffolding/attribute.rb
@@ -178,10 +178,10 @@ class Scaffolding::Attribute
       # TODO: We're preserving cloudinary_image here for backwards compatibility.
       # Remove it in a future major release.
       options[:height] = 200
-      "image"
+      "image#{"s" if is_multiple?}"
     when "image"
       options[:height] = 200
-      "image"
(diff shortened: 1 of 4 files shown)


gazayas:
I didn't get a chance to review this one before it was merged, but this is a good quality improvement. I wonder if there are any other fields we're adding improper migrations for by using the `{multiple}` option?

PR #910, opened 2024-09-12
Update to OpenAPI spec v3.1.0

hopsoft:
The current implementation uses the deprecated `example` field which causes test failures whenever creating a new Bullet Train application, and whenever a new Super Scaffold is generated.

This PR updates the implementation to use the current `examples` field instead.

SEE: https://spec.openapis.org/oas/latest.html#fixed-fields-5

> [!NOTE]
> This change satisifies the test failure problem but we should verify that it returns what's truly needed for the OpenAPI docs. 

Related: https://github.com/bullet-train-co/bullet_train/issues/1520

### DIFF
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -97,7 +97,10 @@ module Api
             "description" => attribute_name.titleize.to_s
           }
 
-          attributes_output["example"].merge!({attribute_name.to_s => nil})
+          attributes_output["examples"] = [{
+            "summary" => "default",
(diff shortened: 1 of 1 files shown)


hopsoft:
Note that I still have a few places to patch up. Working on that now.

hopsoft:
That first stab was an AI recommendation. It points the way... sort of, but I believe the issue is in the [bullet-train-co/jbuilder-schema](https://github.com/bullet-train-co/jbuilder-schema) gem. Taking a deeper look.

hopsoft:
Abandoning this and will leave to the OpenAPI domain experts that did the original implementation.

PR #911, opened 2024-09-12
dependent_fields_frame: fix if toggled by checkbox

pascallaliberte:
If the dependable field is a checkbox, `field.value` won't store the `checked` state of the checkbox, and so the dependent fields frame will always be evaluated to true.

This PR fixes this issue.


jagthedrummer:
@pascallaliberte I'll try to pull this down and give it a run tomorrow. There are some failed tests, and on a re-run they didn't resolve themselves. At a quick glance they both look like they might be legit failures related to dependent fields in addresses.

pascallaliberte:
@jagthedrummer Oh snap, I missed a bug. Pushed an update.

Issue #917, opened 2024-09-21
Turbo prefetches set the current_team inadvertently


pascallaliberte:
By default, [Turbo prefetches links on hover](https://turbo.hotwired.dev/handbook/drive#prefetching-links-on-hover) to add more snapiness to the application.

This can be disabled application-wide by adding the following line to the head, for example, in an application's local `app/views/layouts/_head.html.erb` file:

    <meta name="turbo-prefetch" content="false">

`current_team` is a global property that helps disambiguate a page's team context when the team is neither in the URL as param nor discoverable by looking at a resource's member action (e.g. `#show`) when the resource has a link back to its team. Since most resources in BT are tied to a Team by default, that's usually not a problem, and we even have better access to the current team in views via the `@team` property. Even deeply-nested resources will have access to the global `@team` property, because of the way `loads_and_authorizes_resource` is configured.

`current_team` is set on each page view, by default, through the `ensure_onboarding_is_complete` in `bullet_train/app/controllers/concerns/account/controllers/base.rb`, just to be sure.

The combination of an automatic `current_team` update and turbo prefetch automatically visiting a link can add some surprises.

An example of a surprise is detailed in issue #916, where the `ability.rb` file is evaluated based on a stale `current_team`, set in a prior request compared to the current_team of the current page's team context.

PR #918, opened 2024-09-21
Prevent turbo-prefetch on team switch links

pascallaliberte:
As a first step to alleviate issues such as #917, where `current_team` gets set on page load, we can at least disable turbo-prefetch on team switch links.

CleanShot 2024-09-21 at 11 30 59


jagthedrummer (approved):
Good call on this, @pascallaliberte! I think this may actually explain some weird behavior that I've noticed before but hadn't been able to pin down.

PR #923, opened 2024-10-11
For sortable tables, only show main column when dragging

pascallaliberte:
Fixes the display of table rows while dragging, when sortable.

Before:

CleanShot 2024-10-11 at 10 44 12

After:

CleanShot 2024-10-11 at 10 45 00


jagthedrummer (approved):
Good call, @pascallaliberte!

PR #927, opened 2024-10-18
Imporoved Heroku Detection

jagthedrummer:
Heroku recently made some changes to how they set the `PATH` ENV var and now it no longer includes `/app/.heroku/` so our previous detection method is broken.

Now we'll look for `ENV['HEROKU_APP_NAME']`.


zackgilbert (approved):
I think this works. If they end up changing this (stop using this env variable), you can always add it yourself to your own ENV variables and be back to fine.

PR #929, opened 2024-10-21
Fixed issue where Zapier couldn't connect because unique constraint.

michaelkoper:
In Membership::Base a unique constraint of the user_email column was added last year which broke the Zapier's ConnectionWorkflow.

Simply adding user_email when creating the membership fixes this.


jagthedrummer (approved):
Thanks, @michaelkoper!

Issue #926, opened 2024-10-18
Errors raised when setting `config.raise_on_missing_callback_actions = true`


jagthedrummer:
Rails 7.2 introduced a new config option to make it so that an error is raised if the `only` or `except` list in a `before_action` mentions a method that doesn't exist on the controller. Currently the test suite in the starter repo gets a lot of failures if that is enabled.

This test will fail if that config option is set to true:

```
rails test test/controllers/account/scaffolding/completely_concrete/tangible_things_controller_test.rb:28
```

The individual failures look like this and don't include much information:

```
Account::Scaffolding::CompletelyConcrete::TangibleThingsControllerTest
  test_should_get_index                                           FAIL (0.70s)
        Expected response to be a <2XX: success>, but was a <404: Not Found>
        test/controllers/account/scaffolding/completely_concrete/tangible_things_controller_test.rb:30:in `block in <class:TangibleThingsControllerTest>'


Failures and errors:
Account::Scaffolding::CompletelyConcrete::TangibleThingsControllerTest
  test_should_get_index                                           FAIL (0.70s)
        Expected response to be a <2XX: success>, but was a <404: Not Found>
        test/controllers/account/scaffolding/completely_concrete/tangible_things_controller_test.rb:30:in `block in <class:TangibleThingsControllerTest>'
```

If you tail the test log you'll see an error at the bottom of the test run that says this:

```
AbstractController::ActionNotFound (The reorder action could not be found for the #<Proc:0x0000000158051b40 /Users/jgreen/.asdf/installs/ruby/3.3.4/lib/ruby/gems/3.3.0/gems/cancancan-3.6.1/lib/cancan/controller_resource.rb:15>
callback on Account::Scaffolding::CompletelyConcrete::TangibleThingsController, but it is listed in the controller's
:only option.

Raising for missing callback actions is a new default in Rails 7.1, if you'd
like to turn this off you can delete the option from the environment configurations
or set `config.action_controller.raise_on_missing_callback_actions` to `false`.
):
```

That message points to a file in the `cancancan` gem:

https://github.com/CanCanCommunity/cancancan/blob/c63cde14d419474a9da2ddb03956440045348577/lib/cancan/controller_resource.rb#L15

But that line in `cancancan` is just applying options defined in the controller, so I don't think that `cancancan` is at fault here.

I think this line is what's causing the problem:

[...]

PR #930, opened 2024-10-29
Allow the starter repo to set `config.raise_on_missing_callback_actions = true`

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train-core/issues/926

Joint PR: https://github.com/bullet-train-co/bullet_train/pull/1725

### DIFF
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -101,7 +101,7 @@ module Api
         end
       end
 
-      if has_strong_parameters?("Api::#{@version.upcase}::#{model.name.pluralize}Controller".constantize)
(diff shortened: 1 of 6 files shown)


bullet_train-api/app/helpers/api/open_api_helper.rb
```diff
@@ -183,9 +183,11 @@ def description_for(model)
...
     private
 
-    def has_strong_parameters?(controller)
-      methods = controller.action_methods
-      methods.include?("create") || methods.include?("update")
+    def has_strong_parameters?(controller_name)
+      "#{controller_name}::StrongParameters".constantize
+      true
+    rescue NameError
+      false
```
jagthedrummer, line 190: These changes make it so that this method actually does what it says. Previously it did something closer to `should_have_strong_parameters?` and that led to problems if `create` or `update` was defined but a `StrongParameters` module was not.

PR #944, opened 2024-11-14
fix model_class_name variable in exception raise

JeremiahChurch:
`model_class_name` doesn't exist in the file anywhere - instead of raising the string exception an undefined local var exception is returned.

This is a super small fix so that the proper error is returned

current:
```
NameError
undefined local variable or method `model_class_name' for class
```

with change:
```
"Your 'account_load_and_authorize_resource' is broken. Tried to reflect on the `team` association of [\"example\"], but didn't find one.
```


jagthedrummer (approved):
Thanks, @JeremiahChurch!

PR #947, opened 2024-11-16
transformer routes rescue notices: Slack -> Discord

pascallaliberte:
Ran across this message. This points the message to write us in Discord.

> We weren't able to automatically add your `#{routes_namespace}` routes for you. In theory this should be very rare, so if you could reach out on Slack, you could probably provide context that will help us fix whatever the problem was. In the meantime, to add the routes manually, we've got a guide at https://blog.bullettrain.co/nested-namespaced-rails-routing-examples/ .


jagthedrummer (approved):
Good catch, @pascallaliberte! With this being an error message I'm just curious if there's also a bug or something unexpected that landed you here.

PR #974, opened 2024-11-26
Allow address fields to be disabled

### LINKED ISSUE #573: Disabled field partials should indicate that they're disabled
You can currently disable (most #571, #572) field partials by passing `options: {disabled: true}` like so:

```erb
<%= render 'shared/fields/text_field',
      method: :name,
      options: { disabled: true}
%>
```

But when you do so the field _is_ disabled but we don't indicate in any way visually that the field is disabled. It looks totally normal but just doesn't accept any input.

CleanShot 2023-09-21 at 10 15 39

CleanShot 2023-09-21 at 10 16 17

#892 Fixes this for most field types, the ones still left to address are:

* [x] `color_picker`
* [x] `options` with `multiple: true` (aka checkboxes)
* [x] `super_select`
* [x] `address`
* [x] `trix_editor`

jagthedrummer:
This allows you to pass `options: {disabled: true}` to disable the entire address field.

Like this:

```erb
<%= render 'shared/fields/address_field', method: :address_value, options: {disabled: true} %>
```

It disabled every input that's a part of the address, regardless of whether the address has been populated previously.

Empty initial state:

CleanShot 2024-11-26 at 10 24 13

With a previously populated address:

CleanShot 2024-11-26 at 10 24 44

This finally fixes #573

### DIFF
--- a/bullet_train-themes-light/app/assets/stylesheets/light/fields/super_select.css
+++ b/bullet_train-themes-light/app/assets/stylesheets/light/fields/super_select.css
@@ -91,6 +91,10 @@
     }
   }
 
+  .select2-container--default.select2-container--disabled .select2-selection--single{
+    @apply bg-slate-200 dark:bg-slate-700 hover:bg-slate-200 hover:dark:bg-slate-700
+  }
+
   .select2-container .select2-dropdown {
     @apply border-2 border-slate-300 shadow-sm overflow-hidden !important;
     @apply border-primary-500 !important;
--- a/bullet_train-themes-light/app/assets/stylesheets/light/fields/super_select.css
+++ b/bullet_train-themes-light/app/assets/stylesheets/light/fields/super_select.css
@@ -91,7 +91,8 @@
     }
   }
 
-  .select2-container--default.select2-container--disabled .select2-selection--single{
+  .select2-container--default.select2-container--disabled .select2-selection--single,
+  .select2-container--default.select2-container--disabled .select2-selection--multiple{
     @apply bg-slate-200 dark:bg-slate-700 hover:bg-slate-200 hover:dark:bg-slate-700
   }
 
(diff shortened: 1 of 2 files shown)


bullet_train-themes-light/app/assets/stylesheets/light/fields/super_select.css
```diff
@@ -91,6 +91,11 @@
     }
   }
 
+  .select2-container--default.select2-container--disabled .select2-selection--single,
+  .select2-container--default.select2-container--disabled .select2-selection--multiple{
+    @apply bg-slate-200 dark:bg-slate-700 hover:bg-slate-200 hover:dark:bg-slate-700
+  }
+
```
jagthedrummer, line 98: This makes it so that disabled super selects have the same background colors as other disabled inputs. Without this disabling the address looks like this:

CleanShot 2024-11-26 at 10 46 41

bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/fields/_address_field.html.erb
```diff
@@ -4,19 +4,37 @@
...
+
+country_html_options ||= {}
+country_html_options[:data] ||= {}
+address_one_options ||= {}
+address_two_options ||= {}
+city_options ||= {}
+postal_code_options ||= {}
+postal_code_options[:data] ||= {}
+region_html_options ||= {}
+region_html_options[:data] ||= {}
```
jagthedrummer, line 18: We need to create separate `options` hashes for each individual input because if we pass the same hash to each input they all end up with the same `id` on the actual `input` elements. The duplicate `id` thing happens because of this line:

https://github.com/bullet-train-co/bullet_train-core/blob/c330e8750be54764395bc4f3d693f93f0a145a70/bullet_train-themes-light/app/views/themes/light/fields/_field.html.erb#L33

pascallaliberte (approved):
Rock on

Issue #654, opened 2023-11-09
Allow for non-existent event types in `Webhooks::Outgoing::Endpoint#event_types`


RichStone:
It's possible that event types are created dynamically and deleted afterward. The below method raises an error in case the event type was deleted and not cleaned up on all the endpoints:

https://github.com/bullet-train-co/bullet_train-core/blob/5e6d042448abba11abe929c3c2ec11618a6301de/bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/endpoint_support.rb#L36-L38

I'd propose to account for that case with an extra param like this:

```ruby
def event_types(ignore_deleted: false)
```

Because I don't think it can somehow be enforced that the deleting call cleans up related endpoints (I mean, it could, but I don't think we would want to go that route).

jagthedrummer:
@RichStone I can see the potential for the error your described, but I'm curious how and why you would run into it. Like, can you describe a reproduction scenario where I could see it fail?

If we went with the `ignore_deleted` param that you mentioned, would we need to update other parts of the code to pass that param with different values at different times, or is this something you ran into with custom code?

Are there times when you would want it to raise an error if we encounter the scenario you describe? Or would it be reasonable to change that method so that it _never_ raises an error in that scenario? What if change the method to be this?

```ruby
def event_types 
  Webhooks::Outgoing::EventType.where(event_type_ids) 
end
```

RichStone:
> @RichStone I can see the potential for the error your described, but I'm curious how and why you would run into it. Like, can you describe a reproduction scenario where I could see it fail?

Hey Jeremy, I think you should easily run into this if you:

1. Define an event type in event_types.yml.
2. Create a Webhooks::Outgoing::Endpoint with that type.
3. Remove that type.
4. Call Webhooks::Outgoing::Endpoint#event_types.

> If we went with the `ignore_deleted` param that you mentioned, would we need to update other parts of the code to pass that param with different values at different times, or is this something you ran into with custom code?
>
> Are there times when you would want it to raise an error if we encounter the scenario you describe? Or would it be reasonable to change that method so that it _never_ raises an error in that scenario? What if change the method to be this?
> 
> ```ruby
> def event_types 
>   Webhooks::Outgoing::EventType.where(event_type_ids) 
> end
> ```

[...]

PR #983, opened 2024-12-16
Allow for non-existent event types in `Webhooks::Outgoing::Endpoint#event_types`

jagthedrummer:
Fixes: https://github.com/bullet-train-co/bullet_train-core/issues/654

### The Set Up

Let's say you had created a `Widget` model, and had added `widget.*` events to `config/models/webhooks/outgoing/event_types.yml`:

```yaml
scaffolding/absolutely_abstract/creative_concept:
  # This automatically generates `created`, `updated`, and `deleted` webhook events.
  - crud
scaffolding/completely_concrete/tangible_thing:
  - crud
invitation:
  - crud
membership:
  - crud
widget:
  - crud
```

And then you (or one of your users) sets up an Outgoing Webhook using one of the widget events. Effectively this:

```ruby
  endpoint = Webhooks::Outgoing::Endpoint.create!(
    url: "https://example.com/webhook",
    name: "test",
    team: @team,
    event_type_ids: ["widget.created", "membership.created"]
  )
```

Then at a later time you decide that the `Widget` model was a bad idea, and so you remove it from `event_types.yml`.

### Old Behavior

Previously, if you then called `event_types` on that `endpoint` we'd raise an `ActiveHash::RecordNotFound` error:

```ruby
> endpoint.event_types
bullet_train-outgoing_webhooks (1.11.0)
  app/models/concerns/webhooks/outgoing/endpoint_support.rb:37:
  in `block in event_types':
  Couldn't find Webhooks::Outgoing::EventType with ID=widget.created
  (ActiveHash::RecordNotFound)
        from bullet_train-outgoing_webhooks (1.11.0) app/models/concerns/webhooks/outgoing/endpoint_support.rb:37:in `map'
        from bullet_train-outgoing_webhooks (1.11.0) app/models/concerns/webhooks/outgoing/endpoint_support.rb:37:in `event_types'
```

### New Behavior

Now if you call `event_types` on that `endpoint` we'll return an array of `Webhooks::Outgoing::EventType` objects for the still-good `event_type_ids` for the endpoint. Since `widget.created` is no longer a valid type, it won't show up in the `event_types` array _and_ we won't raise an error.

```ruby
> endpoint.event_types
=> [
      #<Webhooks::Outgoing::EventType:0x000000014dddc068 
        @attributes={:id=>"membership.created"}>
   ]
```

### DIFF
--- a/bullet_train-outgoing_webhooks/test/models/webhooks/outgoing/endpoint_test.rb
+++ b/bullet_train-outgoing_webhooks/test/models/webhooks/outgoing/endpoint_test.rb
@@ -1,0 +1,40 @@
+require "test_helper"
+
+class Webhooks::Outgoing::EndpointTest < ActiveSupport::TestCase
+  include ActiveJob::TestHelper
+
+  setup do
+    @team = Team.create!(name: "test-team")
+    @endpoint = Webhooks::Outgoing::Endpoint.create!(url: "https://example.com/webhook", name: "test", team: @team)
+  end
+
+  test "#create should accept valid event_types" do
+    @endpoint = Webhooks::Outgoing::Endpoint.create!(
+      url: "https://example.com/webhook",
+      name: "test",
+      team: @team,
+      event_type_ids: [Webhooks::Outgoing::EventType.all.first]
+    )
+    assert @endpoint.persisted?
+  end
+
+  test "#create should accept non-existent event_types" do
+    @endpoint = Webhooks::Outgoing::Endpoint.create!(
+      url: "https://example.com/webhook",
+      name: "test",
+      team: @team,
+      event_type_ids: ["fake-thing.create"]
+    )
+    assert @endpoint.persisted?
+  end
+
+  test "#event_types should not raise an error for non-existent event_type_ids" do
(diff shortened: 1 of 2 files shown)


bullet_train-outgoing_webhooks/test/models/webhooks/outgoing/endpoint_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  test "#create should accept non-existent event_types" do
+    @endpoint = Webhooks::Outgoing::Endpoint.create!(
+      url: "https://example.com/webhook",
+      name: "test",
+      team: @team,
+      event_type_ids: ["fake-thing.create"]
+    )
+    assert @endpoint.persisted?
+  end
```
RichStone, line 29: Yeah, maybe this was one of the reasons why the `.find()` was there, namely to raise on creation, since there is no other validation in place for that. It did make sense that we raised before here.

RichStone (commented):
Nice one adding tests here, I now see the most probable reason for the raise here 👀

Issue #999, opened 2024-12-26
Move dependency on `extended_email_reply_parser` into the Conversations gem


jagthedrummer:
For some reason `extended_email_reply_parser` is listed as a dependency on the main `bullet_train` gem, but it's not actually used anywhere in `core` or the starter repo. It seems like it was added to make the conversations gem work, but I'm not sure why it was added to `bullet_train` and not that gem.

`extended_email_reply_parser` depends on `charlock_holmes` which is very finicky to install due to native dependencies. Lots (possibly a majority?) of the problems that people run into getting Bullet Train running have to do with installing `charlock_holmes`. Removing it as a dependency in `core` should make things easier for people just starting out.

PR #1000, opened 2024-12-30
Remove `extended_email_reply_parser` dependency

jagthedrummer:
For some reason `extended_email_reply_parser` was added as a dependency to the main `bullet_train` gem even though it's not used anywhere in `core` or in the starter repo.

It _is_ used by `bullet_train-conversations` and is declared as a dependency there, so we don't need it here.

`extended_email_reply_parser` relies on `charlock_holmes` which is notoriously finicky to install due to native extensions. Removing this dependency should eliminate one of the big problems that people often run into when first setting up Bullet Train.

Fixes https://github.com/bullet-train-co/bullet_train-core/issues/999


jagthedrummer:
Closing in favor of https://github.com/bullet-train-co/bullet_train-core/pull/1002

PR #1065, opened 2025-02-20
support array IDs in obfuscates decode

### LINKED ISSUE #1064: Avo actions broken - decode IDs from array rather than singular
while trying to add actions to avo for singular actions I receive errors in the obfuscation decode concern

```
NoMethodError in Avo::ActionsController#handle
undefined method `to_i' for an instance of Array

def decode_id(id)
  return id if id.to_i > 0
  hashids.decode(id).first
end

id => ["kNWkNV"]

bullet_train-obfuscates_id (1.9.0) app/models/concerns/obfuscates_id.rb:21:in `decode_id'
bullet_train-obfuscates_id (1.9.0) app/models/concerns/obfuscates_id.rb:26:in `block in find'
bullet_train-obfuscates_id (1.9.0) app/models/concerns/obfuscates_id.rb:26:in `map'
bullet_train-obfuscates_id (1.9.0) app/models/concerns/obfuscates_id.rb:26:in `find'
avo (3.16.1) lib/avo/resources/base.rb:70:in `block in <class:Base>'
avo (3.16.1) lib/avo/execution_context.rb:41:in `instance_exec'
avo (3.16.1) lib/avo/execution_context.rb:41:in `handle'
avo (3.16.1) lib/avo/resources/base.rb:233:in `find_record'
avo (3.16.1) lib/avo/resources/base.rb:256:in `find_record'
avo (3.16.1) app/controllers/avo/actions_controller.rb:67:in `set_query' 
```

if I use a basic action in avo - the exception occurs before anything in the action is run

IE

```
class Avo::Actions::MyAction < Avo::BaseAction
  self.name = "Example"
  self.visible = -> { view != :index }

  def handle(query:, fields:, current_user:, resource:, **args)
    # this will never get run - the exception occurs earlier in the stack
    asdf
  end
end
```

A quick hack fix in the obfuscates_id.rb concern seems to fix my issue. Thought I'm unsure of any problems this creates.

```
# original
def decode_id(id)
    return id if id.to_i > 0
    hashids.decode(id).first
  end

# fixed
  def decode_id(id)
    if id.is_a?(Array)
      id.map { |single_id| decode_single_id(single_id) }
    else
      decode_single_id(id)
    end
  end
  
  def decode_single_id(id)
    return id if id.to_i > 0
    hashids.decode(id).first
  end
```

Would love to get feedback - happy to turn this into a PR. Without this I'm pretty much unable to make use of Avo actions. <3

JeremiahChurch:
resolves #1064

### DIFF
--- a/bullet_train-obfuscates_id/app/models/concerns/obfuscates_id.rb
+++ b/bullet_train-obfuscates_id/app/models/concerns/obfuscates_id.rb
@@ -24,7 +24,7 @@ module ObfuscatesId
         decode_single_id(id)
       end
     end
-    
+
     def decode_single_id(id)
       return id if id.to_i > 0
       hashids.decode(id).first
--- a/bullet_train-obfuscates_id/app/models/concerns/obfuscates_id.rb
+++ b/bullet_train-obfuscates_id/app/models/concerns/obfuscates_id.rb
@@ -18,6 +18,14 @@ module ObfuscatesId
     end
 
     def decode_id(id)
+      if id.is_a?(Array)
+        id.map { |single_id| decode_single_id(single_id) }
+      else
+        decode_single_id(id)
+      end
+    end
+    
+    def decode_single_id(id)
       return id if id.to_i > 0
       hashids.decode(id).first
     end


jagthedrummer (approved):
Looks great, @JeremiahChurch! Thank you for this. Apologies for letting it sit here so long, it managed to sneak past me a few weeks ago.

PR #1071, opened 2025-03-13
rm local: true from forms. it's old tech

yshmarov:
although I don't think it removes `local: true` from the super scaffold generator. I am not sure where the `_form.erb.tt` for super scaffold is declared.

Screenshot 2025-03-13 at 16 16 28


jagthedrummer:
Thanks, @yshmarov! This will get the super scaffolder as it uses either `bullet_train-super_scaffolding/app/views/account/scaffolding/absolutely_abstract/creative_concepts/_form.html.erb` or `bullet_train-super_scaffolding/app/views/account/scaffolding/completely_concrete/tangible_things/_form.html.erb` as the form template, I forget which. (We don't use "normal" generators for super scaffolding, so things work a little differently than other generators.)

PR #1072, opened 2025-03-18
Allow a subset of roles to be passed to permit

dansingerman:
This allows a call to `permit` to only create permissions for some of the Roles in the Roles.yml file.

### DIFF
--- a/bullet_train-roles/README.md
+++ b/bullet_train-roles/README.md
@@ -165,6 +165,12 @@ To access the array of all roles available for a particular model, use the `assi
 <% end %>
 ```
 
+You may want to restrict the roles considered at runtime too. To do this you can use the `included_roles` keyword of `permit`:
+
+```ruby
+permit user, through: :memberships, parent: :team, included_roles: Membership.assignable_roles
+```
+
 ## Checking user permissions
 
 Generally the CanCanCan helper method (`account_load_and_authorize_resource`) at the top of each controller will handle checking user permissions and will only load resources appropriate for the current user.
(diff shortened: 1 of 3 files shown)


jagthedrummer:
@dansingerman Can you say a little bit about the use case you have in mind for this? I see what's happening, but I don't really understand the problem that's being solved.

jagthedrummer:
Whatever the use case it, it might be good to add it to the documentation so that people know that they can do it, and when they should.

dansingerman:
It's a bit of an esoteric use case. 

If you have multiple `permit` calls, granting access to different models, or through different parent models, you may want only some roles to apply to some of those `permit` calls.

I think it's equivalent functionality to the `roles_only` logic here https://github.com/bullet-train-co/bullet_train-core/tree/main/bullet_train-roles#restricting-available-roles - but we need to apply it based on a feature gate, so we can't define it at class level.

jagthedrummer:
OK, thanks for the added context, @dansingerman. That looks like a decent place to document this addition. Could you add a bit there that lets people know that this new feature exists? And ideally it would have a somewhat concrete example like the one you pointed to. Something that describes a concrete scenario/requirement that can't be accomplished without the new feature. Like: "Imagine you have a Widget model and roles `admin` & `editor` with these permissions ____ and let's say you wanted to make it so that ____, the normal `permit` call would do _____ which isn't what you want, but if you add `included_roles` then ____"

jagthedrummer:
Also, FYI, I'm gonna be slammed for the next few days so it'll be next week before I can get back to this.

dansingerman:
@jagthedrummer I have added documentation and a minimal use case to the README.md

jagthedrummer (approved):

PR #1075, opened 2025-03-21
turbo_confirm instead of confirm; button_to for non-GET requests

yshmarov:
rails 7 without `rails-ujs` is supposed to use `turbo_confirm` instead of `confirm`

confirmation dialog does not show up unless we make it `turbo_confirm`

Screenshot 2025-03-21 at 21 02 21


jagthedrummer:
@yshmarov it looks like this has broken some tests. Can you look into those?

jagthedrummer:
@yshmarov I see that you made some additional changes, but there are still some failing tests on this.

jagthedrummer:
@yshmarov I just looked at this again and I think I'm just going to close it and start fresh. You changed a bunch of `link_to` tags to `button_to` for some reason, and they seem arbitrary. And you didn't make any styling changes to account for them changing appearance. And `turbo_confirm` needs to be paired with `turbo_method` which is why there are a bunch of broken tests.

PR #1076, opened 2025-03-22
Add code_editor field

pascallaliberte:
Adds a rich code editor based on `monaco-editor`. See https://microsoft.github.io/monaco-editor/

Because `monaco-editor` is a massive library with many optional parts (coding languages, themes), we're using an on-demand loading approach. Although `monaco-editor` is mentioned in yarn.lock files, it isn't packaged as part of the js payload, but loaded via injecting a script tag in the head on Stimulus `connect()`.

When offline, the code_editor field will fallback to a plain textarea.

Included in this PR:
- [x] `super_scaffold` support
- [x] a `code_snippet` attribute partial, defaulting to show the whole code snippet inline

For another PR:
- [ ] specifying the `language` as part of the super_scaffold command

CleanShot 2025-03-21 at 21 22 19@2x
CleanShot 2025-03-21 at 21 22 29@2x
CleanShot 2025-03-21 at 21 22 08@2x
CleanShot 2025-03-21 at 21 21 49@2x

## `code_snippet` partial

CleanShot 2025-03-27 at 15 06 46@2x


bullet_train-fields/package.json
```diff
@@ -48,6 +48,7 @@
     "emoji-mart": "^5.1.0",
     "microbundle": "^0.13.0",
     "moment-timezone": "^0.5.43",
+    "monaco-editor": "^0.52.2",
```
jagthedrummer, line 51: Since we're not actually using this package directly I think it would be better to remove this, so that we don't end up forcibly including this as part of an application JS bundle. Is that reasonable/possible?

jagthedrummer (changes requested):
Cool addition, @pascallaliberte. I didn't know about Monaco before. I had thought that it would be cool to add a [Codemirror](https://codemirror.net/) field type for this kind of thing, but I never got around to it.

Relatedly, I wonder if we should change the field type to something like `monaco_editor` since we already do `trix_editor` instead of `richtext_editor`. 🤔 What do you think on that?

I think we should also add this to the super scaffolder so that you can generate these along with your model and/or easily add them after the fact. And we should mention it in the field partials docs.

bullet_train-fields/package.json
pascallaliberte: Because `code_editor_controller.js` is only extracting the current version number (for fetching a precise CDNed version), nothing else from the package is included in the application.js output. You can verify this via `assets:precompile`.

We should keep this here to benefit from dependencies auto-updates via CI.

pascallaliberte:
@jagthedrummer:

> I wonder if we should change the field type to something like monaco_editor since we already do trix_editor instead of richtext_editor. 🤔 What do you think on that?

I understand that `trix` was a bit of a branded feature, and so I understand how it had its own named field type, whereas `date_field` (`daterangepicker`) and `color_picker` (`pickr`) did not. Funny enough, a large BT app we worked on replaced trix with rhino-editor, and kept trix_editor as the partial name in the project anyway. I think `trix_editor` should be renamed to `rich_text_editor` personally, even if the `ckeditor` partial exists.

Up to you, but I recommend calling it `code_editor`. Ejecting the partial and overriding the Stimulus controller would be the way to override that choice in a child project. Not much except the `monaco_config`, `theme_light` and `theme_dark` params indicates inherently that this is a monaco editor, in the partial anyway.

bullet_train-fields/package.json
jagthedrummer: So, if we're gonna have it in `package.json` why load it from CDN? Seems like an unusual pattern that I haven't seen before and don't really understand.

pascallaliberte: But you're asking why load it from CDN. Because this package is heavy, and is best loaded in pieces: the theme file you need, the code language you need.

jagthedrummer:
Sorry for the delay on this, @pascallaliberte. I'm going to pull it down and give it a good run through today and will hopefully get it out this week.

bullet_train-fields/package.json
jagthedrummer: OK, I think that makes sense.

jagthedrummer (approved):
🚀 🤘

PR #1081, opened 2025-03-28
Setup and add tests for automatic_components_for helper

Riddlerrr:
Continue of https://github.com/bullet-train-co/bullet_train-core/pull/1080 with test setup for `api` gem and added tests for `automatic_components_for` helper

I decided to put it in a separate PR in case we have any issues/comments with the test setup or anything else not related to the fix itself.
This way we can merge the fix (https://github.com/bullet-train-co/bullet_train-core/pull/1080) first and tests (this PR) separately.
Or we can just merge this PR only.

✅ Initial fix included

---

PR Description from #1080 (which is included in this PR):

✅ The changes are tested in a real project with docs generation.

This PR allows you to disable a required field for OpenApi generation.
It was previously impossible to do this when a field has a presence validation.

It can be very useful for update params, where you can skip some parameters in the payload if you don't want to update them.

For example:

```
automatic_components_for Workspace, parameters: {update: {add: {subdomain: {required: false}}}
```

will return:
```
WorkspaceParametersUpdate:
      type: object
      title: Workspaces
      description: Workspaces
      required:
      - name
      properties:
        name:
          type: string
          description: Workspace name
        subdomain:
          type: string
          description: Subdomain
```

Since it's post-processing, this will work even if the field has presence validation.

P.S. I tried to add a test for this because it doesn't seem to be set up for the `api` gem and it throws 
`Minitest::UnexpectedError:         ActiveRecord::StatementInvalid: Could not find table 'teams'`, so I gave up.

### DIFF
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -50,8 +50,10 @@ module Api
     def automatic_components_for(model, **options)
       locals = options.delete(:locals) || {}
 
-      path = "app/views/api/#{@version}"
-      paths = [path, "app/views"] + gem_paths.product(%W[/#{path} /app/views]).map(&:join)
+      view_path = "app/views"
+      api_path = "#{view_path}/api/#{@version || "v1"}"
+      paths = [Rails.root.join(api_path).to_s, Rails.root.join(view_path).to_s]
+      paths += gem_paths.product(%W[/#{api_path} /#{view_path}]).map(&:join)
 
       # Transform values the same way we do for Jbuilder templates
       Jbuilder::Schema::Template.prepend ValuesTransformer
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -249,6 +249,9 @@ module Api
           if details["required"]
             original["required"] << property
             details.delete("required")
+          elsif details["required"] == false
+            original["required"].delete(property)
+            details.delete("required")
           end
           original["properties"][property] = details
           if details["example"]
--- a/bullet_train-api/app/helpers/api/open_api_helper.rb
+++ b/bullet_train-api/app/helpers/api/open_api_helper.rb
@@ -248,11 +248,12 @@ module Api
         custom["add"].each do |property, details|
           if details["required"]
             original["required"] << property
-            details.delete("required")
-          elsif details["required"] == false
+          else
             original["required"].delete(property)
-            details.delete("required")
           end
+          # Always delete required because this attribute should only go
+          # into the required array according to OpenAPI 3.1.
+          details.delete("required")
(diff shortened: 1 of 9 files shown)


bullet_train-api/test/factories/teams.rb
```diff
@@ -0,0 +1,16 @@
+FactoryBot.define do
+  factory :team, class: Team do
+    name { "Test Team" }
+    slug { "test_team" }
+    time_zone { "Pacific Time (US & Canada)" }
+  end
+
+  factory :team_example, class: Team do
+    id { 42000 }
```
RichStone, line 9: FYI @jagthedrummer, we hardcode the team ID for doc gen so all child records have the same team_id reference.

bullet_train-api/test/helpers/open_api_helper_test.rb
```diff
@@ -0,0 +1,157 @@
...
+
+class Api::OpenApiHelperTest < ActiveSupport::TestCase
+  class TestClass
+    include Api::OpenApiHelper
+
+    def initialize
+      @version = "v1"
+    end
+  end
+
```
RichStone, line 11:
```suggestion

  # TODO: In the future, we can solve this more elegantly than providing
  # a comment here and doing individual asserts. For example, by having a
  # team_attributes.yml fixture and then doing one assert_equal against a
  # string or a `expected_hash.to_yml`.
  #
  # An example of the resulting OpenAPI schema:
  #
```

Added a TODO comment, open for discussion if I'm on the wrong path here.

RichStone (approved):
This is grand, thanks man!

bullet_train-api/app/helpers/api/open_api_helper.rb
```diff
@@ -50,8 +50,10 @@ def automatic_paths_for(model, parent, except: [])
     def automatic_components_for(model, **options)
       locals = options.delete(:locals) || {}
 
-      path = "app/views/api/#{@version}"
-      paths = [path, "app/views"] + gem_paths.product(%W[/#{path} /app/views]).map(&:join)
+      view_path = "app/views"
+      api_path = "#{view_path}/api/#{@version || "v1"}"
+      paths = [Rails.root.join(api_path).to_s, Rails.root.join(view_path).to_s]
+      paths += gem_paths.product(%W[/#{api_path} /#{view_path}]).map(&:join)
```
RichStone, line 56: Kudos to finding the solution here for the dummy app!

RichStone:
@Riddlerrr I would celebrate this with a new "Tests" section in the BT API README:

https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-api/README.md

RichStone:
> Looks good. Thanks for getting the testing situation sorted out and for adding tests of the new functionality. 🤘 

Yeah, this is a pretty slick milestone for the API gem and will allow us to move forward with more fun and confidence to improve that gem. Thanks for the merge!

PR #1082, opened 2025-03-31
replace hashids with sqids

yshmarov:
https://github.com/peterhellberg/hashids.rb

Screenshot 2025-03-31 at 08 58 41

=> use https://github.com/sqids/sqids-ruby

### DIFF
--- a/bullet_train-obfuscates_id/Gemfile.lock
+++ b/bullet_train-obfuscates_id/Gemfile.lock
@@ -4,6 +4,7 @@ PATH
     bullet_train-obfuscates_id (1.19.2)
       hashids
       rails (>= 6.0.0)
+      sqids
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


bullet_train-obfuscates_id/Gemfile.lock
```diff
@@ -4,6 +4,7 @@ PATH
     bullet_train-obfuscates_id (1.19.2)
       hashids
```
jagthedrummer, line 5: Running `bundle install` should have removed this `hashids` line. The fact that it's still here is causing test failures. Can you remove it, please?

jagthedrummer:
@yshmarov I just tested this branch and currently `sqids` is producing a different obfuscated id than `hashids` which would break links that people have already published to resources in their app. If there's a way to get `sqids` to produce the same value as `hashids` then we could consider switching them out, but breaking published links is going to be a definite blocker and deal breaker. Can you see if there's something we can do about that?

jagthedrummer:
I looked into this and I don't believe it's possible to move to `sqids`. They use a different algorithm and produce different hashed ids. That's a deal breaker for us since we have to support applications that are already in the wild. I'm going to close this, but if you find that there's a way to make `sqids` backward compatible then feel free to re-open it.

Issue #1110, opened 2025-04-02
We can remove `np` from all of the `package.json` files


jagthedrummer:
Since we release new versions via a GitHub Actions workflow, and it doesn't use `np` I don't think there's any reason to have it in the project.

PR #1111, opened 2025-04-02
Remove `np` from all the `package.json` files

jagthedrummer:
Since we release new versions via a GitHub Actions workflow, and it doesn't use `np` I don't think there's any reason to have it in the project.

Fixes https://github.com/bullet-train-co/bullet_train-core/issues/1110

### DIFF
--- a/bullet_train-fields/package.json
+++ b/bullet_train-fields/package.json
@@ -20,9 +20,7 @@
     "clean": "rimraf app/assets/javascripts",
     "prewatch": "$npm_execpath run update-local-installs",
     "watch": "npm-watch",
-    "update-local-installs": "$npm_execpath run build; yalc push",
-    "prerelease": "$npm_execpath run build",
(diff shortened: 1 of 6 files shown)


jagthedrummer:
@pascallaliberte, do you know any reason we shouldn't do this?

pascallaliberte:
It simplified releasing a npm package at the time, a process we need to publish new versions of our fields package, etc.

We'd need to find something else to fill the shoes of the release script.

jagthedrummer:
@pascallaliberte We currently release JS packages via a GitHub Actions workflow that pushes them and the Ruby gems all at once, and that process doesn't use `np`.

pascallaliberte (approved):

jagthedrummer:
Thanks, @pascallaliberte. I should have been more descriptive in the PR description about there already being a replacement in place. Sorry about that.

PR #1112, opened 2025-04-04
Replace cp-cli for copy:intl-tel-input-utils script

### LINKED ISSUE #1109: We need to replace or remove `cp-cli` in `bullet_train-fields`
The latest version of this is 6 years old, and the repo has been archived by the owner.

It's currently keeping some JS packages pinned to old versions that have [security vulnerabilities](https://github.com/bullet-train-co/bullet_train-core/security/dependabot/15).

@pascallaliberte, just curious if you happen to know about anything that would be a good alternative.

pascallaliberte:
Uses an inline `node -e` script for copying `node_modules/intl-tel-input/build/js/utils.js` into place for the `copy:intl-tel-input-utils` command.

Closes #1109

### DIFF
--- a/bullet_train-fields/package.json
+++ b/bullet_train-fields/package.json
@@ -14,8 +14,9 @@
     "update-local-installs": "app/javascript/**/*.js"
   },
   "scripts": {
-    "prebuild": "$npm_execpath run clean; $npm_execpath run copy:intl-tel-input-utils",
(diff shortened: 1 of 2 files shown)


jagthedrummer (approved):
Great idea on using a `node -e` approach! (I just pushed a small tweak to get CI working.)

Issue #854, opened 2024-06-13
Automatically disabling dead webhook endpoints


RichStone:
BT currently has a very short webhook delivery schedule and a brute force webhook delivery mechanism that does not have a notion of `Webhooks::Outgoing::Endpoints` being dead. This can lead to a large amount of delivery attempts to endpoints that are abandoned or misconfigured unintentionally. At the same time, app webhook endpoint creators (admins, API client platforms) currently don't have a way to know if they made a typo in their endpoint or if their endpoint is unhealthy otherwise for one reason or another. 

One solution one be to go for a delivery model similar to what Stripe does, where an endpoint is marked inactive after a period of time without any successful responses, e.g., within 3 days. This also means that we'll extend the webhook delivery attempts for the same time with an exponential backoff. 

Customers should be notified multiple time in the process: 

1. The endpoint started to be unhealthy.
2. The endpoint is about to be deactivated.
3. The endpoint is deactivated.

We'd need to keep in mind that individual deliveries may fail (e.g., because of validation issues) but that the endpoint is intact overall because other deliveries succeed. Stripe deactivates endpoints ~only when _all_ deliveries within a certain amount of time are failing~ according to a dynamic algorithm that decides whether an endpoint is "performing".

In the ideal scenario, we would have reasonable defaults that are also configurable, e.g., in case a BT user wants to keep endpoints alive for shorter or longer periods of time.

jagthedrummer:
@RichStone I haven't personally used outgoing webhooks in BT, so I'm not at all familiar with how everything works currently (and our documentation about it is very lacking). Do you already have a feel for exactly how we should approach this? Like, are there existing models and record keeping that we can lean on for this, or are we building the foundations for it from scratch?

Also, what exactly does "customers should be notified" mean? Notified how? Via email? Some sort of in-app notification system (I don't think we currently have such a system)? Who exactly should be notified? All members of the team that owns an endpoint?

RichStone:
### Delivery schedule suggestion

The current schedule is

```ruby
  ATTEMPT_SCHEDULE = {
    1 => 15.seconds,
    2 => 1.minute,
    3 => 5.minutes,
    4 => 15.minutes,
    5 => 1.hour,
  }
```

According to the requirements, we need to increase that schedule for a way longer time, an exponential backoff is used most of the time that looks something like that:

```ruby
  ATTEMPT_SCHEDULE = {
    1 => 15.seconds,
    2 => 1.minute,
[...]

PR #1125, opened 2025-05-21
Webhook deactivation system

Riddlerrr:
This is an implementation for the issue: https://github.com/bullet-train-co/bullet_train-core/issues/854

This PR introduces the `Webhooks::Outgoing::EndpointsDeactivationJob`, which automatically deactivates endpoints that fail to process webhooks for any reason.

The job is supposed to be a scheduled job and run every X (configurable) amount of time. For this reason, `sidekiq-scheduler` was introduced to the main `bullet_train` template in [the PR](https://github.com/bullet-train-co/bullet_train/pull/2074). However, the BT core does not have any scheduled mechanisms.

This PR adds some new config settings to `outgoing_webhooks`:
```
          automatic_deactivation_endpoint_enabled: false,
          automatic_deactivation_endpoint_settings: {
            max_limit: 50,
            deactivation_in: 1.day
          },
```

I think it's better to disable the feature by default so that it can only be enabled if needed.

### How it works

The `EndpointsDeactivationJob` is split into two parts:
1. Mark endpoints for deactivation when the maximum limit of failed deliveries is reached.
2. Deactivate the marked endpoints if there are no successful deliveries.

#### Mark endpoints for deactivation

There are two new fields added to `webhooks_outgoing_endpoints`: `deactivation_limit_reached_at` and `deactivated_at`.

The job searches for the not delivered `Webhooks::Outgoing::Delivery` for each endpoint and if it is >= the `max_limit` from the config, and sets the `deactivation_limit_reached_at`.

A delivery is considered a failure if it is not delivered within the allotted time frame:

https://github.com/Riddlerrr/bullet_train-core/blob/81f7fca191aecc073ecaf71c3c646391e5eb08bf/bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb#L15-L17

We will send a notification when an endpoint is marked to deactivate.

Since the `deactivation_limit_reached_at` set, we will wait for `deactivation_in` (1 day by default) and if no successful deliveries were made during the time the job will deactivate it by setting `deactivated_at`.

For any successful delivery, the `deactivation_limit_reached_at` must be cleared so that the endpoint is no longer considered "marked for deactivation."

#### Deactivate the marked endpoints

The job checks for endpoints with a set `deactivation_limit_reached_at`. If there were no successful deliveries since the  `deactivation_limit_reached_at`, it marks the endpoint as deactivated by setting the `deactivated_at` field.

### Deactivated endpoints

Even when an endpoint is deactivated, the system will create deliveries for debugging purposes to confirm that the event occurred. However, actual delivery attempts will not be created, so the webhooks will not fire.

### UI and controller actions

This PR adds the ability to activate/deactivate endpoints. For that reason, two new actions were introduced to `Account::Webhooks::Outgoing::EndpointsController` and `Api::V1::Webhooks::Outgoing::EndpointsController`.

CleanShot 2025-05-28 at 13 39 41@2x

CleanShot 2025-05-28 at 13 28 07@2x

### ⚠️ BEFORE MERGE

[The PR](https://github.com/bullet-train-co/bullet_train/pull/2074) has to be merged first.
It also should fix the CI in the PR, because it's failing because of missing fields in the master branch of the template repo.


bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
```diff
@@ -32,6 +40,7 @@ def deliver_async
   end
 
   def deliver
+    return if endpoint.deactivated?
```
Riddlerrr, line 43: This line prevents the creation of a new attempt and delivery of the webhook.

bullet_train-outgoing_webhooks/test/dummy/app/models/ability.rb
Riddlerrr, line 1: This is part of the controller's test setup.

bullet_train-outgoing_webhooks/test/dummy/config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  devise_for :users
```
Riddlerrr, line 2: Fix for make `sign_in @user` works for controller's tests

bullet_train-themes-tailwind_css/app/views/themes/tailwind_css/attributes/_attempt.html.erb
```diff
@@ -6,6 +6,8 @@
   <span class="inline-block">
     <% if object.send(success_method) %>
       <i class="fal fa-check ti ti-check text-green-500"></i>
+    <% elsif object.send(failure_method) %>
```
Riddlerrr, line 9: I moved that up because the `still_attempting?` method will return true even after the delivery time has expired.

jagthedrummer:
@Riddlerrr I'll check this out next week. When you have a few minutes can you rebase your branch against `main`? I think a bunch of the test failures are due to your branch missing some changes from `main`.

Riddlerrr:
> I think a bunch of the test failures are due to your branch missing some changes from main.

It is up to date. CI fails due to missing fields (migration) in the main branch of the template repo.

jagthedrummer:
OK, @Riddlerrr we have a process for getting CI working on these cross-repo type PRs.

1. We need to get you permissions to be able to push branches to `core` and the starter repo, if you don't already have them.
2. You name your branches in both projects the same.
3. You push your branches to the repo in both projects and open PRs.
4. Then when CI runs in each repo it will checkout the matching branch in the other repo so that everything matches up.

I kind of assume that since this is coming from your own fork that you do not have permissions to push a branch to this repo. Is that right? If so, the first step is on me.

Issue #1084, opened 2025-04-02
Add tests for `obfuscates_id`


jagthedrummer:
We should add a basic "gut check" test for obfuscating IDs to ensure that we don't accidentally change something that would change the way that IDs are obfuscated.

For instance, say that a Team with ID 42 gets an obfuscated id of `abc`. And people have published links in their docs to `/account/teams/abc`. If we did something to change the way we obfuscate IDs and then that team id was being obfuscated as `xyz`, all of the existing links to `/account/teams/abc` would break and would need to be `/account/teams/xyz` instead.

This test should basically have one or more hard coded pre-obfuscated ids to check against to ensure that we don't change the method of obfuscation accidentally.

Related to: https://github.com/bullet-train-co/bullet_train-core/pull/1082

PR #1126, opened 2025-05-21
Add a test for `obfuscates_id`

jagthedrummer:
This is a basic stability test that will prevent us from accidentally changing the obfuscation/decoding method.

Fixes https://github.com/bullet-train-co/bullet_train-core/issues/1084


RichStone:
Nice one.

PR #1132, opened 2025-05-22
WIP: Can we make super_scaffolding a dev dependency?

jagthedrummer:


jagthedrummer:
Nope. At least not easily. It's wrapped up in a bunch of different gems in different ways, and some of them don't declare a proper dependency on the `super_scaffolding` gem which makes the dependency sneaky.

For instance the main `bullet_train` gem doesn't have a dependency for `bullet_train-super_scaffolding` (at least not a direct one) but the file at `bullet_train/app/models/concerns/memberships/base.rb` declares a relationship to a class that is part of the `super_scaffolding` gem:

```
has_many :scaffolding_completely_concrete_tangible_things_assignments, class_name: "Scaffolding::CompletelyConcrete::TangibleThings::Assignment", dependent: :destroy
```

A few other gems seem to have similar invisible dependencies on `bullet_train-super_scaffolding`.

Issue #907, opened 2024-09-02
Pagy problem with changing collection in API 


alsted:
For an API endpoint index action that has a simple filter action as:

```
   def index
      # Filter @agreements if ext_ref param is present
      if params[:ext_ref].present?
        @agreements = @agreements.where(ext_ref: params[:ext_ref])
       end
    end
```

This causes a change to the collection that Pagy is initialized with in:
```bullet_train-core/bullet_train-api/app/controllers/concerns/api/controllers/base.rb``` 

where initialization takes place as:

```
  def apply_pagination
    collection = instance_variable_get collection_variable
    @pagy, collection = pagy_cursor collection, after: params[:after], order: {id: :asc}
    instance_variable_set collection_variable, collection
  end
```

Pagination headers are set later by:
  ```
  def set_pagination_headers
      return unless @pagy

      if @pagy.has_more?
        if (collection = instance_variable_get(collection_variable))
          next_cursor = collection.last.id
          link_header = response.headers["Link"]
          link_value = "<#{modify_url_params(request.url, after: next_cursor)}>; rel=\"next\""
          response.headers["Link"] = link_header ? "#{link_header}, #{link_value}" : link_value
          response.headers["Pagination-Next"] = next_cursor
        end
      end
    end
```

The the simple filter shown is run in between apply_pagination and  set_pagination_header. The filtering can cause an empty collection which leads to an error in `next_cursor = collection.last.id` as collection is has no members.

jagthedrummer:
@alsted I understand the problem you're describing here, but I'm not sure how we'd make a fix for it inside BT.

In your example `@agreements` gets loaded *and* paginated as a convenience **before** the `index` method is called. The idea being that in the default case there is no filtering and it means that we can kind of "auto implement" the `index` method so that your controller doesn't have to load and paginate manually.

In your example you can think of it like BT is doing this:

```ruby
before_action :load_and_paginate_agreements
after_action :set_pagination_header

def load_and_paginate_agreements
  @agreements = current_team.agreements
  @pagy, @agreements = pagy(@agreements)
end
```

The actual code is a bit more complicated, but the end result is very similar to doing that.

So, if you're going to alter the `@agreements` collection inside of `index` I think you should re-paginate it by also doing:

```ruby
  @pagy, @agreements = pagy(@agreements)
```

So something like this all together:

```ruby
    def index
      # Filter and re-paginate @agreements if ext_ref param is present
      if params[:ext_ref].present?
[...]

PR #1136, opened 2025-05-27
Add an `apply_filters` method that descendent controllers can override

jagthedrummer:
Fixes: https://github.com/bullet-train-co/bullet_train-core/issues/907


bullet_train-api/app/controllers/concerns/api/controllers/base.rb
```diff
@@ -109,6 +110,13 @@ def collection=(new_collection)
     instance_variable_set collection_variable, new_collection
   end
 
+  def apply_filters
+    # An empty method that descendant controllers can override
+    # A possible implementaiton might look like:
+    #
+    # self.collection = collection.where(status: params[:filter_status]) if params[:filter_status]
```
RichStone, line 117:
```suggestion
    # self.collection = collection.where(status: params[:filter_status]) if params[:filter_status]
    #
    # Keep in mind that for adding filters to auto-generated documentation, you
    # will need to override index.yaml.erb or _paths.yaml.erbs.
```

bullet_train-api/app/controllers/concerns/api/controllers/base.rb
```diff
@@ -109,6 +110,13 @@ def collection=(new_collection)
     instance_variable_set collection_variable, new_collection
   end
 
+  def apply_filters
+    # An empty method that descendant controllers can override
+    # A possible implementaiton might look like:
```
RichStone, line 115:
```suggestion
    # A possible implementation might look like:
```

RichStone (approved):
Looks like a good simple solution. Hacking in a test in a sample stubbed controller would be cool-to-have.

Issue #866, opened 2024-06-24
Refactor the `devise-two-factor` integration to use turbo streams


jagthedrummer:
In [this discussion](https://github.com/bullet-train-co/bullet_train-core/pull/865#pullrequestreview-2133293190) @pascallaliberte said:

> Real quick, I'm thinking a slightly larger refactor is recommended. It's probably worth:
>
> 1. replacing`create.js.erb` and other `ujs` js responses with a turbo response. I'm thinking responding with a turbo-stream `update` (not `replace`) of `#two-factor`'s internals using the contents of the `devise/registrations/two_factor`
> 2. replacing the two `link_to`'s in `_two_factor.html.erb` with `button_to` to create small proper forms, which will return a turbo_stream as in step 1

jagthedrummer:
OK, I've started looking into this and I think there are 4 total flows involving `js.erb` responses that will need to be updated.

3 of them are around setting up and managing 2FA

* Setting up 2FA initially : `account/two_factors/create.js.erb`
* Verifying that your authenticator works: `account/two_factors/verify.js.erb`
* Remvogin 2FA: `account/two_factors/destroy.js.erb`

And one is around signing in:

```
devise/sessions/pre_otp.js.erb
```

The 3 around managing 2FA seem like they'll be a little more straightforward than the sign in one. For the management ones I think the relevant sections of the page can be wrapped in a `turbo_frame_tag` and then we respond with a turbo stream that replaces the contents of the frame.

The sign in one doesn't really replace sections of the page like turbo frames would, instead it returns some JS code that ends up hiding the `email` field and then un-hiding the `password` and `otp` fields. My understanding is that you can't really execute JS directly in a turbo stream response the same way that you can in a `.js.erb` response, so I think we need a different approach.

I'm thinking probably a Stimulus controller that attaches to the sign in form when 2FA is active. When the user hits the "Next" button after entering their email address the controller would make an AJAX request to the `pre_otp` endpoint where we check to see if we have a user record for the email provided. If so we hide and un-hide the correct things just like `devise/sessions/pre_otp.js.erb` does currently.

@pascallaliberte I'm wondering if you already had a good idea about how this should work, and/or if what I've described seems reasonable.

jagthedrummer:
[...]

PR #1146, opened 2025-05-30
Update the two-factor experience for turbo

jagthedrummer:
Fixes: https://github.com/bullet-train-co/bullet_train-core/issues/866

This makes the two factor flows use Turbo instead of `rails-ujs`.

For the `pre_otp` action we're using a [Self Destructing Stimulus Controller](https://boringrails.com/articles/self-destructing-stimulus-controllers/) to replace the previous `.js.erb` response that only fired some Javascript.

The 2FA management flows use normal Turbo Frames/Streams to switch out sections of the page, very similar to what the `.js.erb` was doing.

### DIFF
--- a/bullet_train/app/views/devise/registrations/_two_factor.html.erb
+++ b/bullet_train/app/views/devise/registrations/_two_factor.html.erb
@@ -1,62 +1,64 @@
-<%= render 'account/shared/box', divider: @backup_codes do |box| %>
-  <% box.title t("users.edit.two_factor.header") %>
-  <% box.description t("users.edit.two_factor.description_#{@user.otp_required_for_login? ? 'enabled' : 'disabled'}") %>
-  <% if current_user.otp_secret %>
-    <% if @backup_codes %>
-      <% box.body do %>
-        <%= render 'account/shared/alert' do %>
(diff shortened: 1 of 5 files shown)


bullet_train/app/views/devise/registrations/_two_factor.html.erb
```diff
@@ -1,62 +1,64 @@
-<%= render 'account/shared/box', divider: @backup_codes do |box| %>
-  <% box.title t("users.edit.two_factor.header") %>
-  <% box.description t("users.edit.two_factor.description_#{@user.otp_required_for_login? ? 'enabled' : 'disabled'}") %>
-  <% if current_user.otp_secret %>
-    <% if @backup_codes %>
-      <% box.body do %>
-        <%= render 'account/shared/alert' do %>
-          <%= t('users.edit.two_factor.warning').html_safe %>
-        <% end %>
+<%= turbo_frame_tag "two-factor-frame" do %>
```
jagthedrummer, line 1: This diff for this file looks messier than it is because I wrapped the whole file in a `turbo_frame_tag` so everything got indented by a level. Turning on "hide whitespace" here will make it easier to see what's going on:

CleanShot 2025-05-30 at 13 21 58

PR #1151, opened 2025-06-04
Webhook deactivation system

Riddlerrr:
This is an implementation for the issue: https://github.com/bullet-train-co/bullet_train-core/issues/854

This PR introduces the `Webhooks::Outgoing::EndpointsDeactivationJob`, which automatically deactivates endpoints that fail to process webhooks for any reason.

The job is supposed to be a scheduled job and run every X (configurable) amount of time. For this reason, `sidekiq-scheduler` was introduced to the main `bullet_train` template in [the PR](https://github.com/bullet-train-co/bullet_train/pull/2074). However, the BT core does not have any scheduled mechanisms.

This PR adds some new config settings to `outgoing_webhooks`:
```
          automatic_deactivation_endpoint_enabled: false,
          automatic_deactivation_endpoint_settings: {
            max_limit: 50,
            deactivation_in: 1.day
          },
```

I think it's better to disable the feature by default so that it can only be enabled if needed.

### How it works

The `EndpointsDeactivationJob` is split into two parts:
1. Mark endpoints for deactivation when the maximum limit of failed deliveries is reached.
2. Deactivate the marked endpoints if there are no successful deliveries.

#### Mark endpoints for deactivation

There are two new fields added to `webhooks_outgoing_endpoints`: `deactivation_limit_reached_at` and `deactivated_at`.

The job searches for the not delivered `Webhooks::Outgoing::Delivery` for each endpoint and if it is >= the `max_limit` from the config, and sets the `deactivation_limit_reached_at`.

A delivery is considered a failure if it is not delivered within the allotted time frame:

https://github.com/Riddlerrr/bullet_train-core/blob/81f7fca191aecc073ecaf71c3c646391e5eb08bf/bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb#L15-L17

We will send a notification when an endpoint is marked to deactivate.

Since the `deactivation_limit_reached_at` set, we will wait for `deactivation_in` (1 day by default) and if no successful deliveries were made during the time the job will deactivate it by setting `deactivated_at`.

For any successful delivery, the `deactivation_limit_reached_at` must be cleared so that the endpoint is no longer considered "marked for deactivation."

#### Deactivate the marked endpoints

The job checks for endpoints with a set `deactivation_limit_reached_at`. If there were no successful deliveries since the  `deactivation_limit_reached_at`, it marks the endpoint as deactivated by setting the `deactivated_at` field.

### Deactivated endpoints

Even when an endpoint is deactivated, the system will create deliveries for debugging purposes to confirm that the event occurred. However, actual delivery attempts will not be created, so the webhooks will not fire.

### UI and controller actions

This PR adds the ability to activate/deactivate endpoints. For that reason, two new actions were introduced to `Account::Webhooks::Outgoing::EndpointsController` and `Api::V1::Webhooks::Outgoing::EndpointsController`.

CleanShot 2025-05-28 at 13 39 41@2x

CleanShot 2025-05-28 at 13 28 07@2x

### Email notifications previews

(Open the spoiler to see the default mailer view)

<details>
<summary>Deactivation limit reached</summary>

CleanShot 2025-05-29 at 18 22 50@2x

</details>

<details>
<summary>Deactivated</summary>

CleanShot 2025-05-29 at 18 19 23@2x

</details>

### DIFF
--- a/bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/endpoint_support.rb
+++ b/bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/endpoint_support.rb
@@ -99,8 +99,10 @@ module Webhooks::Outgoing::EndpointSupport
     max_attempts_period = Webhooks::Outgoing::Delivery.max_attempts_period + 1.hour # Adding 1 hour to ensure it covers all delays
     max_limit = BulletTrain::OutgoingWebhooks::Engine.config.outgoing_webhooks.dig(:automatic_endpoint_deactivation_settings, :max_limit)
     last_successful_delivery = deliveries.where.not(delivered_at: nil).maximum(:delivered_at)
+    # There is a recent successful delivery, so we don't deactivate
     return false if last_successful_delivery && last_successful_delivery > (Webhooks::Outgoing::Delivery.max_attempts_period + 1.hour).ago
 
+    # All recent deliveries are failed and it's number is enough to trigger deactivation
     failed_deliveries = deliveries.where(delivered_at: nil).where("created_at < ?", max_attempts_period.ago).last(max_limit).pluck(:delivered_at)
     return false if failed_deliveries.empty?
 


bullet_train-outgoing_webhooks/app/jobs/webhooks/outgoing/endpoints_deactivation_job.rb
```diff
@@ -0,0 +1,17 @@
+class Webhooks::Outgoing::EndpointsDeactivationJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    return unless automatic_deactivation_endpoint_enabled?
+
+    endpoint_health = Webhooks::Outgoing::EndpointHealth.new
+    endpoint_health.deactivate_failed_endpoints!
```
Riddlerrr, line 8: First, deactivate to avoid processing the endpoints that have just been marked.

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
```diff
@@ -32,6 +40,7 @@ def deliver_async
   end
 
   def deliver
+    return if endpoint.deactivated?
```
Riddlerrr, line 43: This line prevents the creation of a new attempt and delivery of the webhook.

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
```diff
@@ -55,7 +64,19 @@ def still_attempting?
...
+
+  def not_attempted?
+    attempt_count.zero?
+  end
+
+  def attempts_schedule_period_elapsed?
+    created_at < max_attempts_period.ago
+  end
+
+  def failed_or_not_attempted_or_elapsed?
```
Riddlerrr, line 78: It is used for the UI to mark deliveries as failed when there have been no attempts or the maximum period has elapsed.

bullet_train-outgoing_webhooks/app/views/webhooks/outgoing/endpoint_mailer/deactivated.html.erb
```diff
@@ -0,0 +1 @@
+<h1>This is a mockup of the deactivated notification email. It must be overridden in the bullet_train template.</h1>
```
Riddlerrr, line 1: This is only a stub because I moved the real template to the template repo.

bullet_train-outgoing_webhooks/lib/bullet_train/outgoing_webhooks/engine.rb
```diff
@@ -22,7 +22,12 @@ class Engine < ::Rails::Engine
           allowed_schemes: %w[http https],
           custom_block_callback: nil,
           custom_allow_callback: nil,
-          audit_callback: ->(obj, uri) { Rails.logger.error("BlockedURI obj=#{obj.persisted? ? obj.to_global_id : "New #{obj.class}"} uri=#{uri}") }
+          audit_callback: ->(obj, uri) { Rails.logger.error("BlockedURI obj=#{obj.persisted? ? obj.to_global_id : "New #{obj.class}"} uri=#{uri}") },
+          automatic_deactivation_endpoint_enabled: false,
```
Riddlerrr, line 26: New config values here

bullet_train-outgoing_webhooks/test/dummy/config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  devise_for :users
```
Riddlerrr, line 2: Fix for make `sign_in @user` works for controller's tests

bullet_train-outgoing_webhooks/app/jobs/webhooks/outgoing/endpoints_deactivation_job.rb
```diff
@@ -0,0 +1,17 @@
+class Webhooks::Outgoing::EndpointsDeactivationJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    return unless automatic_deactivation_endpoint_enabled?
```
RichStone, line 5:
```suggestion
    return unless automatic_endpoint_deactivation_enabled?
```

Otherwise it reads like there as some kind of an endpoint responsible for automatic deactivation.

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
RichStone: I can see scenarios where developers might wanna send deliveries to deactivated endpoints for whatever reason. Maybe we should make it configurable `def deliver(check_deactivated = true)`?

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
RichStone: Just a note for @jagthedrummer: we do the first attempt quite quickly in the regular flow, but we can end up in a situation where we have not_attempted attempts and the endpoint deactivated in which case they won't be attempted anymore.

bullet_train-outgoing_webhooks/app/jobs/webhooks/outgoing/endpoints_deactivation_job.rb
RichStone: Seems like a weird naming combo: `endpoint_health.deactivate_failed_endpoints!`. Maybe something like `Webhooks::Outgoing::EndpointDoctor` :D

bullet_train-outgoing_webhooks/lib/bullet_train/outgoing_webhooks/engine.rb
RichStone: Jeremy had a good question here in my PR about whether this will be configurable in the starer repo that I haven't verified yet. Maybe you have?
(I only verified I'm seeing it in `Rails.config`. I assume we can just override them in the environment files)

bullet_train-outgoing_webhooks/app/controllers/account/webhooks/outgoing/endpoints_controller.rb
```diff
@@ -60,6 +60,32 @@ def destroy
...
+        format.html { redirect_to [:account, @parent, :webhooks_outgoing_endpoints], alert: I18n.t("webhooks/outgoing/endpoints.notifications.activation_failed") }
+        format.json { render json: @endpoint.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  # DELETE /account/webhooks/outgoing/endpoints/:id/deactivate
+  def deactivate
+    respond_to do |format|
+      if @endpoint.update(deactivated_at: Time.current, deactivation_limit_reached_at: nil)
```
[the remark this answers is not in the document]
Riddlerrr: Well, if you deactivate an endpoint with a set `deactivation_limit_reached_at` and then activate it after some time, you would expect the deactivation_limit_reached_at to not be populated, I guess.

I decided that, if there is manual action, we should reset the deactivation_limit_reached_at because it is set automatically and there is no other way to reset it.

jagthedrummer (changes requested):
Whew, this is a big one! I requested a few changes, and left some comments for general discussion, but overall this is looking really good @Riddlerrr. 🤘 Good job on the detailed tests.

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
Riddlerrr: How about `def deliver(force: false)`?

bullet_train-outgoing_webhooks/lib/bullet_train/outgoing_webhooks/engine.rb
Riddlerrr: ✅ I checked, it works

If you set for example in `environments/development.rb` works the same from any rb file from `config/initializers`

```ruby
Rails.application.configure do
  config.outgoing_webhooks[:automatic_deactivation_endpoint_enabled] = true
end
```

It will be enabled in the app:
```ruby
> Rails.configuration.outgoing_webhooks[:automatic_deactivation_endpoint_enabled]
=> true
```

bullet_train-outgoing_webhooks/app/controllers/account/webhooks/outgoing/endpoints_controller.rb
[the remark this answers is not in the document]
Riddlerrr: Ok, I clear the `deactivation_limit_reached_at` when deactivating, but not when activating.

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
```diff
@@ -7,6 +7,14 @@ module Webhooks::Outgoing::DeliverySupport
...
     has_one :team, through: :endpoint unless BulletTrain::OutgoingWebhooks.parent_class_specified?
     has_many :delivery_attempts, class_name: "Webhooks::Outgoing::DeliveryAttempt", dependent: :destroy, foreign_key: :delivery_id
+
+    after_commit :clear_endpoint_deactivation_limit_reached_at, if: :delivered?
+  end
+
+  class_methods do
+    def max_attempts_period
+      ATTEMPT_SCHEDULE.values.sum
+    end
```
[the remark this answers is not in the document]
Riddlerrr: ✅ I've added the last attempt in 24 hours.

bullet_train-outgoing_webhooks/app/mailers/webhooks/outgoing/endpoint_mailer.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  private
+
+  def set_values(endpoint)
+    @values ||= {
+      endpoint_name: endpoint.name,
+      endpoint_events: endpoint.event_type_ids.join(", "),
+      cta_url: account_webhooks_outgoing_endpoint_url(endpoint),
+    }
```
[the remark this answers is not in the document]
Riddlerrr: I did it the same way it works in UserMailer. This means that we will send `@values` to every translation in the template regardless of whether they are needed or not.

bullet_train-outgoing_webhooks/app/models/webhooks/outgoing/endpoint_health.rb
```diff
@@ -0,0 +1,73 @@
...
+    # Send notifications for endpoints that have been deactivated
+    endpoints_to_deactivate.each do |endpoint_id|
+      Webhooks::Outgoing::EndpointNotificationJob.perform_later(
+        endpoint_id: endpoint_id,
+        notification_type: "deactivated"
+      )
+    end
+
+    endpoints_to_deactivate
+  end
```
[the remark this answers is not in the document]
Riddlerrr: Oh, it seems like a very good idea 💡

Let me expand on that to describe the full process.

This action will be applied to each failed delivery for which the value of the `still_attempting?` is false. This means that we have exhausted our retries.

- Skip the endpoint is already deactivated
- If the endpoint is marked for deactivation, we check if the cooling-off period (`deactivation_in` setting) has passed. If so, we mark it as deactivated.
- We mark items for deactivation if the last 50 deliveries (`max_limit` setting) have not been delivered. SQL should be pretty simple.
- We clear the deactivation_limit_reached_at if a delivery is successful for an endpoint (already implemented)

I see two major advantages to this approach:
1. We can eliminate the scheduled task, so we can delete `sidekiq-scheduler`. Fewer dependencies ✅
2. There will be no more complicated and potentially slow SQL queries. Easier to maintain ✅

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/endpoint_support.rb
RichStone, line 1: Just a thought, all the logic here could go into its own `endpoint_health_support.rb` or `endpoint_deactivateable.rb` concern.

RichStone (changes requested):
There are some migrations and schema changes that don't belong to this PR. ⚠️ 

--- 

I think the last iteration is a big improvement in terms of lowering the complexity and removing the scheduler dependency, but I wouldn't go for it without exploring to store the health state in Redis, unless there are any major concerns with doing it.

I would probably start fresh on a new branch and take all the good stuff from this PR, tests, mailers, UI, account, and resketch the deactivation functionality.

Redis has an INCR counter feature that we can increment and reset once LIMIT is reached.

I think this could overall further simplify some parts of our code.

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
RichStone: > Are you thinking of some sort of one-off manual attempt scenario? (Like, isn't avoiding delivery attempts to deactivated endpoints pretty much the whole point of this feature?)

Yeah @jagthedrummer, like an endpoint failed and a customer asks the developer to rerun the failed attempts but leave the endpoint deactivated (e.g., because they already have a new endpoint active to do the job).

At some point, we'll also have a "send/resend event manually" feature in the UI. Someone might want to do this on a deactivated endpoint.

> How about def deliver(force: false)?

I think this might do the job.

But we still want the fix below to avoid `Webhooks::Outgoing::Event` records being created for deactivated endpoints:

https://github.com/bullet-train-co/bullet_train-core/blob/5e6d042448abba11abe929c3c2ec11618a6301de/bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/team_support.rb#L15-L17

bullet_train-outgoing_webhooks/app/controllers/concerns/api/v1/webhooks/outgoing/endpoints/controller_base.rb
```diff
@@ -25,7 +25,8 @@ def endpoint_params
   included do
     account_load_and_authorize_resource :endpoint,
       through: BulletTrain::OutgoingWebhooks.parent_association,
-      through_association: :webhooks_outgoing_endpoints
+      through_association: :webhooks_outgoing_endpoints,
+      member_actions: [:activate, :deactivate]
```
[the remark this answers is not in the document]
Riddlerrr: > I would go the Stripe route, and "simply" allow to change the state of the endpoint in the update action:

Good point! That seems like perfect solution 🔥

bullet_train-outgoing_webhooks/app/models/concerns/webhooks/outgoing/delivery_support.rb
jagthedrummer: > But we still want the fix below to avoid Webhooks::Outgoing::Event records being created for deactivated endpoints

@RichStone what's your reasoning on this? I would think that any time we attempt to send a webhook for any reason that we should log it the same as always.

PR #1164, opened 2025-06-27
Fix outgoing webhook reader lag failure scenario

benjaminwood:
With reader lag, we can end up in a situation where the event is not found when attempting to deliver a webhook.

This change checks for the presence of the event and returns early with an error message if it is not found.

I've taken this opportunity to write comprehensive unit test coverage for delivery attempts.


RichStone:
@benjaminwood should we close this and have another PR for the latest stable approach to this?

PR #1170, opened 2025-07-10
Multiple locales: fix broken button_field_options

pascallaliberte:
When adding a second locale and navigating to user settings (user/edit), an error is shown.

The call to the button field uses the deprecated `options` key, and it should be replaced with `button_field_options`


jagthedrummer (approved):
Thanks, @pascallaliberte!

PR #1173, opened 2025-07-15
fix: `to_param` for non-persisted records

Paul-Bob:
This PR fixes a bug where calling `to_param` would raise `can't convert nil into Integer` for non-persisted records.


RichStone:
Hey @Paul-Bob, I'm curious, in which scenario would `encode` be called for non-persisted records?

natannobre:
> @RichStone I flagged the issue to @Paul-Bob . 
> 
> 
> 
> Our bullettrain app has avo-pro global search that breaks when visiting a record#new page within Avo

I also got into this problem trying to use STI in bullettrain.

Issue #1220, opened 2025-09-29
Update sortable docs to include more details about re-issued events


jagthedrummer:
In [this section](https://bullettrain.co/docs/super-scaffolding/sortable#events) we should mention that you need to add an ancestor element that wraps the `sortable` element and that declares a `data-controller` attribute to connect the confirmation controller.

In practice this might look like adding the controller to the outer `table` element like so:

```html
<table class="table" data-controller="confirm-reorder">
```

Issue #1215, opened 2025-09-16
Can we replace `dragula` in `bullet_train-sortable`?


jagthedrummer:
`dragula` seems like it's abandoned and it relies on some packages that are completely deprecated.

In exploring making a transition to importmaps `dragula` and it's dependencies are giving me trouble. So if we can't find a work around for those problems then this issue may be a blocker on moving to importmaps.

Related sortable issue: https://github.com/bullet-train-co/bullet_train-core/issues/826

jagthedrummer:
[`stimulus-components/sortable`](https://www.stimulus-components.com/docs/stimulus-sortable) looks interesting. But it's based on [`SortableJS`](https://github.com/SortableJS/Sortable) which seem about as abandoned as `dragula`.

jagthedrummer:
We may be better off doing something like this so that we're not relying on any dependencies. https://onrails.blog/2024/03/08/hotwire-tutorial-how-do-i-drag-and-drop-items-in-a-list/

Issue #826, opened 2024-05-07
Sortable lists should be more obviously sortable


jagthedrummer:
Currently if you super scaffold a sortable model you'll end up with an index page that is technically sortable, but there's no indication given to the user that the items can be reordered.

```
rails g super_scaffold Blog Team name:text_field --sortable
```

![CleanShot 2024-05-07 at 11 10 14](https://github.com/bullet-train-co/bullet_train-core/assets/58702/f13b1580-1c8d-4c1c-b55e-6c2119a850a3)

I think we should probably add a defined "drag handle" (something like ↕️) to help indicate that the items are draggable.

It would also be nice to change the cursor when someone is hovered over the draggable item. Probably `cursor: grab` when hovered over the item, and then `cursor: grabbing` when the drag has been started.

The lack of dragging indicators might be at play in this comment: https://github.com/bullet-train-co/bullet_train-core/pull/825#issuecomment-2098188135

swombat:
Agreed. I ended up disabling on most lists this since it made clicking less reliable, and gave rise to users saying "the list isn't working" because they moved the mouse a tiny bit while clicking.

I'd advise:
- drag handle
- light dashed outline

Between those two things it should be more obvious that this is draggable.

swombat:
This from the Refactoring UI book suggests an even better solution, visually:

<img width="824" alt="image" src="https://github.com/bullet-train-co/bullet_train-core/assets/4488/c7d2204d-98b9-4422-941a-7d41d1aad7a8">

PR #1222, opened 2025-09-30
New sortable controller that doesn't rely on `dragula`

jagthedrummer:
This PR moves the old `sortable_controller.js` to `dragula-sortable_controller.js` and replaces the contents of `sortable_controller.js` with a controller that doesn't rely on `dragula`. And it removes `dragula` and `jquery` as dependencies of `bullet_train-sortable`.

It also updates the scaffolding transformer to add drag handles directly in the template when a new sortable model is scaffolded. (For ease of upgrades the new controller will default to programmatically adding drag handles to pre-existing sortable tables.)

The new `dragula-sortable_controller.js` file is **not** included in the build for `bullet_train-sortable` because doing so causes JS compilation to fail in the downstream app if the app doesn't have `dragula` in the `dependencies` object in `package.json`.

Joint PR: https://github.com/bullet-train-co/bullet_train/pull/2228

Fixes: https://github.com/bullet-train-co/bullet_train-core/issues/1215
Fixes: https://github.com/bullet-train-co/bullet_train-core/issues/826
Fixes: https://github.com/bullet-train-co/bullet_train-core/issues/1220


jagthedrummer:
@pascallaliberte I made all of those changes that you suggested.

Can you give me a gut check on something? I'm looking at the event emissions from the sortable controller and trying to figure out what we should do.

I'm kind of leaning towards defining a new "main" set of events that correspond to identifiable/notable events for the list itself (as opposed to events for the html elements that make up the list) and which do not collide with and of the old`dragula` events. Then also having a small set of "backwards compatibility" events where we make a best effort to emit the events that the old controller emitted at roughly the same time. And then a set of "we don't emit these or never did" events that were dragula specific.

Main:
* `sortable:start` - When a drag via drag handle is started
* `sortable:reordered` - When the list reorders mid-drag
* `sortable:end` - When dragging stops
* `sortable:saved` - When the new order has been persisted after a drop

Backward compat:
* `sortable:drag` - When a drag is started
* `sortable:dragend` - When dragging stops
* `sortable:drop` - When dragging stops
* `sortable:shadow` - When the list reorders mid-drag

Unsupported:
* `sortable:cancel`
* `sortable:remove`
* `sortable:over`
* `sortable:out`
* `sortable:cloned`

Does that general approach sound reasonable? (I really have no idea if any/many people are using the events emitted from that controller.)

jagthedrummer:
> I wonder about us using event.stopPropagation()

I've also wondered about this, and I'm also not sure. I think we could theoretically issue our own events to signal logic changes to the list and also allow the native events to propagate. I do question how useful that would be, but I'm really just speculating and theorizing about use cases that I don't have. So... 🤷

bullet_train-sortable/app/javascript/controllers/dragula-sortable_controller.js
```diff
@@ -0,0 +1,88 @@
+import { Controller } from "@hotwired/stimulus"
```
jagthedrummer, line 1: This is the old `sortable_controller.js` just renamed to `dragula-sortable_controller.js` so that it can still be referenced by people who want to use it.

bullet_train-sortable/app/javascript/controllers/sortable_controller.js
```diff
@@ -1,88 +1,337 @@
 import { Controller } from "@hotwired/stimulus"
```
jagthedrummer, line 1: This is really a whole new controller, but it looks like a bunch of changes to `sortable_controller.js` because `git` is kinda dumb about replacing the entire contents of files. It may be easier to review by looking at the raw source of the controller rather than the diff.

bullet_train/docs/super-scaffolding/dragula-sortable.md
```diff
@@ -0,0 +1,188 @@
+# Super Scaffolding with the `--sortable` option
```
jagthedrummer, line 1: This is the old sortable docs, but updated with some info about how to continue to use `dragula` if you want to.

bullet_train/docs/super-scaffolding/sortable.md
```diff
@@ -1,5 +1,17 @@
 # Super Scaffolding with the `--sortable` option
```
jagthedrummer, line 1: This is really new docs for the new sortable controller, but it looks like extensive changes to the old docs Because Git.

jagthedrummer:
OK, @pascallaliberte, I think I've addressed all of the lingering todos. Let me know if you think of anything else. Thanks for all your help on this! I think it's going to be a good improvement.

bullet_train/docs/super-scaffolding/dragula-sortable.md
```diff
@@ -0,0 +1,188 @@
...
+
+Since we don't include `dragula` and `jquery` as dependencies anymore you need to include them in your own `package.json`.
+
+```
+yarn add dragula jquery
+```
+
+### 2. Copy `dragula-sortable_controller.js` into your project
+
+You can grab the old controller [from the `core` repo here](https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-sortable/app/javascript/controllers/dragula-sortable_controller.js).
```
jagthedrummer, line 28: This link is pointing to where the file will live once this PR is merged.

bullet_train-sortable/app/javascript/controllers/sortable_controller.js
jagthedrummer: I think that's what I did. Or maybe I meant to and didn't...

bullet_train/docs/super-scaffolding/sortable.md
```diff
@@ -124,13 +157,47 @@ export default class extends Controller {
...
+</thead>
+```
+
+### `app/views/account/pages/_page.html.erb`
+
+In `app/views/account/pages/_index.html.erb` you should add a `<td>` as the first cell in the `<tr>`. Be sure to include `data-sortable-target="handle"` on the cell so that the controller recognizes it as a handle.
+
+```html
+<tr data-id="<%= page.id %>">
+    <td class="cursor-grab active:cursor-grabbing" data-sortable-target="handle"><i class="ti ti-menu opacity-25 group-hover:opacity-100"></i></td> <!-- Add this line! -->
```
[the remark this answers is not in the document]
jagthedrummer: I like the idea but it doesn't seem to work. It seems like `group-hover` takes precedence over `hover` and so it never goes darker. I think the cursor changing to `grab` is probably enough confirmation that the handle is active and will do something.

PR #1237, opened 2025-10-15
light theme: on initialize, warn if darkMode set in tailwind.config.js

pascallaliberte:
If a user has added a `darkMode` setting in their local `tailwind.config.js`, it probably means they wanted to show the `light` mode to all their users.

```js
themeConfig.darkMode = 'selector' // or the deprecated 'class'
```

Or maybe they once did, and then backtracked, and set the value to `media` back again.

```js
themeConfig.darkMode = 'media'
```

In either case, applying the update to the `light` theme gem will for them produce a surprise. In the first case the dark mode will show to some users who are running dark mode in their system preferences, and in the second, some styles will appear off (for components like the super_select, date fields, and trix editor), as we no longer use the `@media` rule for selecting dark mode.

This PR solves this issue by warning devs and encouraging them to remove their custom darkMode setting entirely, and introducing the new `force_color_scheme_to` setting.

CleanShot 2025-10-15 at 10 01 01@2x

### DIFF
--- a/bullet_train-themes-light/lib/bullet_train/themes/light/engine.rb
+++ b/bullet_train-themes-light/lib/bullet_train/themes/light/engine.rb
@@ -15,55 +15,55 @@ module BulletTrain
 
           Rails.application.config.after_initialize do
             tailwind_config_path = Rails.root.join("tailwind.config.js")
-            
+
             if tailwind_config_path.exist?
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
This is a nice touch.

PR #1254, opened 2025-10-31
Add some documenation about how to contribute

jagthedrummer:


jagthedrummer:
@pascallaliberte, can you think of anything that would be useful to add here?

PR #1263, opened 2025-11-06
Consolidate `@tailwind` calls into a single file

jagthedrummer:
This is in service of somewhat simplifying the way that we integrate Tailwind to hopefully make the transition to v4 a bit easier.


jagthedrummer:
Apparently the `@tailwind` directive needs to be in a file that's `@import`ed by `light.tailwind.css`. It's weird. `light.tailwind.css` is itself `@import`ed by `application.css` in the starter repo, but there seems to be some sort of "boundary" that makes some tool or another screw up.

PR #1266, opened 2025-11-06
making previews for attachments in action text

1MasterSpark:
In [this PR](https://github.com/bullet-train-co/bullet_train/pull/2269), I made it so that attachments in action text render previews for images, videos, GIFs, PDFs, and files that can't be rendered like ZIP files. This time I replaced all the hardcoded style attributes in _blob with Tailwind CSS, and I'm making two PRs, this one and [this one](https://github.com/bullet-train-co/bullet_train/pull/2293). @jagthedrummer what do you think?

### DIFF
--- a/bullet_train-themes/app/views/themes/base/attributes/_html.html.erb
+++ b/bullet_train-themes/app/views/themes/base/attributes/_html.html.erb
@@ -5,10 +5,9 @@
 <% if object.public_send(attribute).present? %>
   <%= render 'shared/attributes/attribute', object: object, attribute: attribute, strategy: strategy, url: url do %>
     <% if object.send(attribute).is_a?(ActionText::RichText) %>
-      <%= html_sanitize(object.public_send(attribute).body.to_s).html_safe %>
-    <% else %>
-      <% # `.to_s` is for action text. %>
-      <%= html_sanitize(object.public_send(attribute).to_s).html_safe %>
+      <%= tag.div(class: "trix-content") do %>
+        <%= object.public_send(attribute) %>
+      <% end %>
     <% end %>
   <% end %>
 <% end %>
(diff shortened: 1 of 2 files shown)


bullet_train-themes/app/views/themes/base/attributes/_html.html.erb
```diff
@@ -5,10 +5,9 @@
 <% if object.public_send(attribute).present? %>
   <%= render 'shared/attributes/attribute', object: object, attribute: attribute, strategy: strategy, url: url do %>
     <% if object.send(attribute).is_a?(ActionText::RichText) %>
-      <%= html_sanitize(object.public_send(attribute).body.to_s).html_safe %>
-    <% else %>
```
jagthedrummer: I suspect we still want this else block.

jagthedrummer (commented):
After a quick peek at this, without pulling it down and running anything, I'm a little concerned that some conditionals have been removed. I suspect that we're going to need them, but it's hard to say for sure without running some tests. I'll try to dig in further tomorrow. Thanks, @1MasterSpark!

bullet_train-themes/app/views/themes/base/attributes/_image.html.erb
```diff
@@ -1,13 +1,8 @@
...
 
-<% if cloudinary_enabled? %>
-  <%= cloudinary_image_tag object.public_send(attribute), options %>
-<% else %>
-  <% if object.public_send(attribute).attached? %>
-    <%= image_tag photo_url_for_active_storage_attachment(object.public_send(attribute), options) %>
-  <% end %>
+<% attachment = object.public_send(attribute) %>
+<% if attachment && attachment.attached? %>
+  <%= render 'active_storage/blobs/blob', blob: attachment, options: options %>
```
jagthedrummer: This is an interesting idea. It would (potentially) be nice to centralize some of this stuff, but we do still need to support Cloudinary, so we can't just throw that stuff out.

jagthedrummer:
@1MasterSpark, we just released version `1.37.0` which contains all of the PRs I linked above. I think it should address many of the issues that you've discovered, though some of them were addressed in different ways than you proposed. Please give it a try when you can and let us know if there are still any things giving you trouble.

PR #1275, opened 2025-11-12
Render videos that are embeded in a TrixEditor

jagthedrummer:
This makes it so that we render a video embedded in a `trix_editor` field.

Inside the editor we show a thumbnail of the first frame:

CleanShot 2025-11-12 at 15 21 00

But after saving the record we show an embedded player:

CleanShot 2025-11-12 at 15 20 37

**NOTE:** This uses a native `<video>` element, so support for any given video will be dependent on what video formats are supported by the underlying browser/OS combo. I think there's an argument to be made that this isn't the standard we want to keep for BT and that it would be better to just show a thumbnail and link to the video itself. I would love to hear what other folks think about that.

Note this is a stacked PR and points into the branch for https://github.com/bullet-train-co/bullet_train-core/pull/1274. We should merge https://github.com/bullet-train-co/bullet_train-core/pull/1274 before this one and then release them at the same time.

### DIFF
--- a/bullet_train/app/views/active_storage/blobs/_blob.html.erb
+++ b/bullet_train/app/views/active_storage/blobs/_blob.html.erb
@@ -3,6 +3,11 @@
     <video controls width="<%= local_assigns[:in_gallery] ? 800 : 1024 %>" preload="metadata" class="max-w-full h-auto rounded-md">
       <source src="<%= rails_blob_url(blob) %>" type="<%= blob.content_type %>">
     </video>
+    <%
+      # This next thing kind of sucks, but seems necessary. If we don't force an image representation to be created after a record
+      # is saved then when you edit the record we show a broken image link instead of a good one in the editor.
+    %>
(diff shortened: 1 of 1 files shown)


bullet_train-fields/app/helpers/fields/html_editor_helper.rb
```diff
@@ -5,7 +5,10 @@ def html_sanitize(string)
     return string unless string
     # TODO this is a hack to get around the fact that rails doesn't allow us to add any acceptable protocols.
     string = string.gsub("bullettrain://", TEMPORARY_REPLACEMENT)
-    string = sanitize(string, tags: %w[div br strong em b i del a h1 blockquote pre ul ol li action-text-attachment figure figcaption img], attributes: %w[href sgid content-type url filename filesize width height presentation src class])
+    string = sanitize(string,
+                      tags: %w[div br strong em b i del a h1 blockquote pre ul ol li action-text-attachment figure figcaption img video source],
+                      attributes: %w[href sgid content-type url filename filesize width height presentation src class controls preload type]
+                     )
```
jagthedrummer: Here I added line breaks to make things more readable, and added `video source` to `tags` and `controls preload type` to `attributes`.

PR #1276, opened 2025-11-13
Set a poster for the video player

jagthedrummer:
This makes the player a little friendlier to users, and has the benefit of causing Rails to generate an image representation without us needing to include a hidden image on the page.


pascallaliberte (approved):
Nice

PR #1283, opened 2025-11-26
passing html options down to _base

1MasterSpark:
On my own app, I had to eject some partials and add html_options to them so that I can put `{ data: { turbo_frame: "_top" } }` on some links where it needed to be. I thought this would be helpful for everyone else too. If you like this change, I will also do it with the rest of the attributes partials.

### DIFF
--- a/bullet_train-themes/app/views/themes/base/attributes/_attribute.html.erb
+++ b/bullet_train-themes/app/views/themes/base/attributes/_attribute.html.erb
@@ -3,7 +3,7 @@
 <% url ||= nil %>
 <% link_html_options ||= {} %>
 
-<%= render 'shared/attributes/base', strategy: strategy, url: url, attribute: attribute, html_options: link_html_options do |p| %>
+<%= render 'shared/attributes/base', strategy: strategy, url: url, attribute: attribute, link_html_options: link_html_options do |p| %>
   <% p.heading t("#{object.class.name.pluralize.underscore}.fields.#{attribute}.heading", default: object.class.human_attribute_name(attribute)) %>
   <% p.body yield %>
 <% end %>


1MasterSpark:
Me too. Is it better now?

1MasterSpark:
Thanks! I just added `link_html_options` to the rest of the attributes, except for `_image` and` _images`. I wasn't sure if I should add it to them because they don't use `_attribute` and they already have options

1MasterSpark:
Hi @pascallaliberte 

Can you please merge this? Also, I'm sorry for causing this, but there is a problem with action_text_sanitizer.rb. I tried to install bullet train in a new project and got this error:
```
/home/spark/.local/share/mise/installs/ruby/3.3.6/lib/ruby/gems/3.3.0/gems/bullet_train-1.37.1/config/initializers/action_text_sanitzer.rb:8:in `block in <top (required)>': uninitialized constant ActionText::Attachment (NameError)

  default_allowed_attributes = Rails::HTML5::Sanitizer.safe_list_sanitizer.allowed_attributes + ActionText::Attachment::ATTRIBUTES.to_set
                                                                                                          ^^^^^^^^^^^^
Did you mean?  ActionText::Attachments
```

bullet_train-themes-light/app/views/themes/light/attributes/_base.html.erb
```diff
@@ -1,7 +1,8 @@
 <% url ||= nil %>
 <% strategy ||= current_attributes_strategy || :none %>
+<% link_html_options ||= {} %>
 
-<% body = link_to_if url.present?, partial.body, url %>
+<% body = link_to_if url.present?, partial.body, url, link_html_options %>
```
jagthedrummer: Since we're calling the rails helper `link_to_if` here isn't this the spot where we need to transition from `link_html_options` to just `html_options`?

```suggestion
<% body = link_to_if url.present?, partial.body, url, html_options: link_html_options %>
```

1MasterSpark: So should it just be `link_html_options` in both `attribute` and `base`?

1MasterSpark: Or maybe we could make it so that `base` accepts both html_options and link_html_options, combines them into one hash, then puts it in the `link_to_if`

jagthedrummer (approved):
OK, @1MasterSpark this looks great and I think it's ready to go. I'm not sure if I'll be able to get it merged and released today or if it'll be early next week. But definitely soon.

Issue #1208, opened 2025-09-03
Document how to replace the main font


pascallaliberte:
Currently, the main `inter` font is included from a `cdn` directly in the theme's `layouts/_head`.

It would be nice to expose a `_fonts` partial (easier to eject), and an optional and empty `fonts.css` file that we could override locally.

Also, it would be nice to add an entry in our documentation on how to define fonts, particularly how we recommend loading local font file assets.

PR #1284, opened 2025-11-27
Document how to customize fonts

pascallaliberte:
Joint PR: https://github.com/bullet-train-co/bullet_train/pull/2325

This PR:
- adds a way to more easily remove the default `Inter` font: creates a new `head/_fonts` partial.
- updates the docs with instructions on Customizing Fonts

Fixes https://github.com/bullet-train-co/bullet_train-core/issues/1208


pascallaliberte:
I think `inter` is definitely part of the `light` theme, and so should stay bundled with it.

To slim down the instructions, I think this still be better than, say, including the inter font from a fonts.css we could eject (which not be picked up locally anyway without a custom postcss-import-config variable e.g. `$ThemeFontsStylesheet`

